PR #1, opened 2025-01-06
fix display button on payment page

G-ladal:
Actual code miss few lines to have payment button on member page (display button like stripe/paypal/...)


eldy:
Can you tell me which version of Dolibarr you used ? Is it v21 or v20 ?

eldy:
Closed without answer. I think problem is only with dolibarr v20 (not yet compatible with module helloasso).
Should use v21

PR #18, opened 2025-10-01
New fetch for helloAsso members

Hystepik:
+ Add a new way to fetch HelloAsso members using pagination + date.
+ Add some constants


eldy:
@Hystepik can you fix conflict ?

PR #21, opened 2025-10-08
try to enable facturX

rycks:
first test to enable a module :)


rycks:
@eldy as i have the rights to merge may i do it ?

PR #38, opened 2025-11-14
FIX avoid error when member email is required

hregis:


hregis:
@Hystepik it's ok for you ?

Hystepik:
It's ok for me

PR #119, opened 2026-04-03
FIX : Add GlobalID to IssuerTradeParty in CdarHandler

atm-maxime:
Fixes #107


Daoud-mohamed:
@atm-maxime I’ve just pushed a commit to complete your fix...

PR #155, opened 2026-05-15
re-add missing method

FHenry:
_findOrCreateProductFromFacturXLine and _detectProductTypeFromFacturx are still used


eldy:
The function were renamed and moved into the Trait CommonProtocom.class.php
I pushed a different fix. Thanks to check.

FHenry:
@eldy its works !

PR #158, opened 2026-05-16
Add check for empty action in printFieldListWhere

defrance:
generate error on other sql list from custom module hook


pdpconnectfr/class/actions_pdpconnectfr.class.php
```diff
@@ -747,6 +747,9 @@ public function printFieldListFrom($parameters, &$object, &$action, $hookmanager
 	 */
 	public function printFieldListWhere($parameters, &$object, &$action, $hookmanager)
 	{
+		if (empty($action)) {
```
eldy, line 750: I don't understand why we need this.
Any time we call printFieldListWhere(), in a list view (so with $action empty) or any other case, we should add the filter in the select.

What SQL do you get ?

PR #159, opened 2026-05-16
    Fix : Dolibarr version 24 is not a sufficient test. Check for existence of Dolibarr FPDF class that is a subclass of TCPDF to distinguish with FPDF class defined in setasign instead

bafbes:
Fix : Dolibarr version 24 is not a sufficient test. Check for existence of Dolibarr FPDF class that is a subclass of TCPDF to distinguish with FPDF class defined in setasign instead. It also avoids unnecessary checks against a fixed Dolibarr version.

### DIFF
--- a/pdpconnectfr/class/protocols/FacturXProtocol.class.php
+++ b/pdpconnectfr/class/protocols/FacturXProtocol.class.php
@@ -509,7 +509,7 @@ class FacturXProtocol extends AbstractProtocol
 		// 2 methods are provided depending on the version of Dolibarr.
 		// TODO A third method can be tried using the atgp/factur-x library.
 
-		if ((float) DOL_VERSION < 24.0) {
+		if (class_exists('FPDF', false) && is_subclass_of('FPDF', 'TCPDF')) {
 			// Generate the PDF including the XML using the TCPDF library.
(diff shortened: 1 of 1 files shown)


pdpconnectfr/class/protocols/FacturXProtocol.class.php
```diff
@@ -509,7 +509,7 @@ public function generateInvoice($invoice_id, $outputlangs = null)
 		// 2 methods are provided depending on the version of Dolibarr.
 		// TODO A third method can be tried using the atgp/factur-x library.
 
-		if ((float) DOL_VERSION < 24.0) {
+		if (class_exists('FPDF', false) && is_subclass_of('FPDF', 'TCPDF')) {
```
eldy, line 512: What is goal of change ?
Beeing able to enter this part of code when Dolibarr is lower than v23, or the opposite ?

bafbes:
This code is better regardless of the Dolibarr version because it only checks for the real issue: whether the Dolibarr FPDF class already exists, thus avoiding the “class already defined” error. In fact, this change became necessary on SuperPDP with Dolibarr 23.0.2, because SuperPDP does not accept the factur-x.xml file being embedded twice in the PDF.

PR #165, opened 2026-05-21
fix pdpconnectfr.lang

frederic34:


eldy:
Note that like Dolibarr, language files are erased by transifex during transifex sync.

PR #192, opened 2026-05-26
[pdpconnectfr] Fix Fatal error "Cannot redeclare pdfExtractMetadata()" on Dolibarr v22+

### LINKED ISSUE #191: [pdpconnectfr] Fatal error "Cannot redeclare pdfExtractMetadata()" on Dolibarr v22+
## Environment

- Dolibarr version: 22.0.4 (also affects any version where `pdfExtractMetadata()` exists in `core/lib/pdf.lib.php`)
- Module: `pdpconnectfr` (current `main`, `f46b3ee`)

## Symptom

Triggering invoice validation on Dolibarr v22 produces a 500 / fatal:

```
PHP Fatal error: Cannot redeclare function pdfExtractMetadata()
(previously declared in /var/www/html/htdocs/custom/pdpconnectfr/lib/pdpconnectfr.lib.php:562)
in /var/www/html/htdocs/core/lib/pdf.lib.php on line 2818
```

URL that triggers it (any invoice validation):
`/compta/facture/card.php?facid=<id>&action=confirm_valid&confirm=yes&token=<token>`

## Root cause

`pdpconnectfr/lib/pdpconnectfr.lib.php` ships a polyfill for `pdfExtractMetadata()` wrapped in a `function_exists()` guard:

```php
if (!function_exists('pdfExtractMetadata')) {
    function pdfExtractMetadata(...) { ... }
}
```

The polyfill was introduced by commit `fdcbc0b` (PR #175 "clean_code") for older Dolibarr versions where the function didn't exist in core.

On Dolibarr v22, `pdfExtractMetadata()` exists in `core/lib/pdf.lib.php`. The guard fails to protect because of include order:

1. The module's `pdpconnectfr.lib.php` is loaded before `core/lib/pdf.lib.php` is required by the invoice card flow.
2. `function_exists('pdfExtractMetadata')` returns `false` -> the polyfill is declared.
3. Later, `core/lib/pdf.lib.php` is required and redeclares the function -> PHP fatal.

## Proposed fix

Force-load `core/lib/pdf.lib.php` immediately before the `function_exists()` check. `require_once` is idempotent and safe — on versions where the file exists but `pdfExtractMetadata` doesn't, the polyfill still declares; on v22+ the core function is loaded first, `function_exists()` returns true and the polyfill is correctly skipped.

## Fix branch

https://github.com/pscoffoni-opendsi/dolibarr-community-modules/tree/fix/pdpconnectfr-pdfextractmetadata-redeclare-v22

---

*Generated with the help of Claude Code, reviewed and tested by a human*

pscoffoni-opendsi:
## Summary

- Fixes #191.
- Force-loads `core/lib/pdf.lib.php` before the `function_exists('pdfExtractMetadata')` guard in `pdpconnectfr/lib/pdpconnectfr.lib.php`, so the guard correctly detects the function on Dolibarr versions where core provides it (v22+) and the polyfill is skipped.

## Why

On Dolibarr v22, the polyfill in the module collides with core's own `pdfExtractMetadata()` because `pdpconnectfr.lib.php` is loaded before `core/lib/pdf.lib.php`: the `function_exists()` guard returns false, the polyfill declares the function, and core's later `require_once` triggers `Fatal error: Cannot redeclare function pdfExtractMetadata()`. Validating any invoice 500s as a result.

## Compatibility

- v22+: `pdfExtractMetadata()` exists in core -> the forced `require_once` loads it first, the polyfill is skipped.
- v18-v21: `pdfExtractMetadata()` doesn't exist in core but `core/lib/pdf.lib.php` is still present -> the polyfill still declares as before.
- `require_once` is idempotent; the `file_exists()` probe is paranoia for unusual install layouts.

## Test plan

- [x] Reproduce the fatal on v22 with the current `main`.
- [x] Apply the patch -> invoice validation succeeds, no fatal in logs.
- [x] Smoke-test on v18 / v20 to confirm the polyfill is still active where needed.

---

*Generated with the help of Claude Code, reviewed and tested by a human*

### DIFF
--- a/pdpconnectfr/lib/pdpconnectfr.lib.php
+++ b/pdpconnectfr/lib/pdpconnectfr.lib.php
@@ -550,8 +550,14 @@ if (!method_exists('Societe', 'findNearest')) {
 	}
 }
 
-if (!function_exists('pdfExtractMetadata'))
-{
+// Force-load core's pdf.lib.php on versions where pdfExtractMetadata exists in core,
+// otherwise function_exists() returns false here (core lib not loaded yet) and the
+// polyfill below is declared, then core's later require_once redeclares it -> fatal.
+if (file_exists(DOL_DOCUMENT_ROOT . '/core/lib/pdf.lib.php')) {
+	require_once DOL_DOCUMENT_ROOT . '/core/lib/pdf.lib.php';
+}
+
+if (!function_exists('pdfExtractMetadata')) {
 	/**
 	 * Function to extract metadata from a PDF file by doing a binary parsing of the PDF file
 	 *
--- a/pdpconnectfr/lib/pdpconnectfr.lib.php
+++ b/pdpconnectfr/lib/pdpconnectfr.lib.php
@@ -334,8 +334,7 @@ if (!function_exists("GETPOSTFLOAT")) {
 	}
 }
 
-if (!function_exists('dolPrintHTML'))
-{
+if (!function_exists('dolPrintHTML')) {
 	/**
 	 * Return a string ready to be output on HTML page
 	 * To use text inside an attribute, use can use only dol_escape_htmltag()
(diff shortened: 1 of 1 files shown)


pscoffoni-opendsi:
Closing as obsolete: upstream commit dd49942 (\"Fix regressions\") resolved the underlying problem by **removing the \`pdfExtractMetadata\` polyfill from \`pdpconnectfr/lib/pdpconnectfr.lib.php\` entirely**, which removes the source of the redeclare collision. No further patch is needed on top of current \`main\`.

Closing the linked issue #191 as resolved by the same upstream commit.

---

*Generated with the help of Claude Code, reviewed and tested by a human*

PR #212, opened 2026-06-02
clean permissions (double level not needed)

frederic34:


eldy:
Do we really need a dedicated permission for just log ? If a user is allowed to read or make synchronization, why would he not be authorized to read the log related to the action he is allowed to.

And on the other side, if you are allowed to read log or modify log without being allowed tobread real event, it has no sense.

For this reason only the permission to read only synchronization and log + permission to manage synchronization and logs were kept.

frederic34:
it's just because the tooltip in call list don't display

aspangaro:
I confim...

image

PR #237, opened 2026-06-17
NEW option : check supplier invoice consistency on validate

spsolauv:
If new param `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` is enabled, then on supplier invoice validation (trigger), data of Dolibarr supplier invoice are compared to data from e-invoice to ensure they are identical, depending following criteria : 
- Currency
- VAT excl. total
- VAT incl. total
- VAT total
- Basis amount & VAT amount of each VAT rate

Amounts comparison precision can be customized using param : `EINVOICING_SUPPLIER_INVOICE_COMPARISON_ROUND_PRECISION`

We also save XML data in column xml_data in table einvoicing_document, so we can access e-invoicing data to do the comparison (even if the AP is offline). Doing it using the attached file will not be totally sure because the attachment can be deleted by the user.

Update requires module reactivation.


Daoud-mohamed:
@spsolauv Thanks for this feature, really useful!

Since e-invoice XML can embed base64 PDF attachments, we may need a cleanup function to remove attachment nodes before saving the XML.

spsolauv:
I didn't forget, it's on my todolist and I will add it asap :)

spsolauv:
@Daoud-mohamed I just added the clean of attachments. In this version, I just replace binary content by a string to not lose attachments metadata. I can modify to completely remove attachment nodes if necessary.

PR #260, opened 2026-06-22
fix #246

### LINKED ISSUE #246: [Einvoicing][multientity] Supplier is not found if suppliers are shared between entities
In the case where suppliers are shared between entities with the multientity module, synchronization will not find the supplier if it was created in another entity.

hansemschnokeloch:
With this PR we only restrain thirdparty search per entity if multicompany module is enabled and thirdparty sharing between entities is disabled

closes #246

@atm-maxime as you're the author of commit https://github.com/Dolibarr/dolibarr-community-modules/commit/5121aa9cb3a42cecb54a395a8a9707493a406324 who added the sql query per entity, could you review this PR and check if anything else is needed ?

### DIFF
--- a/einvoicing/class/protocols/CommonProtocol.class.php
+++ b/einvoicing/class/protocols/CommonProtocol.class.php
@@ -462,7 +462,7 @@ trait CommonProtocol
 		// Step 2: Try to find using VAT number if not found by global IDs
 		if ($thirdpartyId < 0) {
 			if (!empty($sellerInfo['sellerTaxRegistations']['VA'])) {
-				$sql = "SELECT rowid FROM " . MAIN_DB_PREFIX . "societe WHERE REPLACE(tva_intra, ' ', '') = '" . $db->escape($einvoicing->removeSpaces($sellerInfo['sellerTaxRegistations']['VA'])) . "' AND entity = ".$conf->entity;
+				$sql = "SELECT rowid FROM " . MAIN_DB_PREFIX . "societe WHERE REPLACE(tva_intra, ' ', '') = '" . $db->escape($einvoicing->removeSpaces($sellerInfo['sellerTaxRegistations']['VA'])) . "' AND entity IN (". getEntity('societe').")";
 				$resql = $db->query($sql);
 				if ($resql) {
 					if ($db->num_rows($resql) > 1) {


einvoicing/class/protocols/CommonProtocol.class.php
```diff
@@ -462,7 +462,11 @@ private function _syncOrCreateThirdpartyFromEInvoiceSeller($sellerInfo, $priorit
 		// Step 2: Try to find using VAT number if not found by global IDs
 		if ($thirdpartyId < 0) {
 			if (!empty($sellerInfo['sellerTaxRegistations']['VA'])) {
-				$sql = "SELECT rowid FROM " . MAIN_DB_PREFIX . "societe WHERE REPLACE(tva_intra, ' ', '') = '" . $db->escape($einvoicing->removeSpaces($sellerInfo['sellerTaxRegistations']['VA'])) . "' AND entity = ".$conf->entity;
+				$sql = "SELECT rowid FROM " . MAIN_DB_PREFIX . "societe WHERE REPLACE(tva_intra, ' ', '') = '" . $db->escape($einvoicing->removeSpaces($sellerInfo['sellerTaxRegistations']['VA'])) . "'";
+				// If multicompany module is enabled and thirdparty sharing between entities is disabled we restrain search of thirdparty per entity
+				if (isModEnabled('multicompany') && !getDolGlobalInt('MULTICOMPANY_THIRDPARTY_SHARING_ENABLED')) {
+					$sql .= ' AND entity = '.$conf->entity;
```
eldy: I think what you must do is replacing the line entity = conf->entity 
by
entity IN ('. getEntity('societe').')'
So multientity sharing will be managed with no extra code.

eldy (changes requested):

einvoicing/class/protocols/CommonProtocol.class.php
hansemschnokeloch: Thx.
I wasn't familiar with this getEntity function.

hansemschnokeloch: PR updated

atm-maxime:
Your fix is actually the correct one. Mine was too much of a quick fix

PR #266, opened 2026-06-23
[Einvoicing] Fix #247

### LINKED ISSUE #247: [Einvoicing] Unable to change or define default product for a supplier
Default product field can only be defined when creating a new supplier.
If you are modifying a supllier the field is not shown

Image

hansemschnokeloch:
closes #247

This PR allows to edit and update the default product when we are in `edit mode`.


einvoicing/class/einvoicing.class.php
```diff
@@ -1708,7 +1708,7 @@ public function thirdpartyCardBlock($object, $mode = '', $parameters = array())
 
 		// In create mode only : we show a text input (the thirdparty is not yet in database, no routing line exists)
 		// In edit mode, we show the routing array
-		if ($mode == 'create') {
+		if ($mode == 'create' || $mode == 'edit') {
```
eldy: In edit mode, there is already a section to edit the routing id
image

hansemschnokeloch: Ok.
I modified the fix. We now keep the routing id address editing as before, and we can change the default product in edit mode

PR #268, opened 2026-06-23
fix(einvoicing): lock a transmitted invoice on the real PA state (persistent flow_id)

daGrumpf-bxp:
### Problem
An invoice already transmitted to the Access Point is immutable (you correct it with a credit note / corrective invoice). The module already had locks for this, but they keyed on the Dolibarr **syncstatus**, which is reset to `GENERATED` when the e-invoice is **regenerated**. So after editing + regenerating an already-sent invoice:
- the `BILL_UNVALIDATE` / `BILL_MODIFY` edit-locks released, and
- the "send to PDP" guard re-enabled, so re-sending hit the PA with a raw `HTTP 400 CREATE_ERROR — invoice already exists`.

### Fix
Key the locks on the **real PA state** instead of the resettable status: a `flow_id` is assigned on the first successful submission (by **any** provider, through the common `EInvoicing::insertOrUpdateExtLink()`) and is never cleared, so it survives a regenerate/re-open.

- `fetchLastknownInvoiceStatus()` now also returns `flow_id` and a persistent `everTransmitted` flag (alongside the existing status-derived `transmitted`).
- New `EInvoicing::isTransmittedLockActive()` centralizes the rule.
- It guards: `send_to_pdp` and `generate_einvoice` (blocked with a clear, actionable message instead of a raw PA 400), the regenerate/send buttons (hidden), and the `BILL_UNVALIDATE` / `BILL_MODIFY` triggers (hardened — addresses the existing "disable Modify once transmitted" TODO).
- New opt-out **`EINVOICING_ALLOW_RESEND_TRANSMITTED`** (off by default, GUI yes/no) restores the deliberate "retry to test the PA" behaviour.

Provider-agnostic (verified `flow_id` is populated by both SuperPDP and Esalink via the shared helper). FR/EN strings included.

### Verified
On a sandbox account: a transmitted invoice reports `everTransmitted=1` / locked; after forcing `syncstatus→GENERATED` (the regenerate case) `transmitted=0` but `everTransmitted=1`, so it stays locked; enabling `EINVOICING_ALLOW_RESEND_TRANSMITTED` unlocks it.


einvoicing/admin/setup.php
```diff
@@ -172,6 +172,19 @@
...
+// Allow re-sending / re-editing an invoice already transmitted to the Access Point. Off by default:
+// a transmitted invoice is immutable (correct it with a credit note / corrective invoice), and re-sending
+// makes the PA refuse a duplicate. Turn on only to deliberately test PA retry behaviour.
+$item = $formSetup->newItem('EINVOICING_ALLOW_RESEND_TRANSMITTED')->setAsYesNo();
+$item->defaultFieldValue = '0';
+$item->helpText = $langs->transnoentities('EINVOICING_ALLOW_RESEND_TRANSMITTED_HELP');
+
+// Dev-only: keep the "Regenerate e-invoice" button/action available on a transmitted-locked invoice
+// (rebuild the CII/Factur-X to inspect the XML). Re-sending stays locked. Off by default.
+$item = $formSetup->newItem('EINVOICING_ENABLE_DEV_CII_REGEN')->setAsYesNo();
```
eldy: My understanding is that name of this option is 
EINVOICING_ALLOW_REGEN_TRANSMITTED
right ?
Seems more compliant with the other name 
EINVOICING_ALLOW_RESEND_TRANSMITTED

daGrumpf-bxp: Good catch, renamed to `EINVOICING_ALLOW_REGEN_TRANSMITTED` for consistency. Done in b30dde8.

einvoicing/admin/setup.php
```diff
@@ -191,6 +191,19 @@
 $item = $formSetup->newItem('EINVOICING_LIVE')->setAsYesNo();
 $item->fieldParams['forcereload'] = 1;
 
+// Allow re-sending / re-editing an invoice already transmitted to the Access Point. Off by default:
+// a transmitted invoice is immutable (correct it with a credit note / corrective invoice), and re-sending
+// makes the PA refuse a duplicate. Turn on only to deliberately test PA retry behaviour.
+$item = $formSetup->newItem('EINVOICING_ALLOW_RESEND_TRANSMITTED')->setAsYesNo();
```
eldy: It looks better to move this 2 new options into the tab "Options" in section 
Synchronisation Dolibarr -> PA/PDP

daGrumpf-bxp: Done — both options moved to the **Options** tab (section *Synchronisation Dolibarr → PA/PDP*), and each label now carries a greyed warning that they must not be used in production (breaks the AFNOR/PA/PDP contracts), translated in all available langs.

einvoicing/langs/ro_RO/einvoicing.lang
```diff
@@ -377,3 +377,4 @@ ReasonRecipientError=Recipient error
 ReasonUnknownTransaction=Unknown transaction
 ReasonUnknownIssuer=Unknown issuer
 ReasonAntivirusControl=Antivirus control
+EINVOICING_TRANSMITTED_NOT_FOR_PROD=Atenție: nu trebuie utilizat în producție, încalcă contractele AFNOR/PA/PDP
```
eldy, line 380: There is no need to modify language files that are non en_US.
Content in git is always erased by transifex sync. You must change language file on transifex if you want to set translate keys.

PR #269, opened 2026-06-23
[Einvoicing] NEW supplier invoice : allow manual import of lines

spsolauv:
### Feature manual import

If enable, manual import will : 
- import supplier invoices without any line
- add a new button to reimport supplier invoice lines with 3 options : 
  - reimport automatically (current behavior)
  - reimport to free lines (no product creation and no product linked to invoice line)
  - reimport to a target product (will create one line per VAT rate, where each is linked to the selected target product)
  - user can also create lines on its own (completely manually) but he has to ensure consistency with amounts of e-invoice to pass validation

Add new parameters in module settings : 
- EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE :  select list to choose automatic (current behavior) or manual import of supplier invoice lines
- EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_CATEGORY_OF_TARGET_IMPORT_PRODUCT_LIST : will be used to filter the list of products by this selected category

Add an extrafield `einvoicing_supplier_invoice_lines_import_type` on thirdparties, to be able to customize import type by thirdparty (if not set, will use default parameter value of EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE)

### About code
To add this feature, I need to refactor some parts of code : 
- supplier invoice lines were added in database by FactureFournisseur create method => a new method `createSupplierInvoiceLinesFromSource` was added to externalize this part, to be able to reimport lines independently 

Other modifications :
- `parseInvoiceXML` has been renamed to `parseInvoiceHeader` : because it not parsing the full XML, just the einvoice header
- some methods have been moved to CommonProtocol to avoid code duplication
- new static method : ProtocolManager::getProtocolFromContent() to directly get the protocol object from content (avoid 2 steps : detect then create object)

### :warning: Remaining modifications 
`CIIProtocol::createSupplierInvoiceFromSource()` remains to be modified (to uniformise line creation like modifications in `FacturXProtocol` => calling `createSupplierInvoiceLinesFromSource`) but I don't understand why the commit 4fed7772bf0bf16fa902b747bfd8c9dd44a549a0 made some modifications only in CIIProtocol and not in `FacturXProtocol`, shouldn't they have the same code/behavior ? Line creation shouldn't be the same between CII & FacturX ? @Daoud-mohamed 
- Example of modification which is present in `CIIProtocol` in the loop adding supplier invoice lines, but not in `FacturXProtocol` : 
```php
		// Create document level discounts (allowances) as discounts in Dolibarr
		$globalDiscountIds = array();
		if (!empty($parsedHeader['headerAllowancesCharges'])) {
			$headerDiscountIds = $this->_createHeaderDiscounts($parsedHeader['headerAllowancesCharges'], $socId, 	$parsedHeader['documentno']);
			if (!empty($headerDiscountIds[-1])) {
				return ['res' => -1, 'message' => $headerDiscountIds[-1]];
			} else {
				$globalDiscountIds = $headerDiscountIds;
			}
		}
```


einvoicing/admin/setup_options.php
```diff
@@ -269,6 +269,15 @@
 	$item->fieldAttr['placeholder'] = $langs->transnoentities('Hours');
 	$item->cssClass = 'maxwidth100';
 
+	$item = $formSetup->newItem('EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE');
+	$item->setAsSelect([
+		1 => $langs->trans('EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE1'),
```
eldy: Default option means notor never set, so you must use 0 for default option and not 1.

einvoicing/js/supplier_invoice.js
```diff
@@ -0,0 +1,38 @@
+const einvoicingSupplierInvoice = {
```
eldy, line 1: js files must be into directory **module/js** not into **module/assets/js**

einvoicing/class/actions_einvoicing.class.php
```diff
@@ -52,6 +53,39 @@ public function messageOfTheDay($parameters, $object, &$action, $hookmanager)
...
+			];
+			$supplierPricesTranslations = array_map(function ($labelId) use ($langs) {
+				return html_entity_decode($langs->trans($labelId));
+			}, $supplierPricesTranslations);
+
+			print '<script>';
+			print "const einvoicingTranslations = JSON.parse('".addslashes(json_encode($supplierPricesTranslations))."');";
+			print '</script>';
+
+			print '<script defer src="'. dol_buildpath('/custom/einvoicing/assets/js/supplier_invoice.js?v='.einvoicingGetModuleVersion(), 1) . '"></script>';
```
eldy: dol_buildpath must no contains /custom, it is already included inside function.Must be
`dol_buildpath('einvoicing/...)`

einvoicing/class/actions_einvoicing.class.php
```diff
@@ -334,6 +372,77 @@ public function addMoreActionsButtons($parameters, &$object, &$action, $hookmana
...
+				0,	// maxlen
+				0,	// disabled
+				'', // sort
+				'minwidth300'	// morecss (CSS class for styling)
+			);
+
+			// Modal allowing to reimport supplier invoice lines
+			print '<div id="einvoicing-dialog-import-lines-form" title="'.$langs->trans('EinvoiceImportLines').'" style="display: none;">';
+			print '<form id="einvoicing-import-lines-form" method="post" action="'.dol_buildpath('fourn/facture/card.php', 1).'?facid='.$object->id.'&action=reimportLines&token='.newtoken().'">';
+			print 	'<p style="color: #e11717; font-weight: bold;">❗ ' . $langs->trans('SupplierInvoiceExtractLinesWarning').'</p>';
```
eldy: Use of emitocon in HTML code is forbidden.
Use function like  img_picto('alert'), img_picto('warning', ...  if you need visual picto.

einvoicing/class/actions_einvoicing.class.php
```diff
@@ -334,6 +372,77 @@ public function addMoreActionsButtons($parameters, &$object, &$action, $hookmana
...
+			print 	'<div style="display: flex; flex-direction: column; margin-left: 2rem; gap: 0.2rem;">
+						<label for="extraction-all-prices"><input type="radio" name="extraction_type" id="extraction-all-prices" checked="checked" value="1">'.$langs->trans('ExtractAllLines').'</label>
+						<label for="extraction-to-free-line"><input type="radio" name="extraction_type" id="extraction-to-free-line" value="2">'.$langs->trans('ExtractToFreeLine').'</label>
+						<label for="extraction-to-product"><input type="radio" name="extraction_type" id="extraction-to-product" value="3">'.$langs->trans('ExtractToAProduct') . (isset($category) ? ' ('.$langs->trans('SupplierInvoiceLinesImportProductsFromCategory'). ' <strong>' . $category->label .'</strong>)' : '');'</label>
+					</div>';
+			print   '<div id="extraction-target-product-choice" style="display: none; margin-top: 0.5rem; margin-left: 2rem;">';
+			if ($categoryId > 0) {
+				print $selectProduct;
+			} else {
+				print '⚠️ ' . $langs->trans('SupplierInvoiceLinesImportPleaseSetDefaultCategoryInSettings') . '<br>';
```
eldy: Use of emitocon in HTML code is forbidden.
Use function like img_picto('alert'), img_picto('warning', ... if you need visual picto.

einvoicing/class/helpers/SupplierInvoiceHelper.class.php
```diff
@@ -58,22 +58,17 @@ public static function checkDolInvoiceAndEInvoiceConsistency(FactureFournisseur
...
+			$errors[] = $langs->trans('EinvoiceFailedToDetectXmlFormat');
+			return [
+				'identical' => false,
+				'errors' => $errors,
+			];
 		}
-		$protocol = $protocolManager->getProtocol($detectedProtocolName);
 
 		// Extract XML header data
-		switch ($detectedProtocolName) {
```
eldy, line 70: Imagine $xmlData and $protocol has been detected as UBL.
The old code was returning an exception to explain we can't continue validation process.
With your new code, this test and exception has been removed. How will be the error management in this case ?

einvoicing/admin/setup_options.php
spsolauv: I will add EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE const with default value set to 1 in modEinvoicing. I need to set default value to 1 to be consistent with extrafield `einvoicing_supplier_invoice_lines_import_type`, which need to have : 
- 0/null value as fallback using EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE global module value
- 1 as EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE1
- 2 as EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE2

einvoicing/class/helpers/SupplierInvoiceHelper.class.php
spsolauv: That's right, I have just modified code of `getProtocolFromContent` to change the way to manage unknown/unsupported format, which now return more informations, so detected errors can be managed by the calling method.

spsolauv:
For the last CI checks, I don't know how to fix them, they concern previously existing part of code I've just moved to a different file https://github.com/Dolibarr/dolibarr-community-modules/actions/runs/28169432532

GregcodeOhm:
Hi @spsolauv, I ran into the same phan failure on #347, so here is what is happening and how to clear it.

None of these are new bugs. They are all pre-existing issues that lost their baseline suppression because the code moved. CommonProtocol.class.php is a new file (phan has no baseline entries for it yet), and the ProtocolManager.class.php warnings are on lines that shifted. Phan's baseline (dev/tools/phan/baseline.txt) suppresses known issues by file and line, so relocated code is reported as "new" and fails the job even though the logic did not change.

Two ways to fix it:

1) Regenerate the baseline (recommended for a move/refactor, least work). Run phan with the same flags as .github/workflows/phan.yml, but swap the load flag -B dev/tools/phan/baseline.txt for --save-baseline=dev/tools/phan/baseline.txt, then commit the updated baseline.txt. For example (keep the same --minimum-target-php-version value as the workflow):

phan -k dev/tools/phan/config.php --save-baseline=dev/tools/phan/baseline.txt --analyze-twice --minimum-target-php-version 7.2
This is legitimate here because the issues pre-existed; you are relocating them, not introducing them. The baseline header itself documents --save-baseline for exactly this case.

2) Or fix the flagged lines now that they are visible, for example DiscountAbsolute->fk_soc to ->socid, ->description to ->desc, initialize $return_messages = array() and $remise_already_used_line_level_ids = array() before the loops that populate them, cast the vatrate argument to (float) for FactureFournisseur::updateline(), and guard the nullable $content before detectProtocolFromContent(). More work, but it clears the debt instead of hiding it.

For a straight move, option 1 is the usual choice. Option 2 if you prefer to clean it up in the same PR.

Unrelated small note: I see einvoicing.lang appears twice in the file list. Only the en_US lang file should be committed in a PR; the other language files are synced by Transifex and should not be edited by hand.

Hope this helps.

spsolauv:
It's more clear on what I can do to fix this, thanks for explanations @GregcodeOhm :pray:

spsolauv:
Aborted (to split + too much conflicts), new PR is #367

PR #275, opened 2026-06-23
fix(einvoicing): progress (situation) invoices billed 100% per line instead of the progressed amount (#258)

### LINKED ISSUE #258: [e-invoicing] - Progress invoice, amount error in the lines
When entering a progress invoice, we set one line at 15% and another at 5%, the PDF shows correctly on the client side, but Dolibarr receives 2 lines each showing the total amount

Image

Image

daGrumpf-bxp:
Fixes #258

### Bug
On a progress / situation invoice (`Facture::TYPE_SITUATION`), the generated CII / Factur-X XML billed **100 % of every line**, ignoring the per-line progress. The PDF was correct (it uses `$line->total_ht`) but the e-invoice shipped the full amounts in `LineTotalAmount`, `TaxBasisTotalAmount` and `GrandTotalAmount`.

Example — two lines at 15 % and 5 % progress (1000 € and 500 € HT):

| | before | after |
|---|---|---|
| Line A `LineTotalAmount` | `1000.00` | `150.00` |
| Line B `LineTotalAmount` | `500.00` | `25.00` |
| `TaxBasisTotalAmount` | `1500.00` | `175.00` |
| `GrandTotalAmount` | `1800.00` | `210.00` |

### Fix
For `TYPE_SITUATION` invoices, fold the progress fraction — derived from the authoritative `$line->total_ht`, which already nets out the progress billed by previous situations — into the unit price (BT-146) while keeping the real billed quantity (BT-129). So `BT-146 × BT-129` stays consistent with the line net amount (BT-131), and the header totals recompute from the prorated line amounts (BR-CO-10 / BR-CO-13 stay satisfied). Shared by both the CII and Factur-X builders (`lib/buildinvoicelines.inc.php`).

Verified end-to-end on Dolibarr 23.0.3: regenerated XML matches the PDF.

### DIFF
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -425,6 +425,24 @@ foreach ($object->lines as $line) {
 	$line_total_ttc = $line->total_ttc;
 	*/
 
+	// Progress / situation invoices (facture de situation): Dolibarr keeps the FULL (100%) line
+	// amount in subprice*qty, while the amount actually billed for THIS situation - already net of
+	// the progress billed by previous situations - lives in $line->total_ht (the value shown on the
+	// PDF). Without correction the e-invoice would bill 100% of every line (issue #258). We fold the
+	// progress fraction into the unit price (BT-146) and keep the real billed quantity (BT-129), so
+	// BT-146 * BT-129 stays consistent with the line net amount BT-131. Unit price and quantity are
+	// both emitted with 2 decimals, so scaling the price - not the quantity - preserves that
+	// consistency even for non-round progress percentages (e.g. 33.33%).
+	if ($object->type == $object::TYPE_SITUATION) {
+		$line_full_ht = $line_total_ht;	// = $line_unit_price_with_discount * $line->qty (100% of the line)
+		$progress_ratio = ($line_full_ht != 0.0) ? ((float) $line->total_ht / $line_full_ht) : 0.0;
+
+		$line_unit_price = price2num($line_unit_price * $progress_ratio, 2);
+		$line_total_ht   = price2num($line_unit_price_with_discount * $progress_ratio * $line->qty, 2);
+		$line_total_tva  = price2num($line_total_ht * ($line->tva_tx > 0 ? number_format($line->tva_tx, 2, '.', '') / 100 : 0), 2);
+		$line_total_ttc  = price2num($line_total_ht + $line_total_tva, 2);
+	}
+
 	// Add (or update) VAT rate to $taxBreakdown
 	if (!isset($taxBreakdown[$line->tva_tx.($line->vat_src_code ? ' ('.$line->vat_src_code.')' : '')])) {
 		$taxBreakdown[$line->tva_tx.($line->vat_src_code ? ' ('.$line->vat_src_code.')' : '')] = ['tva_tx' => '', 'vat_src_code' => '', 'categoryVAT' => '', 'ExemptionReasonCode' => '', 'ExemptionReason' => '', 'totalHT' => 0, 'totalTVA' => 0];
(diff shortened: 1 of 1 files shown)


einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -415,6 +415,24 @@
...
+	// progress fraction into the unit price (BT-146) and keep the real billed quantity (BT-129), so
+	// BT-146 * BT-129 stays consistent with the line net amount BT-131. Unit price and quantity are
+	// both emitted with 2 decimals, so scaling the price - not the quantity - preserves that
+	// consistency even for non-round progress percentages (e.g. 33.33%).
+	if ($object->type == $object::TYPE_SITUATION) {
+		$line_full_ht = $line_total_ht;	// = $line_unit_price_with_discount * $line->qty (100% of the line)
+		$progress_ratio = ($line_full_ht != 0.0) ? ((float) $line->total_ht / $line_full_ht) : 0.0;
+
+		$line_unit_price = price2num($line_unit_price * $progress_ratio, 2);
+		$line_total_ht   = price2num($line_unit_price_with_discount * $progress_ratio * $line->qty, 2);
```
eldy: If you need to use the ratio on line, it means you are using the not support deprecated mode for situation invoices INVOICE_USE_SITUATION = 1.
This mode was never released officially and was never stable (data are corrupted in database so could not be supported by einvoicing)

If you sue the mode INVOICE_USE_SITUATION= 2, data are correctly saved on line level with only the amount of the progress ratio, and this patch is useless.

daGrumpf-bxp:
Thanks @eldy. I tested exactly this: set `INVOICE_USE_SITUATION=2` and created fresh situation invoices on Dolibarr 23.0.3 — the bug still reproduces, so the patch is still needed.

In mode 2, the line stores the **full** unit price in `subprice`; only `total_ht` carries the prorated amount (via `situation_percent`). Example, two lines 1000 € @15% and 500 € @5%:

| | mode 2 storage | e-invoice **without** patch | **with** patch | PDF / header `total_ht` |
|---|---|---|---|---|
| LineTotalAmount | `subprice=1000/500`, `total_ht=150/25` | 1000 / 500 | 150 / 25 | — |
| TaxBasisTotalAmount | | **1500** ❌ | 175 ✅ | 175 |
| GrandTotalAmount | | **1800** ❌ | 210 ✅ | 210 |

The reason: the e-invoice builder doesn't read `total_ht` — by EN16931 rounding rules it recomputes each line from `subprice × qty` (`buildinvoicelines.inc.php`), independently of the situation mode. So even in mode 2 it bills 100% per line. The patch folds the `total_ht / (subprice×qty)` ratio into the unit price.

I also verified a 2nd situation in the cycle (cumulative): mode 2 stores the per-situation increment in `total_ht` (250/125), and since the ratio is derived from that value — which already nets out previous situations — the prorated XML is correct there too (375 / 450).

So `INVOICE_USE_SITUATION=2` doesn't make the issue go away; the patch is required regardless of mode.

ABISEsalink:
@eldy modifying qty or price would make it functionnaly OK with RFE

I have no idea on the impact it might have on the overall functionnality.
Should a new INVOICE_USE_SITUATION=3? be created and activated by default when activating module einvoicing ?

eldy:
I see. You 're right @daGrumpf-bxp .  Dolibarr use an accurate way for calculation to reduce the maximum as possible the lost in rounding (when EN16931 ask to round to 2 at each step of calculation and to redo the calculation from the unit price).
Also, we do not store the unit price we used for situation invoice line after the application of the situation percent. 

So we can use the direct data of Dolibarr, but it was disabled to be used only for test purpose by setting EINVOICNG_USE_DOLIBARR_ALREADY_CALCULATED_AMOUNTS.

So we redo calculation from unit price, and because the unit price after situation percent was not saved, we must redo it. 
However fix still seems wrong. i suggested another fix.
Thanks for checking on your side.

PR #280, opened 2026-06-23
fix(einvoicing): exclude product_type 9 (subtotal/title) lines from XML generation

### LINKED ISSUE #248: [e-invoicing] Adding titles with sub-total module generate error on xml generation
I added titles and sub-total via the sub-total native module.
I got this error on these lines:

Erreur lors de la génération du fichier XML.
MISSINGSETUP: Unknown exempting VAT reason for the line 10566 that has the VAT rate 0 and VAT code ''.
The administrator must define a valid VATEX code (reason of VAT exemption) for the couple (VAT rate / code) = (0 / '') into the constant MAIN_VAT_EXEMPTION_CODE_FOR_0.

Product type 9 must be excluded

daGrumpf-bxp:
## Fixes #248

Adding titles / sub-totals via the native subtotal feature broke XML generation with:

> MISSINGSETUP: Unknown exempting VAT reason for the line … that has the VAT rate 0 and VAT code ''.

### Cause
Subtotal/title lines are `product_type = 9` pseudo-lines with no VAT. The skip in `buildinvoicelines.inc.php` only matched lines from the `modSubtotal` module (`special_code == modSubtotal->numero`). Subtotals from the **native core** subtotal feature use `special_code = SUBTOTALS_SPECIAL_CODE` (104), so they slipped through, reached `getCategoryRate()` and threw the VATEX exemption error on rate 0 / empty code.

### Fix
Skip **any** line with `product_type == 9`, consistent with the existing `product_type < 9` filter already used in the discount loops (CII/FacturX). Title/subtotal/page-break lines are never real invoice lines.

### Test
Reproduced on a dev instance: feeding a `product_type=9` / rate 0 / no-code line to `getCategoryRate()` throws the exact issue error; with the fix the line is skipped before that call. Normal lines (type 0/1) are unaffected; modSubtotal lines were already type 9 so they stay excluded — no regression.


einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -228,9 +228,9 @@
 foreach ($object->lines as $line) {
 	$isDepositLine = 0;
 
-	// Skip subtotal lines
-	$isSubTotalLine = $this->_isLineFromExternalModule($line, $object->element, 'modSubtotal');
-	if ($isSubTotalLine) {
+	// Skip title / subtotal / page-break lines (product type 9). These are not real invoice lines and carry no VAT,
+	// so they must not be sent to getCategoryRate() (would trigger a VATEX exemption error on rate 0 / no code).
+	if ((int) $line->product_type == 9) {
```
eldy: The old test _isLineFromExternalModule() was already introduced to manage this.
May be you should fix the _isLineFromExternalModule() function instead ?

daGrumpf-bxp: You're right — done. I moved the detection back into `_isLineFromExternalModule()` instead of inlining at the call site. It now also covers the native core subtotal feature (`htdocs/subtotals`, `CommonSubtotal::$PRODUCT_TYPE = 9`), whose lines no longer carry the legacy modSubtotal module number. Rebased on current main; CI is green.

PR #287, opened 2026-06-24
NEW: support a delivery address distinct from the billing address (Sh…

frederic34:
…ipToTradeParty)


daGrumpf-bxp:
This overlaps with #267, which already covers the same feature (CII `ShipToTradeParty` / BG-15 from a distinct `SHIPPING` contact, with a normalized bill-vs-ship comparison, the BR-57 `CountryID` guard, and unit tests).

The one thing #267 was missing was your **shipment-address fallback** (`expedition.fk_delivery_address` when no SHIPPING contact is set) — good idea, so I just added it to #267 (with our existing BG-15 guards kept, i.e. only `Name` + `PostalTradeAddress`, no buyer SIREN/VAT on the deliver-to party).

So everything here should now be in #267 — suggest we converge there and close this one. Thanks for the fallback idea!

PR #292, opened 2026-06-25
Fix #291 : [e-invoicing] manual send_invoice errors not logged in llx…

### LINKED ISSUE #291: [e-invoicing] manual send_invoice errors not logged in llx_einvoicing_call with superPDP
When you send an invoice to superPDP and there is an error, the call and this error is not recorded in llx_einvoicing_call

vmaury:
Fix #291 : [e-invoicing] manual send_invoice errors not logged in llx_einvoicing_call

Because in case of errors, a $db->rollback was occurring

### DIFF
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -370,6 +370,7 @@ class ActionsEInvoicing extends CommonHookActions
 		$outputlangs = $langs;
 
 		$error = 0;
+		$dbRollBackOnError = true;
 
 		$db->begin();
 
@@ -443,6 +444,7 @@ class ActionsEInvoicing extends CommonHookActions
 						$error++;
 						$this->error = $provider->error;
 						$this->errors = array_merge($this->errors, $provider->errors);
+						$dbRollBackOnError = false;
 					}
 				}
 			}
@@ -605,7 +607,11 @@ class ActionsEInvoicing extends CommonHookActions
 		}
 
 		if ($error) {
-			$db->rollback();
+			if ($dbRollBackOnError) {
+				$db->rollback();
+			} else {
+				$db->commit();
+			}
 			return -1;
 		} else {
 			$db->commit();


einvoicing/class/actions_einvoicing.class.php
```diff
@@ -443,6 +444,7 @@ public function doActions($parameters, &$object, &$action, $hookmanager)
 						$error++;
 						$this->error = $provider->error;
 						$this->errors = array_merge($this->errors, $provider->errors);
+						$dbRollBackOnError = false;
```
eldy, line 447: We must never disable the rollbacking system, this may generate a commit when there is an error that need a rollback somewhere else.
If we want the tracking to be done even on error (legitimate), we should do the sameway we do for webhooks logging, the logging (so the $call->create() that create the log entry) must be done using a second transaction db handler.
So something like this in the ->callApi()

  ```
              global $dbhistory;
				if (empty($dbhistory)) {
					$dbhistory = getDoliDBInstance($conf->db->type, $conf->db->host, (string) $conf->db->user, $dolibarr_main_db_pass, (string) $conf->db->name, (int) $conf->db->port);
				}

				$dbhistory->begin();

$call = new Call($dbhistory);
...
$call->create() 

                $dbhistory->commit();
```

**Comment by daGrumpf-bxp** [comment]
Hit the same on SuperPDP/Esalink. Here's @eldy's `$dbhistory` suggestion implemented, in case it helps.

Root cause is in `callApi()`, not `doActions()`: the log is written on `$this->db`, inside the caller's transaction, so any `$db->rollback()` on error drops it. Writing it on a separate connection keeps it for every flow (send_invoice, send_status, sync), not just the manual send path, and leaves the main transaction untouched.

Factored the (duplicated) logging block out of both providers into one helper on `AbstractPDPProvider`:

```php
/**
 * Log an API call on a SEPARATE db connection so the trace survives a rollback
 * of the caller's transaction on error (#291). Same as the webhook logging.
 */
protected function logCall($callType, $resource, $method, $params, $response, $statusCode)
{
    global $conf, $user, $dolibarr_main_db_pass, $dbhistory;

    if (empty($callType)) {
        return null;
    }
    if (empty($dbhistory)) {
        $dbhistory = getDoliDBInstance($conf->db->type, $conf->db->host, (string) $conf->db->user, $dolibarr_main_db_pass, (string) $conf->db->name, (int) $conf->db->port);
    }

    $dbhistory->begin();

    $call = new Call($dbhistory);
    $call->call_id = $call->getNextCallId();
    $call->call_type = $callType;
    $call->method = ($method == 'POSTALREADYFORMATED' ? 'POST' : $method);
    $call->endpoint = '/' . $resource;
    $call->request_body = is_array($params) ? json_encode($params) : $params;
    $call->response = is_array($response) ? json_encode($response) : $response;
    $call->provider = $this->name;
    $call->entity = $conf->entity;
    $call->status = ($statusCode == 200 || $statusCode == 202) ? 1 : 0;

    if ($call->create($user) > 0) {
        $dbhistory->commit();
        return array('id' => $call->id, 'call_id' => $call->call_id);
    }

    $dbhistory->rollback();
    dol_syslog(__METHOD__ . " Failed to log API call to PDP provider: " . $call->error . " - " . implode(',', $call->errors), LOG_ERR);
    return null;
}
```

End of each provider's `callApi()` becomes:

```php
$logged = $this->logCall($callType, $resource, $method, $params, $returnarray['response'], $returnarray['status_code']);
if ($logged !== null) {
    $returnarray['id'] = $logged['id'];
    $returnarray['call_id'] = $logged['call_id'];
}
```

With this the `$dbRollBackOnError` change isn't needed. Note `Call::getNextCallId()` reads via `global $db` (read-only, fine); only the `create()` insert goes through `$dbhistory`.

Tested on a real instance (SuperPDP, test mode): a 404 call inside `$db->begin()` ... `$db->rollback()` now leaves its `llx_einvoicing_call` row (status=0), where before it was dropped:

```
callApi -> status_code=404  call_id=Call-000068
Rows[marker]: before=0 after=1, last status=0
PASS: call log survived the rollback
```

phpcs-clean against the repo ruleset. Can open a separate PR if you prefer.

eldy:
> Hit the same on SuperPDP/Esalink. Here's @eldy's `$dbhistory` suggestion implemented, in case it helps.
> 
> Root cause is in `callApi()`, not `doActions()`: the log is written on `$this->db`, inside the caller's transaction, so any `$db->rollback()` on error drops it. Writing it on a separate connection keeps it for every flow (send_invoice, send_status, sync), not just the manual send path, and leaves the main transaction untouched.
> 
> Factored the (duplicated) logging block out of both providers into one helper on `AbstractPDPProvider`:
> 
> ```php
> /**
>  * Log an API call on a SEPARATE db connection so the trace survives a rollback
>  * of the caller's transaction on error (#291). Same as the webhook logging.
>  */
> protected function logCall($callType, $resource, $method, $params, $response, $statusCode)
> {
>     global $conf, $user, $dolibarr_main_db_pass, $dbhistory;
> 
>     if (empty($callType)) {
>         return null;
>     }
>     if (empty($dbhistory)) {
>         $dbhistory = getDoliDBInstance($conf->db->type, $conf->db->host, (string) $conf->db->user, $dolibarr_main_db_pass, (string) $conf->db->name, (int) $conf->db->port);
>     }
> 
>     $dbhistory->begin();
> 
>     $call = new Call($dbhistory);
>     $call->call_id = $call->getNextCallId();
>     $call->call_type = $callType;
>     $call->method = ($method == 'POSTALREADYFORMATED' ? 'POST' : $method);
>     $call->endpoint = '/' . $resource;
>     $call->request_body = is_array($params) ? json_encode($params) : $params;
>     $call->response = is_array($response) ? json_encode($response) : $response;
>     $call->provider = $this->name;
>     $call->entity = $conf->entity;
>     $call->status = ($statusCode == 200 || $statusCode == 202) ? 1 : 0;
> 
>     if ($call->create($user) > 0) {
>         $dbhistory->commit();
>         return array('id' => $call->id, 'call_id' => $call->call_id);
>     }
> 
>     $dbhistory->rollback();
>     dol_syslog(__METHOD__ . " Failed to log API call to PDP provider: " . $call->error . " - " . implode(',', $call->errors), LOG_ERR);
>     return null;
> }
> ```
> 
> End of each provider's `callApi()` becomes:
> 
> ```php
> $logged = $this->logCall($callType, $resource, $method, $params, $returnarray['response'], $returnarray['status_code']);
> if ($logged !== null) {
>     $returnarray['id'] = $logged['id'];
>     $returnarray['call_id'] = $logged['call_id'];
> }
> ```
> 
> With this the `$dbRollBackOnError` change isn't needed. Note `Call::getNextCallId()` reads via `global $db` (read-only, fine); only the `create()` insert goes through `$dbhistory`.
> 
> Tested on a real instance (SuperPDP, test mode): a 404 call inside `$db->begin()` ... `$db->rollback()` now leaves its `llx_einvoicing_call` row (status=0), where before it was dropped:
> 
> ```
> callApi -> status_code=404  call_id=Call-000068
> Rows[marker]: before=0 after=1, last status=0
> PASS: call log survived the rollback
> ```
> 
> phpcs-clean against the repo ruleset. Can open a separate PR if you prefer.

Yes please, can you open a new PR ?

eldy:
Replaced by #310

PR #304, opened 2026-06-26
Modify supplier invoice check for existing entries

defrance:
Updated the condition to check for any existing rows instead of exactly one, and commented out the duplicate entry exception handling.


einvoicing/class/helpers/SupplierInvoiceHelper.class.php
```diff
@@ -314,16 +315,17 @@ public static function isEInvoice(int $supplierInvoiceId, bool $checkLinkedDolOb
...
 				if ($checkLinkedDolObjectExistance) {
 					$factureFournisseur = new FactureFournisseur($db);
 					if ($factureFournisseur->fetch((int) $supplierInvoiceId) > 0) {
 						return true;
 					}
 				}
-			} elseif ($db->num_rows($resql) > 1) {
-				throw new Exception('Duplicate entry in einvoicing_document for supplier invoice with id '.$supplierInvoiceId);
 			}
+			//elseif ($db->num_rows($resql) > 1) {
```
eldy, line 326: Why removing this ? If we have this, there is a critical trouble that need to be solved manually (like the same invoice sent on 2 differents AP, with different status).

defrance: Hello Laurent,
This function is called in two places within the module's trigger:
`BILL_SUPPLIER_VALIDATE` and `DOCUMENT_DELETE`.
The initial query suggests that duplicates cannot exist in the `einvoicing_document` table; however, I encountered this scenario during real-world testing. Here is the case:
- Receipt of a supplier order containing a non-existent product, with the automatic creation option disabled.
- Validation of the supplier invoice causes a blockage.

Looking at the function name and what it does—`isEinvoice`—returning `false` simply because there are multiple rows in that table is functionally incorrect.

defrance (commented):
respobnse

daGrumpf-bxp:
Hi @defrance, hi @eldy 👋

I think you're **both right**, and the two points are actually separable — which is good news, because we can satisfy both.

**1. defrance is right on the semantics.** `isEInvoice()` is a boolean predicate — its job is to answer *"is this a supplier e-invoice?"*. Throwing from it (and returning `false` when `num_rows > 1`) is functionally incorrect: the answer is still *yes*. And since this is called during `BILL_SUPPLIER_VALIDATE`, the exception **blocks the invoice validation for the end user** — exactly the real-world case defrance hit (supplier order with a non-existent product + auto-creation disabled).

**2. eldy is right on the risk.** Duplicates in `einvoicing_document` *are* a genuine data-integrity problem (e.g. the same invoice on two access points with diverging statuses) and must not be silently swallowed.

The fix keeps the predicate clean **and** keeps eldy's safety net — by surfacing the duplicate in the logs instead of throwing:

```php
$resql = $db->query($sql);
if ($resql) {
    $num = $db->num_rows($resql);
    if ($num > 1) {
        // Read-only predicate called during BILL_SUPPLIER_VALIDATE: do NOT throw
        // (it would block the invoice validation for the end user). Surface the
        // integrity issue in the logs instead, so it can be fixed manually.
        dol_syslog(__METHOD__ . " : duplicate entry in einvoicing_document for supplier invoice with id " . $supplierInvoiceId, LOG_WARNING);
    }
    if ($num > 0) {
        if ($checkLinkedDolObjectExistance) {
            $factureFournisseur = new FactureFournisseur($db);
            return ($factureFournisseur->fetch((int) $supplierInvoiceId) > 0);
        }
        return true;
    }
}
return false;
```

This also fixes a pre-existing latent bug: with `checkLinkedDolObjectExistance = false`, the function used to fall through and return `false` even for a single matching row (the `return true` was nested inside the `if`).

A couple of small cleanups while we're here: this replaces the commented-out block with the `dol_syslog` line, the `@throws Exception` can be dropped from the docblock since the function no longer throws, and the stray blank line added in the license header can be reverted to keep the diff minimal.

Happy to push this if it works for both of you 🙂

eldy:
> ```
> checkLinkedDolObjectExistance
> ```

Thanks for analysis @daGrumpf-bxp 
So, you suggest to keep only a log if there is a duplicate.
But if this happen, it means there is a critical trouble that need a database intervention to fix corrupted data.

Here because the function is just to know if it is einvoice or not, you suggest to ignore if there is duplicate. it is true it is enough to say it is an einvoice. But the function is called in a context of sync i think and if you do no through error, it means we will make database changement on a corrupted situation, this may risk to make situation more critical.
Why do we need to ignore the duplicate. This is for me a case of stopping all for analysis (a duplicate should never happen here, if we have it we should force to have critical action done), don't you think so ?

daGrumpf-bxp:
@eldy you are right about the goal, and I checked your assumption in the code - it does not hold, which changes the conclusion. Two points, plus a correction of my own wording from June.

## First, I phrased it badly in June

I wrote "throwing from it (and returning `false` when `num_rows > 1`)". That is inaccurate: the function **throws**, it never returns `false` on a duplicate. Sorry for the confusion.

## 1. `isEInvoice()` is not called from any synchronization context

On `main` it has exactly three callers, all in the trigger:

| Line | Context | What the caller does |
|---|---|---|
| `interface_98_…Triggers.class.php:291` | `BILL_SUPPLIER_VALIDATE` | guard before `checkDolInvoiceAndEInvoiceConsistency()`, which **compares and writes nothing** (0 `INSERT`/`UPDATE`/`DELETE`) |
| `:307` | `BILL_SUPPLIER_DELETE` | **blocks** the deletion of the supplier invoice |
| `:319` | `DOCUMENT_DELETE` | **blocks** the deletion of the linked document |

No write behind any of them, and more importantly: two of the three are **deletion guards**. The synchronization path (`createSupplierInvoiceFromSource()`) does not go through this function. The risk you describe - "we will make database changement on a corrupted situation" - does not exist through this predicate.

## 2. The current `throw` does not produce the clean stop you are after

`Interfaces::run_triggers()` calls `runTrigger()` **without a `try/catch`** (`core/class/interfaces.class.php:193`). An exception raised from `isEInvoice()` therefore propagates uncaught: the user gets a PHP fatal error, not a Dolibarr message. On the two deletion guards the result is worse still: their whole purpose is to say "you cannot delete this", and the user gets a stack trace instead.

In other words, today a duplicate does stop everything - by crashing, without telling anyone what to do.

## What I suggest

Keep your requirement - **nothing proceeds on corrupted data, and the user must know a database intervention is needed** - but get it cleanly:

- the predicate answers the question it is asked, and reports the duplicate to the caller (by-reference parameter, or a `countEInvoiceRows()` next to it);
- each of the three callers, on a duplicate, pushes a translated message such as "duplicate entry in `einvoicing_document` for supplier invoice *id*, a database fix is required" and does `return -1`.

`return -1` from a trigger is the way Dolibarr provides to refuse the operation: the validation or the deletion is blocked, the transaction is rolled back, and the user reads an actionable message instead of a trace. Nothing is lost in severity, the diagnosis is gained.

If that works for you, I will push this version - and I can add at the same time the fix for the latent bug I mentioned (with `checkLinkedDolObjectExistance = false`, the `return true` is nested inside the wrong `if`, so a single matching row still returns `false`).

@defrance this also answers your use case: the validation is no longer blocked by an exception, only by a real duplicate, with a message explaining what to do.

eldy:
> If that works for you, I will push this version - and I can add at the same time the fix for the latent bug I mentioned (with `checkLinkedDolObjectExistance = false`, the `return true` is nested inside the wrong `if`, so a single matching row still returns `false`).

Yes, it's ok for me.
PS: We will release a first version, in current state of code, because we need users to start try to install of production env.
Unit tests can now all be run and are green with AllTests.php (that as a condition to make a first release)

daGrumpf-bxp:
Thanks @eldy - pushed as #503, since I cannot push to `defrance:patch-3`. Close this one in favour of it or keep it open, as you prefer; @defrance is credited in the commit and in the description, the diagnosis is his.

Summary of what is in #503: the predicate answers the question and reports the duplicate through a by-reference flag, each of the three callers refuses with a translated message and `return -1`. Nothing proceeds on corrupted data, and the user reads what to do instead of getting an uncaught fatal.

Measured on Dolibarr 18.0.10 and 20.0.5, identical on both:

| einvoicing_document rows | `isEInvoice()` | `$duplicate` |
|---|---|---|
| none | `false` | `false` |
| one | `true` | `false` |
| two | `true` | **`true`** |

Three phpunit tests added for those three states; full suite green on both versions (**39 tests, 205 assertions**), which should keep your release condition satisfied.

**One correction I owe you.** I wrote twice that this would also fix a pre-existing latent bug, "with `checkLinkedDolObjectExistance = false` the `return true` is nested inside the wrong `if`, so a single matching row still returns `false`". I checked before implementing: **that bug is not on `main`** - the current code does carry the `else { return true; }`. It would have been introduced by the diff of this PR, which removes that `else`. Sorry for the wrong claim; #503 does not have it and a test pins the case.

**About the red CI here**, in case it is read as a reason to reject the change: the `Run phan` failure is unrelated to the diff. The run dates from 2026-06-29 and fails on `PhanUndeclaredConstantOfClass: \CommonObject::CLOSECODE_BANKCHARGE` and `CLOSECODE_WITHHOLDINGTAX` in the trigger, which has since been fixed on `main` - phan is clean on today's `main`. The branch is also four weeks behind, which is where the conflict comes from: `SupplierInvoiceHelper.class.php` moved a lot in the meantime. #503 is based on current `main`.

eldy:
Replaced with #503

PR #315, opened 2026-06-29
Fix #303: [einvoicing] link inbound supplier invoice to its purchase order (BT-13)

### LINKED ISSUE #303: [e-invoice] When a customer reference exists in the e-invoince, link it to the order if it matchs
# [einvoice] Link an inbound supplier invoice to its purchase order when the order reference matches

## Context

French e-invoicing reform (*réforme de la facturation électronique* / RFE).

When an inbound supplier invoice (Factur-X / CII) is processed, the module resolves the supplier and creates the supplier invoice in Dolibarr. Many invoices carry the **purchase order reference** (`BT-13`) — the reference of the order the buyer (the customer receiving the invoice) originally placed, echoed back by the supplier on the invoice.

Today this reference is not used to reconcile the invoice with the corresponding Dolibarr purchase order (*commande fournisseur*), so the link has to be created manually.

## User story

```gherkin
As a customer
When I receive a supplier invoice
And the invoice carries a purchase order reference (BT-13)
And a purchase order with that reference exists for the same supplier
Then the invoice should be automatically linked to that order
And otherwise the invoice should be created without a link
```

## Decision summary

If an order number is provided:

- ✅ and a corresponding order exists **for the correct supplier** → link the two
- ❌ otherwise → the invoice is created **without** a link

## Steps to reproduce

1. Have a Dolibarr purchase order (*commande fournisseur*) for a given supplier, e.g. ref `CF2406-0042`.
2. Receive a supplier invoice (CII / Factur-X) from that supplier carrying the same reference in `BT-13`.
3. Observe that the created supplier invoice is **not** linked to the existing purchase order.

## Current behaviour

The supplier invoice is created without any link to the purchase order, even when the order reference present in the invoice matches an existing order for the same supplier.

Relevant XML field:

```xml
<ram:ApplicableHeaderTradeAgreement>
  ...
  <ram:BuyerOrderReferencedDocument>
    <ram:IssuerAssignedID>CF2406-0042</ram:IssuerAssignedID>   <!-- BT-13 Purchase order reference -->
  </ram:BuyerOrderReferencedDocument>
</ram:ApplicableHeaderTradeAgreement>
```

## Expected behaviour

On import, when the invoice carries an order reference (`BT-13`):

- If **exactly one** purchase order with that reference exists **for the same (resolved) supplier** → automatically create the link between the supplier invoice and that purchase order.
- Otherwise → create the invoice **without** a link (unchanged behaviour).

The linking must never block import.

## Edge cases to handle

- **No order reference in the invoice** → no link (unchanged).
- **Order reference present but no matching order** → no link (optionally an informational notice).
- **Matching reference but different supplier** → do **not** link, to avoid attaching the invoice to the wrong order; optionally surface a warning.
- **Several matching orders** → do not auto-link (ambiguous); optionally flag for manual resolution.

## Normative note

`BT-13` (*Purchase order reference*) is the standard EN 16931 field used to carry the buyer's order reference, transported in the French CTC / AFNOR formats like any other business term. The specifications **define and transport** this reference, but they **do not prescribe** how a receiving solution reconciles it with internal orders.

This feature is therefore **internal ERP reconciliation logic, consistent with — but not mandated by — the AFNOR (XP Z12-012/013/014) or AIFE/DGFiP specifications**. (I can add precise references if a normative citation is wanted in the issue.)

## Suggested implementation notes

- Extract the order reference from `ApplicableHeaderTradeAgreement/BuyerOrderReferencedDocument/ram:IssuerAssignedID` (`BT-13`).
- Look up a Dolibarr supplier order (`CommandeFournisseur`) whose `ref` matches, **restricted to the resolved supplier** (`socid`), to enforce the "correct supplier" condition.
- On a single unambiguous match, create the element link between the supplier invoice and the purchase order (Dolibarr object-link mechanism, `facture_fournisseur` ↔ `commande_fournisseur`).
- Keep the match supplier-scoped and reference-exact (after trimming whitespace) to avoid false positives.
- Secondary references that could be considered later: `BT-14` (*Sales order reference*, `SellerOrderReferencedDocument`) and `BT-10` (*Buyer reference*, `ram:BuyerReference`) — but `BT-13` is the natural key for the customer's own purchase order.

## Environment

- Dolibarr community module: `einvoice`
- Context: French e-invoicing reform (RFE), inbound supplier invoice in CII / Factur-X (profile EN 16931).

daGrumpf-bxp:
## Summary

Closes #303.

When an inbound supplier invoice (CII / Factur-X) carries a **purchase order reference** (`BT-13`, `ApplicableHeaderTradeAgreement/BuyerOrderReferencedDocument/ram:IssuerAssignedID`), it is now automatically linked to the matching Dolibarr purchase order (*commande fournisseur*) at import time — instead of requiring a manual link.

## Implementation

New shared helper `CommonProtocol::_linkSupplierInvoiceToPurchaseOrder($supplierInvoice, $socId, $orderReference)`, called from **both** import paths (`CIIProtocol` and `FacturXProtocol`) right after the supplier invoice is created.

The order lookup is:
- **reference-exact** (after trimming),
- **supplier-scoped** (`fk_soc` = resolved supplier) so a matching reference belonging to another supplier is never linked,
- **entity-aware** (`getEntity('supplier_order')`).

On a single unambiguous match it creates the standard Dolibarr object link (`add_object_linked('order_supplier', $orderId)` → `commande_fournisseur` ↔ `facture_fournisseur`).

## Behaviour (issue edge cases)

| Case | Result |
|---|---|
| Exactly one order with that ref for the supplier | ✅ link created |
| No order reference in the invoice | no link (unchanged, silent) |
| Reference present but no matching order | no link (silent, debug log) |
| Reference matches an order of a **different** supplier | ❌ no link + warning |
| **Several** matching orders | ❌ no link + warning (ambiguous, manual resolution) |

Linking **never blocks import** — it is internal ERP reconciliation, consistent with but not mandated by EN 16931 / the French CTC framework (`BT-13` is transported by the specs, which do not prescribe reconciliation). Outcomes are surfaced via the import log, `dol_syslog`, and `setEventMessages(..., 'warnings')` for the warning cases.

## Translations

New keys added to `en_US` and `fr_FR` only (other languages are Transifex-managed): `EInvoiceSupplierInvoiceLinkedToOrder`, `EInvoiceSupplierOrderLinkAmbiguous`, `EInvoiceSupplierOrderRefWrongSupplier`.

## Notes

- `BT-13` was already parsed by `CIIProtocol::parseInvoiceHeader()` as `orderReference`; Factur-X reuses the same parser, so the field is available on both paths.

## Tests

`php -l` clean on the three modified PHP files; phan (`--quick`, 5.5.2, baseline) reports no new findings.

### DIFF
--- a/einvoicing/langs/fr_FR/einvoicing.lang
+++ b/einvoicing/langs/fr_FR/einvoicing.lang
@@ -331,6 +331,9 @@ SupplierInvoiceComparisonVatRateDifference=différence de montant de TVA %s%% (e
 SupplierInvoiceComparisonVatRateNotFound=taux de TVA à %s%% non trouvé dans la facture Dolibarr
 SupplierInvoiceComparisonSuggestVatCalculationMode=recalculer la TVA en Mode %s pour avoir des montants de TVA correspondant à l'e-facture d'origine
 EinvoicingCantDeleteADocumentLinkedToAnExistingSupplierInvoice=Le document n° %s ne peut pas être supprimé car il est lié à la facture fournisseur Dolibarr n° %s
+EInvoiceSupplierInvoiceLinkedToOrder=La facture fournisseur a été automatiquement liée à la commande fournisseur %s (référence de commande BT-13 correspondante).
+EInvoiceSupplierOrderLinkAmbiguous=Plusieurs commandes fournisseur correspondent à la référence de commande %s pour ce fournisseur ; la facture n'a pas été liée automatiquement. Veuillez la lier manuellement.
+EInvoiceSupplierOrderRefWrongSupplier=Une commande fournisseur portant la référence %s existe mais appartient à un autre fournisseur ; la facture n'a pas été liée.
 
 EINVOICING_CLIENT_ID=Client ID
 EINVOICING_CLIENT_SECRET=Client Secret
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -773,6 +773,13 @@ class CIIProtocol extends AbstractProtocol
 		if ($supplierInvoiceId < 0) {
 			return ['res' => -1, 'message' => 'Invoice creation error: ' . $supplierInvoice->error];
 		} else {
+			// Link the invoice to its purchase order (commande fournisseur) when the order reference
+			// (BT-13) matches a single order for the same supplier. Non-blocking. See issue #303.
+			$orderLinkMessage = $this->_linkSupplierInvoiceToPurchaseOrder($supplierInvoice, $socId, $parsedHeader['orderReference'] ?? '');
+			if ($orderLinkMessage !== '') {
+				$return_messages[] = $orderLinkMessage;
+			}
+
 			$create_deposit_line = 0;
 			$fk_remise_for_deposit = 0;
 			// --------------------------------------------------
--- a/einvoicing/class/protocols/CommonProtocol.class.php
+++ b/einvoicing/class/protocols/CommonProtocol.class.php
@@ -1494,4 +1494,84 @@ trait CommonProtocol
 		}
 		return 0;
 	}
+
+
+	/**
+	 * Link an inbound supplier invoice to its Dolibarr purchase order (commande fournisseur).
+	 *
+	 * Uses the purchase order reference (BT-13, BuyerOrderReferencedDocument/IssuerAssignedID) carried by
+	 * the invoice. The lookup is reference-exact (after trimming) AND scoped to the resolved supplier, so a
+	 * matching reference belonging to another supplier is never linked. The link is only created on a single
+	 * unambiguous match; several matches are flagged (no auto-link) and the absence of a match is silent.
+	 *
+	 * This is internal ERP reconciliation logic and must NEVER block import. See issue #303.
+	 *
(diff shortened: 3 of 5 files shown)


einvoicing/langs/fr_FR/einvoicing.lang
```diff
@@ -331,6 +331,9 @@ SupplierInvoiceComparisonVatRateDifference=différence de montant de TVA %s%% (e
 SupplierInvoiceComparisonVatRateNotFound=taux de TVA à %s%% non trouvé dans la facture Dolibarr
 SupplierInvoiceComparisonSuggestVatCalculationMode=recalculer la TVA en Mode %s pour avoir des montants de TVA correspondant à l'e-facture d'origine
 EinvoicingCantDeleteADocumentLinkedToAnExistingSupplierInvoice=Le document n° %s ne peut pas être supprimé car il est lié à la facture fournisseur Dolibarr n° %s
+EInvoiceSupplierInvoiceLinkedToOrder=La facture fournisseur a été automatiquement liée à la commande fournisseur %s (référence de commande BT-13 correspondante).
```
eldy, line 334: Note: Do not waste your time by filling fr_FR lang file. It is erased by transifex. Only en_USmust be included into PR.

PR #318, opened 2026-06-29
Fix #302: [einvoicing] carry customer reference (Réf. client) into BT-13

### LINKED ISSUE #302: [e-invoicing] Add Ref. customer into the generated invoice
# [einvoice] Add the customer reference (BT-13) into the generated e-invoice

## Context

French e-invoicing reform (*réforme de la facturation électronique* / RFE).

This concerns the **emission** side: when the module **generates** a customer e-invoice (Factur-X / CII) from a Dolibarr customer invoice (*facture client*), the **customer reference** filled on that invoice (Dolibarr field *Réf. client* / `ref_client`) is actually not carried into the generated e-invoice.

Here the *Réf. client* is treated as the **customer's purchase order number**, so it is mapped to `BT-13` (*Purchase order reference*). This is the field the **recipient** reads to reconcile the invoice with their own purchase order — i.e. the emission-side counterpart that makes order auto-linking work on reception.

## User story

```gherkin
As a supplier
When an e-invoice is generated for a customer
And the Dolibarr customer invoice carries a customer reference (Réf. client)
Then that reference should be written into the generated e-invoice as the purchase order reference (BT-13)
```

## Current behaviour

The generated CII / Factur-X document does not include the customer reference. `BT-13` is left empty even when *Réf. client* is filled on the Dolibarr invoice.

## Expected behaviour

When generating the e-invoice, map the Dolibarr *Réf. client* (`ref_client`) into `BT-13` — *Purchase order reference* (`BuyerOrderReferencedDocument/IssuerAssignedID`).

The element should only be emitted when the value is non-empty.

## Field mapping

| Dolibarr field | EN 16931 term | CII path |
|---|---|---|
| *Réf. client* (`ref_client`) | `BT-13` Purchase order reference | `ApplicableHeaderTradeAgreement/ram:BuyerOrderReferencedDocument/ram:IssuerAssignedID` |

Example of the expected output:

```xml
<ram:ApplicableHeaderTradeAgreement>
  ...
  <ram:BuyerOrderReferencedDocument>
    <ram:IssuerAssignedID>CMD-2026-0042</ram:IssuerAssignedID>   <!-- BT-13 Purchase order reference -->
  </ram:BuyerOrderReferencedDocument>
</ram:ApplicableHeaderTradeAgreement>
```

> Note: `BT-10` (*Buyer reference*, `ram:BuyerReference`) was considered as an alternative target, but `BT-13` is chosen here because the *Réf. client* holds the customer's order number, and `BT-13` is the field used by the recipient for invoice ↔ order reconciliation. `BT-10` is intentionally **not** populated to avoid mixing it with the public-sector *code service* / *numéro d'engagement* semantics it carries in B2G.

## Normative note

`BT-13` (*Purchase order reference*) is defined by EN 16931 and carried in the French CTC / AFNOR formats. The specification **defines and transports** the field, but **does not prescribe** which internal ERP field should populate it.

Mapping Dolibarr's *Réf. client* into `BT-13` is therefore an **integration mapping choice consistent with — but not imposed by — the AFNOR (XP Z12-012/013/014) / AIFE-DGFiP specifications**. 

## Suggested implementation notes

- On CII generation, read `ref_client` from the Dolibarr invoice object (`Facture->ref_client`).
- Write it to `BuyerOrderReferencedDocument/ram:IssuerAssignedID` (`BT-13`) when non-empty.
- Do not emit empty tags — only output the element if the value is present, to avoid cardinality / Schematron issues.

## Environment

- Dolibarr community module: `einvoice`
- Context: French e-invoicing reform (RFE), **emission** of a customer invoice in CII / Factur-X (profile EN 16931).

Image

daGrumpf-bxp:
## Summary

Closes #302.

On **emission**, the Dolibarr customer reference (*Réf. client* / `ref_client`, used here as the customer's purchase order number) was not written into the generated CII / Factur-X document, so `BT-13` (*Purchase order reference*) stayed empty even when *Réf. client* was filled.

This is the emission-side counterpart of inbound order auto-linking (#303): once `BT-13` is populated, the recipient can reconcile the invoice with their own purchase order.

## Root cause & fix

Two issues, both addressed:

1. **Wrong source field** — `lib/buildinvoicelines.inc.php` fed the order reference from `$object->ref_customer` (unset on invoices) instead of `$object->ref_client` (the field actually shown/edited on the invoice card). Now reads `ref_client`, keeping the existing priority:
   `d4d_promise_code` (Chorus engagement number) → `ref_client` → first linked customer order. This preserves the B2G/Chorus behaviour and fixes B2B.
2. **CII never emitted the element** — `CIIProtocol` generation did not output `BuyerOrderReferencedDocument` at all (Factur-X already emitted it via the horstoeko builder, so the source fix alone covers Factur-X). Added the element to `ApplicableHeaderTradeAgreement`, positioned right after `BuyerTradeParty` per the CII schema sequence, and only when the value is non-empty (no empty tags / cardinality issues).

## Field mapping

| Dolibarr field | EN 16931 | CII path |
|---|---|---|
| *Réf. client* (`ref_client`) | `BT-13` Purchase order reference | `ApplicableHeaderTradeAgreement/ram:BuyerOrderReferencedDocument/ram:IssuerAssignedID` |

`BT-10` (`BuyerReference`) is intentionally **not** populated, to avoid mixing it with the public-sector *code service* / *numéro d'engagement* semantics it carries in B2G.

## Resulting output

```xml
<ram:ApplicableHeaderTradeAgreement>
  ...
  <ram:BuyerOrderReferencedDocument>
    <ram:IssuerAssignedID>CMD-2026-0042</ram:IssuerAssignedID>
  </ram:BuyerOrderReferencedDocument>
</ram:ApplicableHeaderTradeAgreement>
```

## Tests

`php -l` clean; phan (`--quick`, 5.5.2, baseline) reports no new findings. No language changes (generation-only mapping).

### DIFF
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -81,7 +81,8 @@ if (getDolGlobalInt('EINVOICING_USE_CHORUS')) {
 }
 $promise_code = $object->array_options['options_d4d_promise_code'] ?? '';
 if ($promise_code == '') {
-	$promise_code = $object->ref_customer ?? '';
+	// Dolibarr "Réf. client" (ref_client) holds the customer's purchase order number -> BT-13 (see issue #302)
+	$promise_code = $object->ref_client ?? '';
 }
 if ($promise_code == '' && !empty($customerOrderReferenceList)) {
 	$promise_code = $customerOrderReferenceList[0];
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -81,8 +81,9 @@ if (getDolGlobalInt('EINVOICING_USE_CHORUS')) {
 }
 $promise_code = $object->array_options['options_d4d_promise_code'] ?? '';
 if ($promise_code == '') {
-	// Dolibarr "Réf. client" (ref_client) holds the customer's purchase order number -> BT-13 (see issue #302)
-	$promise_code = $object->ref_client ?? '';
+	// Dolibarr "Réf. client" holds the customer's purchase order number -> BT-13 (see issue #302).
+	// The property is ref_client on recent versions and ref_customer on some older ones; accept both.
+	$promise_code = $object->ref_client ?? ($object->ref_customer ?? '');
 }
 if ($promise_code == '' && !empty($customerOrderReferenceList)) {
 	$promise_code = $customerOrderReferenceList[0];
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -1490,6 +1490,18 @@ class CIIProtocol extends AbstractProtocol
 
 		$this->buildParty($doc, $agreement, $invoiceData, 'buyer');
 
+		// Buyer order reference (BT-13): customer purchase order number (Dolibarr "Réf. client").
+		// Emitted only when non-empty to avoid empty tags / cardinality issues. See issue #302.
+		// Position follows the CII schema sequence (after BuyerTradeParty).
+		if (!empty($invoiceData['orderReference'])) {
+			$comment = $doc->createComment('Buyer order reference (BT-13)');
+			$agreement->appendChild($comment);
+
+			$buyerOrderRef = $doc->createElement('ram:BuyerOrderReferencedDocument');
+			$agreement->appendChild($buyerOrderRef);
+			$buyerOrderRef->appendChild($doc->createElement('ram:IssuerAssignedID', htmlspecialchars($invoiceData['orderReference'])));
+		}
+
 
 		// DELIVERY
 


einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -81,7 +81,8 @@
 }
 $promise_code = $object->array_options['options_d4d_promise_code'] ?? '';
 if ($promise_code == '') {
-	$promise_code = $object->ref_customer ?? '';
+	// Dolibarr "Réf. client" (ref_client) holds the customer's purchase order number -> BT-13 (see issue #302)
+	$promise_code = $object->ref_client ?? '';
```
eldy: On some version of Dolibarr, we may have ref_customer. May be we can have both working with

$promise_code = $object->ref_client ?? ($object->ref_customer ?? '');

daGrumpf-bxp: Done in 4511b4c: `$promise_code = $object->ref_client ?? ($object->ref_customer ?? '');` so BT-13 is populated on versions exposing either property.

PR #321, opened 2026-06-29
Fix #319: [einvoicing] do not generate e-invoices for B2C third parties (private individuals)

### LINKED ISSUE #319: No Einvoicing for private individuals not subject to VAT
The module is already working well, but it presents a stumbling block for private customers.
Companies operating in both B2B and B2C sectors cannot validate orders from private customers (due to a SIRET number error). The same applies to payment processing: transactions are rejected if a SIRET number is missing.
Electronic invoices must not be generated when the customer is not subject to VAT.

daGrumpf-bxp:
Closes #319.

## Problem

B2C transactions (private individuals) are **out of the e-invoicing scope** — they fall under *e-reporting*, not e-invoice transmission to a PA. But `EInvoicing::needEInvoiceManagement()` currently returns `STATUS_NOT_GENERATED` for **every** French customer. So invoices for private individuals get an e-invoice attempt, and the user hits SIRET/routing errors at validation and payment (exactly the report in #319).

## Approach — opt-in, conservative

Default behaviour is **unchanged**. One new option gates the feature:

| Option | Default | Role |
|--------|---------|------|
| `EINVOICING_SKIP_B2C` | `0` (off) | skip e-invoicing for third parties detected as private individuals |

Company vs individual detection is **delegated to `Societe::isACompany()`**, which already has its own options (`tva_intra`, `typent_code`, professional ids, `MAIN_UNKNOWN_CUSTOMERS_ARE_COMPANIES`) — so we don't duplicate or hardcode that logic here. When the option is off, nothing changes.

```php
if ($return == self::STATUS_NOT_GENERATED && getDolGlobalInt('EINVOICING_SKIP_B2C')) {
    if (!$object->thirdparty->isACompany()) {
        $return = self::STATUS_IGNORE;
    }
}
```

The existing hard validation behaviour (SIRET/routing) is **left untouched**.

## Changes
- `class/einvoicing.class.php` — opt-in B2C skip in `needEInvoiceManagement()`, gated on `isACompany()`
- `admin/setup_options.php` — the `EINVOICING_SKIP_B2C` toggle
- `langs/en_US` — label/help

## Notes
- `dolics` (phpcs, repo ruleset) clean; `php -l` clean.


einvoicing/class/einvoicing.class.php
```diff
@@ -2685,6 +2685,26 @@ public function needEInvoiceManagement($object)
...
 			$return = self::STATUS_NOT_GENERATED;
 		}
+
+		// B2C (private individuals) is out of the e-invoicing scope: it falls under e-reporting, not e-invoice
+		// transmission. Opt-in (off by default): when EINVOICING_SKIP_B2C is set, skip e-invoicing for third
+		// parties whose entity type (typent_code) is in EINVOICING_B2C_TYPENT_CODES (CSV, default 'TE_PRIVATE').
+		// We only skip on an explicit positive marker, never on missing data (e.g. an empty SIREN), so a company
+		// that is not yet fully filled in is still handled rather than silently ignored.
+		if ($return == self::STATUS_NOT_GENERATED && getDolGlobalInt('EINVOICING_SKIP_B2C')) {
+			$b2cTypes = array_filter(array_map('trim', explode(',', getDolGlobalString('EINVOICING_B2C_TYPENT_CODES', 'TE_PRIVATE'))));
```
eldy: I recommand to replace all this block of test with 
`$object->thirdparty->isACompany()`
So we can drop also option EINVOICING_B2C_TYPENT_CODES because we already have options into this method to control how we want to detect individual vs companies.

daGrumpf-bxp: Done in e322047. Replaced the whole typent_code block with `$object->thirdparty->isACompany()` and dropped the `EINVOICING_B2C_TYPENT_CODES` option (global, setup field and lang keys). The `EINVOICING_SKIP_B2C` toggle is kept; when on, an invoice to a thirdparty that `isACompany()` reports as an individual is ignored. Also dropped the fr_FR additions (en_US only, per your note on #315).

PR #322, opened 2026-06-29
Fix #226: [einvoicing] isolate inbound temp files so concurrent syncs cannot corrupt invoices

### LINKED ISSUE #226: [EInvoicing] Management of temp files may result in invalid invoice data
### Description
There might be some conflicts in tmp files management.

By example, for facturX and supplier invoices : 
- when synchronizing with AP, each synchFlow store file content (got from API) into a tmp file
- the name of the tmp file is `facturx.pdf`
- which is stored by default in `documents/pdpconnect/temp`

But what if 2 synchronisations were launched simultaneously (by 2 different users) ? It may lead to conflits of the tmp file because the 2 processes will try to use same target tmp file. And if the tmp file was replaced by the other one during syncFlow process, it will result in wrong invoice data in the tmp file, and then wrong data in Dolibarr invoice.

### Suggestions
- Add a uniq ids to the tmp files name where the uniq id will be the hash result of the file (using sha256 or random_bytes+bin2hex)
- Review tmp files deletion : 
  - ensure to delete the tmp file when it is no longer needed : when it is added to the invoice as attachment or if an error occurred during process. Currently, the code deletes temp dir just before to put new content and in a specific case, but does not take all cases in account) :
https://github.com/Dolibarr/dolibarr-community-modules/blob/def68fc354ebc683686f8cdd99ed93055bcd07e5/pdpconnectfr/class/protocols/FacturXProtocol.class.php#L1051-L1073

  - on document_list.php : syncFlows() is executed, then if it fails, we delete tmp files. This should be done in syncFlows() or one of its submethods, and document_list.php does not have to manage the tmp files deletion (and of each format : CII, FacturX...) :
https://github.com/Dolibarr/dolibarr-community-modules/blob/def68fc354ebc683686f8cdd99ed93055bcd07e5/pdpconnectfr/document_list.php#L253-L254
https://github.com/Dolibarr/dolibarr-community-modules/blob/def68fc354ebc683686f8cdd99ed93055bcd07e5/pdpconnectfr/document_list.php#L315-L343

  - same risk of taking the wrong tmp file with cdar generation in sendStatusMessage() of providers : 
https://github.com/Dolibarr/dolibarr-community-modules/blob/def68fc354ebc683686f8cdd99ed93055bcd07e5/pdpconnectfr/class/utils/CdarHandler.class.php#L316
https://github.com/Dolibarr/dolibarr-community-modules/blob/def68fc354ebc683686f8cdd99ed93055bcd07e5/pdpconnectfr/class/providers/EsalinkPDPProvider.class.php#L1576-L1600

- in some parts, it may also be possible to not use temp files and keep currently process data in memory (instead of save them into files and just after read them from files) which could simplify code

daGrumpf-bxp:
Addresses #226. Opened as **draft**: needs a reception test on a real sandbox flow before review.

## Problem

Inbound supplier-invoice reception wrote each received document into a **shared** module temp dir (`documents/einvoicing/temp/`) under a **fixed** name (`facturx.pdf` / `einvoice.xml`), **wiped the whole dir** before each write, then **read that same fixed file back** to parse the invoice.

Two concurrent syncs — two users clicking *Synchronize*, or the `cronSyncFlows` cron overlapping a manual sync — race: process B's wipe deletes process A's file between A's write and A's read, so A parses **B's document** and creates a supplier invoice with the **wrong supplier/amount**. Silent data corruption, no error raised. Same fixed-name hazard in `CdarHandler` (`cdar_<condition>.xml`).

## A — kill the race

- `FacturXProtocol` / `CIIProtocol::createSupplierInvoiceFromSource()` write to a **unique per-call** working file (`in_<random>.{pdf,xml}`). The 650-line body moved to a private `doCreateSupplierInvoiceFromSource()` receiving the paths; the public method owns the temp lifecycle via **`try/finally`** (robust against every early return and exceptions).
- The fixed `facturx.pdf` / `einvoice.xml` slot is now **only** the downloadable *"last invoice that could not be processed"* diagnostic shown in the list view: on failure the working file is promoted to it, on success the working files are removed (`AbstractProtocol::cleanupIncomingTempFiles()`).
- `CdarHandler::generateCdarFile()` uses a unique name; `EsalinkPDPProvider` deletes the CDAR temp right after the upload (its only consumer).

## B — separation of concerns

- `syncFlows()` clears the diagnostic slot **once at the start** of a run (`AbstractPDPProvider::clearIncomingDiagnosticFiles()`); failed flows re-create it.
- `document_list.php` no longer deletes temp files — it only displays the diagnostic download links.

## Behaviour

No change on the happy path. The *"last unprocessed invoice"* diagnostic + download links in the document list are preserved (now driven by the sync, not the list controller).

## Validation
- `dolics` (phpcs, repo ruleset) clean; `php -l` clean on all 8 files.
- ⚠️ **Not yet runtime-tested on a reception flow** — to validate on a sandbox PA before un-drafting.

## Not in scope (possible follow-up)
- C: parse fully in-memory (`readAndGuessFromContent` / `getInvoiceDocumentContentFromContent` exist; CII already parses from a string) to drop the working file for the happy path entirely. The diagnostic file must stay on disk for the UI download, so a temp never fully disappears.

### DIFF
--- a/einvoicing/class/protocols/AbstractProtocol.class.php
+++ b/einvoicing/class/protocols/AbstractProtocol.class.php
@@ -125,4 +125,45 @@ abstract class AbstractProtocol
 			dol_syslog(get_class($this) . '::checkFileSizeLimit ' . basename($filepath) . ' size ' . number_format($sizeMB, 2) . ' MB exceeds configured limit of ' . number_format($maxMB, 2) . ' MB', LOG_WARNING);
 		}
 	}
+
+	/**
+	 * Clean up the per-call working temp files of an inbound invoice, while preserving the
+	 * "last invoice that could not be processed" diagnostic shown (and downloadable) in the
+	 * document list view.
+	 *
+	 * Each inbound sync writes the received document to its own unique working file, so two
+	 * concurrent syncs can no longer overwrite each other and parse the wrong invoice (#226).
+	 * On failure, the working file is promoted to the fixed diagnostic slot (overwriting the
+	 * previous one) so it stays downloadable; on success, the working files are simply removed.
+	 *
+	 * @param	string	$tempDir			Module temp directory
+	 * @param	string	$workFile			Unique working file for the received document
+	 * @param	string	$workReadable		Unique working file for the readable view (may not exist)
+	 * @param	string	$diagName			Fixed diagnostic filename for the received document
+	 * @param	string	$diagReadableName	Fixed diagnostic filename for the readable view
+	 * @param	bool	$failed				True if processing failed (keep the diagnostic), false otherwise
+	 * @return	void
+	 */
+	protected function cleanupIncomingTempFiles($tempDir, $workFile, $workReadable, $diagName, $diagReadableName, $failed)
+	{
+		if ($failed) {
+			// Keep the failed file(s) as the downloadable "last unprocessed invoice" diagnostic.
+			if (file_exists($workFile)) {
+				dol_delete_file($tempDir . '/' . $diagName);
+				dol_copy($workFile, $tempDir . '/' . $diagName, '0', 1);
+			}
+			if (file_exists($workReadable)) {
+				dol_delete_file($tempDir . '/' . $diagReadableName);
+				dol_copy($workReadable, $tempDir . '/' . $diagReadableName, '0', 1);
+			}
+		}
+
+		// Always drop the per-call working files.
+		if (file_exists($workFile)) {
+			dol_delete_file($workFile);
+		}
+		if (file_exists($workReadable)) {
+			dol_delete_file($workReadable);
+		}
+	}
 }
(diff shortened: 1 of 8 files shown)


daGrumpf-bxp:
Validated on a sandbox instance (Dolibarr, PHP 8.4, CII protocol):

- **Failure path**: invalid payload → `res=-1`, unique working file cleaned up (0 leftover), diagnostic `einvoice.xml` written as expected.
- **Happy path**: a real received CII created the supplier invoice end-to-end (thirdparty + product + attachment), temp dir left clean, no diagnostic on success.
- **Crash safety**: a fatal mid-processing still cleaned up the working file (try/finally runs).
- **Concurrency** (the actual bug): 8 receptions launched in parallel → 0 `in_*` leftovers, no collision, a single diagnostic file. With the previous shared fixed-name + wipe-all logic this would have corrupted invoices.

FacturX path is the same wrapper/lifecycle pattern as CII (validated); CDAR now uses a unique name too.

PR #323, opened 2026-06-30
needEInvoiceManagement set to ignore for TE_PRIVATE

frederic34:


daGrumpf-bxp:
Nice, same idea as #321 👍

Just for the record: eldy mentioned he'd rather rely on `Societe::isACompany()` than on a single hardcoded `typent_code` — it already covers `tva_intra`, professional ids, `MAIN_UNKNOWN_CUSTOMERS_ARE_COMPANIES`, etc., so a private individual without `TE_PRIVATE` set is still caught.

We took that route in #321 (also kept it opt-in via `EINVOICING_SKIP_B2C` so the default behaviour stays unchanged). Whichever way you go, happy to converge — feel free to reuse anything from there. 🙂

PR #329, opened 2026-07-03
fix(einvoicing): Fix Phan in einvoicing.class.php

mdeweerd:
# fix(einvoicing): Fix Phan in einvoicing.class.php

- Update einvoicing.class.php to fix type mismatches,
- Update baseline.txt.


mdeweerd:
@eldy - conflict fixed.

PR #346, opened 2026-07-04
fix(einvoicing): report fr:212 (paid) on any real payment, not only bank-charge/withholding closings

GregcodeOhm:
## What

Report the AFNOR lifecycle status fr:212 (Encaissee / paid) to the Access Point whenever a transmitted customer invoice is actually paid, instead of only for two rare manual closing reasons.

## Why

On `BILL_PAYED`, the trigger only sent fr:212 when `close_code` was `CLOSECODE_BANKCHARGE` or `CLOSECODE_WITHHOLDINGTAX`. A normal full payment leaves `close_code` empty, so the common case (an invoice paid by one or more payments reaching the total) never reported its payment status to the platform, even though payment reporting is a lifecycle obligation of the reform.

## Fix

Key the fr:212 report on the amount actually received (`$object->getSommePaiement() > 0`) rather than on `close_code`. This:

- covers the common case (normal full payment, `close_code` empty), which was missed;
- still covers the two previous cases (bank charge / withholding tax closings leave a received amount);
- does not fire for a write-off with no payment (abandon / bad debt), where `BILL_PAYED` also fires but nothing was received, so "cashed in" would be wrong.

The existing guards are kept: the module must have Dolibarr-to-AP sync on, the invoice must need e-invoicing, and it must already be transmitted.

## Verification (dolibarr-test, SuperPDP sandbox)

Decision table over the transmitted invoices, old vs new logic:

```
ref           statut  sumpayed  close_code   old   new
FA2512-0797   paid     384.00    (empty)     skip  SEND   <- the fix: a paid transmitted invoice now reports fr:212
FA2607-0802   valid      0.00    (empty)     skip  skip
FA2607-0801   valid      0.00    (empty)     skip  skip
```

The fr:212 CDAR generated by the module is accepted (parsed) by SuperPDP: a live `sendStatusMessage($invoice, 212)` returned a business-level response from the platform (HTTP 400 "the invoice already has a final status", because that particular test invoice had previously been rejected), which confirms the CDAR format is valid and the only reason it was declined is the invoice state, not the message. On an invoice that is live at the platform, the fr:212 is accepted.

`php -l` clean.

## Scope / notes

- fr:212 for customer invoices only. fr:211 (payment transmitted, supplier side) is intentionally out of scope: `sendStatusMessage()` currently lists `STATUS_PAYMENT_SENT` among the statuses "not supported for now" (see `getEinvoiceStatusOptions()`), so enabling it is a separate change.
- If an invoice is already in a final state at the platform (for example previously rejected), the platform declines the fr:212 with a clear message; that is expected (the invoice should be corrected and resent first).
- This does not add the optional payment amount blocks (MEN, BR-FR-CDV-14) to the CDAR; the module builds a minimal CDAR by design, and the status is accepted without them. Amounts can be a follow-up if the platform later requires them.

## Files

- `einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php`

### DIFF
--- a/einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php
+++ b/einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php
@@ -231,9 +231,12 @@ class InterfaceEInvoicingTriggers extends DolibarrTriggers
 			// Check if the invoice is transmitted to EInvoicing and confirm we have received the payment if yes.
 
 			if (!getDolGlobalString('EINVOICING_DISABLE_SYNC_DOLI_TO_AP')) {		// If sync Dolibarr to AP is on
-				// Test on $close_code that are closing code to say we really accept the payment
-				if ($object->close_code == $object::CLOSECODE_BANKCHARGE ||
-					$object->close_code == $object::CLOSECODE_WITHHOLDINGTAX) {
+				// fr:212 (Encaissee) means the invoice was cashed in: report it whenever an actual payment was
+				// received. A normal full payment leaves close_code empty, so the previous test on close_code
+				// (BANKCHARGE/WITHHOLDINGTAX only) missed the common case. Keying on the received amount also
+				// avoids reporting "cashed in" for a write-off with no payment (abandon / bad debt), where
+				// BILL_PAYED still fires but nothing was received.
+				if ($object->getSommePaiement() > 0) {
 					// Test if invoice need EInvoicing
 					$einvoicing = new EInvoicing($this->db);
 


einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php
```diff
@@ -231,9 +231,12 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
 			if (!getDolGlobalString('EINVOICING_DISABLE_SYNC_DOLI_TO_AP')) {		// If sync Dolibarr to AP is on
-				// Test on $close_code that are closing code to say we really accept the payment
-				if ($object->close_code == $object::CLOSECODE_BANKCHARGE ||
-					$object->close_code == $object::CLOSECODE_WITHHOLDINGTAX) {
+				// fr:212 (Encaissee) means the invoice was cashed in: report it whenever an actual payment was
+				// received. A normal full payment leaves close_code empty, so the previous test on close_code
+				// (BANKCHARGE/WITHHOLDINGTAX only) missed the common case. Keying on the received amount also
+				// avoids reporting "cashed in" for a write-off with no payment (abandon / bad debt), where
+				// BILL_PAYED still fires but nothing was received.
+				if ($object->getSommePaiement() > 0) {
```
eldy, line 239: And what about if invoice was paid partially and then abandoned by customer (bad customer thatdisappear after partial payment) ?
Do we need to send 212 too ?

GregcodeOhm: Hi, Merci. 
Good edge case. 
Yes, we should still report fr:212 there, and the current condition (getSommePaiement() > 0) already handles it correctly.

First, the lifecycle mechanics: BILL_PAYED is only fired by Facture::setPaid(), which Dolibarr calls in two cases only, the invoice classified as fully paid (paye = 1, empty close_code), or the invoice closed manually with a close_code (abandoned, bad debt, discount). Recording a plain partial payment does not call setPaid(), so BILL_PAYED does not fire and no fr:212 is sent while the invoice is still open. The getSommePaiement() > 0 gate is therefore only ever evaluated at the moment the invoice is settled or closed.

That is exactly where it does the right thing: fr:212 (Encaissee) reports that money was actually cashed in, which is a reporting obligation of the reform (VAT on collection: the encaissement is the taxable event for services). A partial payment that was received is a real encaissement and stays cashed even if the customer later abandons the balance; the abandonment only concerns the unpaid remainder (a write-off), it does not cancel the duty to report the amount actually collected.

So at close time:

closed with a real amount received (full payment, or partial then abandoned): getSommePaiement() > 0, report fr:212. Correct.
closed with nothing received (pure write-off): getSommePaiement() == 0, do not report. Correct.
Keying on close_code would have inverted this: it missed the common full-payment case (empty close_code) and could not tell a real partial encaissement from a pure write-off.

Optional refinement I can add if you want: make the amount carried by the 212 message explicit (the sum actually received), so a partial encaissement is not read as a full settlement.

Tell me if i'm wrong.
Best

PR #347, opened 2026-07-04
feat(einvoicing): local EN 16931 business rules check on the generated XML

GregcodeOhm:
## What

Local EN 16931 business rules check on the generated CII XML, run at generation time for any PDP provider: a focused subset of the official rules (totals arithmetic, VAT breakdown, line rules, seller SIREN format) verified in PHP with explicit messages quoting the official rule id.

## Why

Today the only business rules verification happens at the Approved Platform (validation_reports for providers that expose a validator, or the rejection itself). Arithmetic inconsistencies produced by a generation bug are discovered late, with a cryptic platform message, and only for providers that have a validator.

A concrete example: the prepaid double count fixed by #341 produced `DuePayableAmount = -384.00` (BR-CO-16 violation). Nothing local caught it; it surfaced only at the platform. With this check the generation itself reports:

```
BR-CO-16: DuePayableAmount is negative (-384.00); TotalPrepaidAmount (768.00) exceeds GrandTotal (384.00)
```

This is not a replacement for the official Schematron (which stays the reference at the platform): it is a fast local safety net, without a network call, that works for every provider.

## What it does

- New class `class/utils/En16931Validator.class.php`: pure DOMXPath validator over the CII XML. Implemented rules: BR-16 (at least one line), BR-27 (no negative net price), BR-CO-10 to BR-CO-17 (sums of lines, allowances, charges, tax basis, VAT breakdown consistency, grand total, prepaid/due), negative DuePayableAmount detection (skipped for credit notes 381), VAT category/rate coherence (S with non-positive rate, zero-rated categories Z/E/AE/K/G/O with a non-zero rate) at line and breakdown level, BR-FR-10 (seller legal registration id with scheme 0002 must be a 9 digit SIREN). Amount comparisons use a one-cent rounding tolerance. `TaxTotalAmount` is matched on the invoice currency when several currencies are present.
- New protected helper `AbstractProtocol::checkBusinessRules($xmlcontent)` (same pattern as the existing `checkFileSizeLimit()`): runs the validator, logs, and either appends warnings or, in strict mode, appends errors and throws (the exception is already caught by both `generateInvoice()` implementations, which return -1 with the messages).
- Called at the two places where the final CII XML is produced: `CIIProtocol::generateXML()` and `FacturXProtocol::generateXML()`.
- Two options in the setup: `EINVOICING_BR_CHECK` (on by default, warnings only, so no behavior change for existing installs beyond visible warnings) and `EINVOICING_BR_CHECK_STRICT` (off by default, aborts the generation on violation).

## Verification (dolibarr-test 22.0, SuperPDP sandbox instance)

- Real generated invoice XML (FA2511-0796_cii.xml): 0 violations (no false positive), end-to-end `generateInvoice()` unchanged (file produced, no warnings).
- Tampered variants each caught with the right rule id:
  - prepaid double count replica (due -384, prepaid 768): BR-CO-16 (twice: equation and negativity)
  - tampered LineTotalAmount: BR-CO-10
  - negative net price: BR-27
  - 8 digit seller SIREN: BR-FR-10
  - tampered VAT CalculatedAmount: BR-CO-17 and BR-CO-14
  - truncated XML: well-formedness error
- Strict mode: exception thrown with the messages, `errors[]` filled, `generateInvoice()` returns -1; valid XML raises nothing; `EINVOICING_BR_CHECK=0` disables everything.
- `php -l` clean on all files.

## Notes

- Messages are technical by design (rule id plus amounts), like a validator report; they are not translated per rule.
- The subset was chosen for yield: totals arithmetic is where generation bugs materialize. More rules (BR-FR mentions, VATEX details) can be added incrementally to the same class.
- Independent from the other open PRs (#344/#345/#346): no common file with #346; the setup_options.php and lang additions are placed next to the existing validation option, in a different hunk than #344/#345, so merges stay clean in any order.

## Files

- `einvoicing/class/utils/En16931Validator.class.php` (new)
- `einvoicing/class/protocols/AbstractProtocol.class.php`: `checkBusinessRules()` helper
- `einvoicing/class/protocols/CIIProtocol.class.php`: call after buildXML
- `einvoicing/class/protocols/FacturXProtocol.class.php`: call after buildXML
- `einvoicing/admin/setup_options.php`: the two toggles
- `einvoicing/langs/en_US/einvoicing.lang`: strings


einvoicing/class/protocols/FacturXProtocol.class.php
```diff
@@ -139,6 +139,10 @@ public function generateXML($invoice, $outputlangs = null)
 			dol_delete_file($xmlfile);
 
 			$xmlcontent = $CII->buildXML($invoiceData, $linesData, 'EXTENDED', $outputlangs);
+
+			// Local EN 16931 business rules safety net (warnings, or abort in strict mode)
+			$this->checkBusinessRules($xmlcontent);
```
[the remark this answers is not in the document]
GregcodeOhm: Fixed in d5d265ee07b7ec5e96cc7e378d822022b8452aaf: checkBusinessRules() is now also called in the external Factur-X builder branch (EINVOICING_USE_EXTERNAL_FACTURX_BUILDER), on the final XML after the CTC-FR patch, right before the file is returned. Strict mode now blocks invalid XML on both branches. Same CII/ram namespaces, so the validator applies to the external builder output too.

einvoicing/admin/setup_options.php
```diff
@@ -186,6 +186,19 @@
 	$item->defaultFieldValue = 0;
 	$item->cssClass = 'minwidth500';
 
+	// Local EN 16931 business rules check (BR, BR-CO, BR-FR subset) on the generated XML.
+	// On by default as non-blocking warnings; strict mode below turns violations into a
+	// generation error. The official Schematron of the Approved Platform stays the reference.
+	$item = $formSetup->newItem('EINVOICING_BR_CHECK')->setAsYesNo();
+	$item->helpText = $langs->transnoentities('EINVOICING_BR_CHECK_HELP');
```
eldy, line 198: Option EINVOICING_BR_CHECK and EINVOICING_BR_CHECK_STRICT are colliding. Having the _CHECK to no and the strict to yes seems to have no sense. 
This means it should be one same option but with 3 possible value:
"nocheck"
"warning_only"
"blocking"

Can you merge the 2 options if I am right ?

eldy (changes requested):

einvoicing/admin/setup_options.php
GregcodeOhm: Done in a280353f. Merged into a single EINVOICING_BR_CHECK select with three values: nocheck, warning_only (default) and blocking. checkBusinessRules() reads that one value (nocheck skips, warning_only reports non-blocking warnings, blocking aborts the generation). You were right, the on/off plus strict on/off pair allowed the contradictory "no check but strict" state.

I also rebased the branch on current main and squashed to a single commit. Two nearby directory-option defaults (EINVOICING_PRECHECK_DIRECTORY and EINVOICING_REQUIRE_ROUTABLE_RECIPIENT) are now typed as strings to keep phan green, since the added block shifted their lines past the line-based baseline. All checks pass.

PR #349, opened 2026-07-05
Qual(einvoicing): Fix Phan notices in FacturXProtocol class

mdeweerd:
# Qual(einvoicing): Fix Phan notices in FacturXProtocol class

- Fixed multiple typing issues in the FacturXProtocol class
- Modified parent class to include $db property (also fixes some notices in other classes)
- Added casts to the buildinvoicelines.inc.php file
- Added stub for included symphony libraries that are reference (to correct false positives)


mdeweerd:
Note: Fixing the Phan notices does not imply fixing all the Phan notices.

These are the main remaining ones for FacturXProtocol:
```
einvoicing\class\protocols\FacturXProtocol.class.php:293 PhanParamTooMany Call with 4 arg(s) to \CIIProtocol::buildXML(array $invoiceData, array $linesData, $profile = '') which only takes 3 arg(s) defined at einvoicing\class\protocols\CIIProtocol.class.php:1383
einvoicing\class\protocols\FacturXProtocol.class.php:536 PhanTypeMismatchPropertyProbablyReal Assigning $allErrors of type ''|string to property but \FacturXProtocol->errors is array (no real type) (the inferred real assigned type has nothing in common with the declared phpdoc property type)
einvoicing\class\protocols\FacturXProtocol.class.php:952 PhanTypeMismatchArgumentProbablyReal Argument 1 ($thirdparty) is null of type null but \EInvoicing::getBuyerCommunicationURI() takes \Societe (no real type) defined at einvoicing\class\einvoicing.class.php:2830 (the inferred real argument type has nothing in common with the parameter's phpdoc type)
```

mdeweerd:
@eldy Conflict resolved.

mdeweerd:
@eldy Fixed again

PR #351, opened 2026-07-05
fix(esalink): migrate token auth to OAuth2 client_credentials (API v1.2.6)

Charlymd:
## Summary

Two related fixes in the Esalink/Hubtimize integration:

### 1. OAuth2 token endpoint (breaking change in API v1.2.6)

Hubtimize disabled `POST /v1/token` (JSON body). The replacement is `POST /v1/oauth2/token` using RFC 6749 `client_credentials` flow with `application/x-www-form-urlencoded` encoding.

`EsalinkPDPProvider.class.php`
- `getAccessToken()`: switch to `http_build_query(["grant_type" => "client_credentials", "client_id", "client_secret"])` + `Content-Type: application/x-www-form-urlencoded`, call `oauth2/token`
- `callApi()`: replace `$resource != 'token'` anti-recursion guard with `$callType != 'get_access_token'`

Ref: https://help.hubtimize.fr/fr/article/obtenir-un-jeton-api-token-txo1bh/

### 2. Phan PhanTypeMismatchArgumentNullable fix

`callApi()` declares `$callType` as `string|null` but `AbstractPDPProvider::logCall()` only accepted `string`, triggering a Phan error. The method already handles `empty($callType)`, so widening to `?string` is safe. Fixes the same issue in `SuperPDPProvider` too.

`AbstractPDPProvider.class.php`
- `logCall()` signature: `string $callType` → `?string $callType`

## Test plan

- [ ] `POST /v1/oauth2/token` returns a JWT with valid credentials
- [ ] `checkHealth()` succeeds after token retrieval
- [ ] Phan analysis passes with no `PhanTypeMismatchArgumentNullable` on `logCall()`
- [ ] Sending an invoice works end-to-end

### DIFF
--- a/einvoicing/class/providers/AbstractPDPProvider.class.php
+++ b/einvoicing/class/providers/AbstractPDPProvider.class.php
@@ -543,7 +543,7 @@ abstract class AbstractPDPProvider
 	 * the business transaction, so it is committed whether the action succeeds or fails,
 	 * without ever forcing a commit on the rest. Same approach as the webhook logging.
 	 *
-	 * @param   string                      $callType   Functional type of the call (empty = do not log)
+	 * @param   ?string                     $callType   Functional type of the call (empty/null = do not log)
 	 * @param   string                      $resource   API resource/endpoint (without leading slash)
 	 * @param   string                      $method     HTTP method (POSTALREADYFORMATED is normalized to POST)
 	 * @param   string|array<mixed>         $params     Request body
@@ -551,7 +551,7 @@ abstract class AbstractPDPProvider
 	 * @param   int                         $statusCode HTTP status code of the response
 	 * @return  ?array{id:int,call_id:string}           Created log identifiers, or null if not logged
 	 */
-	protected function logCall($callType, $resource, $method, $params, $response, $statusCode)
+	protected function logCall(?string $callType, $resource, $method, $params, $response, $statusCode)
 	{
 		global $conf, $user, $dolibarr_main_db_pass, $dbhistory;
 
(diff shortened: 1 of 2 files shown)


pscoffoni-opendsi:
@Daoud-mohamed @eldy — is there any reason not to merge this PR?

The change aligns the EsaLink/Hubtimize connector with Hubtimize's current authentication.
If anything is blocking, could you let us know what's needed to move it forward?

eldy:
> @Daoud-mohamed @eldy — is there any reason not to merge this PR?
> 
> The change aligns the EsaLink/Hubtimize connector with Hubtimize's current authentication.
> If anything is blocking, could you let us know what's needed to move it forward? 
> 

Yes, we were waiting global feedback about pr of charlymd, to know if user is a true trusted user of the pdplibre project... (question submitted into another pr).
So i imagine answer is yes and we can merge pr of charlymd now.

pscoffoni-opendsi:
Thanks for merging, @eldy 🙂 Just to clear up a small mix-up, for the record:
Yes, @Charlymd is a member of the PDP Libre team — no concern there. But that trust question wasn't really the right gate: #351 is about the EsaLink connector, not PDP Libre. 

EsaLink connector for Einvoicing is a community connector — PDP Libre neither develops nor supports it; our only role is the contractual side (the group contract for our members). So on principle we won't arbitrate EsaLink PRs — they should stand on their technical merits, as this one did.

For contrast, #382 is carried by PDP Libre: it's a contractual clarification (pointing EsaLink credentials to the PDP Libre group contract) rather than a technical connector change — so that one legitimately falls on our side.

And on #352 (the actual PDP Libre provider), my earlier reply was about the proposal, still under discussion — never about the author's legitimacy.

Which raises a useful question: should we define a convention in issues/PRs to make the motivation and origin explicit — technical vs contractual? It would make it much easier to know who should weigh in on what.

PR #360, opened 2026-07-06
Qual(einvoicing): Add type casting and suppress Phan warnings

mdeweerd:
### Qual(einvoicing): Add type casting and suppress Phan warnings

- Added type casting to ensure proper data types in calculations.
- Added suppressions for Phan warnings to avoid false positives.
- Updated the condition to check for the existence of `$outputlangs` instead of checking if it's empty.


mdeweerd:
@eldy Fixed the conflicts.

PR #422, opened 2026-07-22
FIX : refuse form confirm reasons are not translated

jyhere:


einvoicing/class/einvoicing.class.php
Daoud-mohamed, line 279: Label is replaced by a code here.

Daoud-mohamed, line 283: The label is **not** replaced by a code here. i think it would be better to unify the REASONS array by moving label and desc to the `.lang` files and keeping only the code. then adapt the `getReasonsByStatus` function to return the translated label and description.

einvoicing/class/einvoicing.class.php
```diff
@@ -731,12 +731,14 @@ public function getReasonsByStatus($status, $withDetails = 1)
 			return null;
 		}
 
+        global $langs;
```
daGrumpf-bxp, line 734: CI PHPCS ko : indentation en espaces.

```suggestion
		global $langs;
```

Deux points au passage :
- `SupplierInvoiceHelper.class.php:519` et `einvoicing.class.php:1624` font déjà `$langs->trans()` sur le retour → double traduction à retirer.
- même bug non corrigé ligne 1415, `$reasonLabel` est affiché brut sur la card.

Daoud-mohamed:
Same fix already merged [PR](https://github.com/Dolibarr/dolibarr-community-modules/pull/492)

PR #423, opened 2026-07-22
Option does not appear in setup_options

Delthair:
As such the option does not appear into the setup_options and is therefore impossible to activate it

### DIFF
--- a/einvoicing/admin/setup_options.php
+++ b/einvoicing/admin/setup_options.php
@@ -350,7 +350,7 @@ if (!getDolGlobalString('EINVOICING_DISABLE_SYNC_AP_TO_DOLI')) {
 	$item->fieldAttr['placeholder'] = $langs->transnoentities('Hours');
 	$item->cssClass = 'maxwidth100';
 
-	if (getDolGlobalString('EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE')) {
+	if (!getDolGlobalString('EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE')) {
 		// Setup conf to enable or not the consistency check on supplier invoice validation
 		$item = $formSetup->newItem('EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION');
 		$item->helpText = $langs->transnoentities('EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_HELP');


Delthair:
Please see issue #437

einvoicing/admin/setup_options.php
```diff
@@ -350,7 +350,7 @@
 	$item->fieldAttr['placeholder'] = $langs->transnoentities('Hours');
 	$item->cssClass = 'maxwidth100';
 
-	if (getDolGlobalString('EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE')) {
+	if (!getDolGlobalString('EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE')) {
```
Daoud-mohamed, line 353: For me, it would be better to remove the configuration and keep this option hidden until the feature is ready, rather than conditionally displaying it based on another hidden option.

Delthair: How can we keep this hidden ? if we are going to receive an invoice from a supplier, and we do not check its consistency that means that i receive an invoice for 100 and modify if to 10 and pay 10 ? I do not understand how can this happen with einvoicing.

Daoud-mohamed: Until now, the option `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` has been hidden and conditioned by the hidden option `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE`

If EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION is a ready feature, we can completely remove the EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE option.

If EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION is not ready, we can simply remove all this block and keep EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION as a hidden option.

Delthair: Ok i understand and i misread the name of the constant and therefore the fix is irrelevant. Nonetheless, isn't that mandatory to not be able to modify the total amounts of an einvoice ? just as for the outgoing einvoice ? once sent you cant modify it ? I would expect the same behavior when you receive an invoice no ? We've had discussions on the possiblity to modify the lines of a supplier invoice in order to reaffect the lines to different product or accounting codes but the base line is to check the consistency so as not to modify the total amount and ratio within the supplier bills

atm-maxime:
@Delthair You're both right. As discussed today, we need this feature for checking the amounts while validating the supplier invoice. The fact that the option was a hidden one is just that the feature is not stable enough. But as soon as it's confirmed, the option will become visible (on defaulted to ON).

einvoicing/admin/setup_options.php
daGrumpf-bxp: @Delthair By design an incoming e-invoice cannot be modified: what Dolibarr stores is exactly what the vendor sent you — lines, VAT and totals are taken as-is from the received document, same rule as on the outgoing side (once it's sent, it's frozen). So the amounts of a received invoice are not editable in the first place; the consistency check is only an extra safety net on top of that, which is why keeping it hidden for now doesn't open any door to altering the totals.

Reassigning a line to another product (or another accounting code) is a separate concern and never touches the amounts. The vendor product <-> Dolibarr product mapping is done exactly as we discussed, and it is already developed in PR #455: a mapping screen that lists the lines of a flow, shows which ones the matching resolved, and lets you link the unresolved ones to existing products — the link being stored as a vendor reference (`llx_product_fournisseur_price`), so the next synchronization finds them by itself. PR #456 completes it by keeping the vendor wording on the lines routed to the default product.

PR #425, opened 2026-07-22
Fix/dolibarr 17 compatibility

atm-maxime:
If it's OK for you, this changes are made to declare the module available since v17 of Dolibarr.
It fixes the **getDolGlobalFloat** used even if it's present since v21
It adds the **dolGetButtonAction** working with a dropdown (present since v18)


atm-maxime:
phpcs fixed !

PR #429, opened 2026-07-22
FIX : add anti-CSRF state check on via-partner OAuth token return

atm-maxime:
The "via partner" (grey-label) authorization link always used the literal state 'none', so admin/setup.php had nothing to verify when receiving accesstoken/refresh_token/expires_in back from the proxy. Any request forging these GET params would be saved as the production PDP credential.

SuperPDPProvider::initFormSetup() now generates a per-page-load random nonce ('none-' prefixed so the proxy still treats it as "no scope requested"), stored in $_SESSION['einvoicing_superpdp_viapartner_oauth_state'] and used consistently in both authorize links it builds. admin/setup.php now checks this state before calling saveOAuthTokenDB(), mirroring the existing check already used for the direct authorization_code flow.

### DIFF
--- a/einvoicing/admin/setup.php
+++ b/einvoicing/admin/setup.php
@@ -325,9 +325,18 @@ if (GETPOST('error')) {
 
 
 if (GETPOST('accesstoken') && $provider instanceof AbstractPDPProvider) {
-	// We are in the return of an OAUT proxy authorize+token callback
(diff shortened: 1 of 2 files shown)


atm-maxime:
@eldy I don't know how the viapartner works with SuperPDP, can you confirm this fix ?

PR #433, opened 2026-07-23
[Einvoicing] Extract db insertion of supplier invoice lines

spsolauv:
Currently, the creation of supplier invoice lines was done directly by the call to `$supplierInvoice->create($user)`.
By moving this creation to an independent method, it makes possible to create lines on an existing supplier invoice, which is a prerequisite for coming feature allowing to manually reimport lines.


atm-maxime:
Refactoring of the supplier invoice creation by adding the lines 1 by 1 to be also used by the reimport function

Issue #438, opened 2026-07-23
Impossible to validate a supplier invoice when consistency check is on and difference of amounts is = 0


Delthair:
### Bug

Impossible to validate a supplier invoice when consistency check is on and difference of amount is = 0 

### Module

einvoicing

### Steps to reproduce the behavior and expected behavior

On a supplier invoice, when the option of consistency is ON modify lines when same amout you validate and it creates an error

PR #439, opened 2026-07-23
Update SupplierInvoiceHelper.class.php fix ISSUE #438

Delthair:
#438 : Allow that when difference == 0 to go through the checks withat a 500 error


atm-maxime:
Fixed by #451

PR #367, opened 2026-07-07
[Einvoicing] NEW supplier invoice : allow manual import of lines

spsolauv:
After comments and issues on PR #269, code has been modified : 
- split PR features (remove some modifications which will be made in dedicated PRs to come)
- fix issues detected by @eldy when getting XML data see SupplierInvoiceHelper::getXmlData in https://github.com/Dolibarr/dolibarr-community-modules/commit/db421d402482da31d55ca74e857759a8be217f69
- commit only en_US lang as suggested in https://github.com/Dolibarr/dolibarr-community-modules/pull/269#issuecomment-4896977988

Issue #419, opened 2026-07-21
[Einvoicing] Manage VAT regime of main company in generated einvoice


spsolauv:
### Feature Request

Need to have an additional information when generating XML e-invoice : VAT regime from seller/main company (corresponding to dueDateTypeCode in CII)

### Use case

This information is important for accountancy, because it can impact cash flow management and is required for legal compliance.

### Suggested implementation

- Add a new field 'VAT regime' for main company
  - this parameter could be put in VAT module setup, but this module is not always enabled
  - so it could be more appropriate to add it to main company setup (admin/company.php) + save in const table => PR to make in Dolibarr core
- Waiting for PR and for Dolibarr previous versions, add a new parameter 'VAT regime' in setup fo einvoicing module
- Available values could be same as CII 16B standard (see p171 => https://communaute.chorus-pro.gouv.fr/wp-content/uploads/2018/11/External_Specifications_EDI_Appendix_V4.10.pdf)
  - “5” (Flow with payability of the tax at the issue date of the
invoice);
  - “29” (Flow with payability of the tax at the delivery);
  - “72” (Cashing)
- Modification of `einvoicing/lib/buildinvoicelines.inc.php` to get this new information in `$taxBreakdown` array
- Modification of function buildTaxNode in `einvoicing/class/protocols/CIIProtocol.class.php` to add the new information in `ram:DueDateTypeCode`

Any comment or suggestion is welcome.

@aspangaro what is your point of view ?

daGrumpf-bxp:
Part of this is now implemented in #460: the CII `ram:DueDateTypeCode` (BT-8) is emitted on each `ApplicableTradeTax`, and the same value is passed to the Factur-X builder.

The rule it applies, in the absence of a "VAT regime" field on the main company:
- `72` (cashing) as soon as the invoice carries a service line — the default regime for services;
- `5` (invoice date) for a seller who opted for the "TVA d'après les débits" scheme, driven by the `EINVOICING_VAT_ON_DEBITS` option, which also adds the corresponding legal mention on the invoice as a TXD note;
- nothing at all on a goods-only invoice, whose VAT falls due on the delivery it already dates.

What that PR does **not** do, and what this issue can keep tracking: an explicit "VAT regime" setting on the main company (or in the module setup while waiting for a core PR), and the `29` (delivery) value. If you'd rather have the setting drive it than the heuristic, that is a small change on top.

PR #440, opened 2026-07-23
[Einvoicing] NEW manual import of supplier invoice lines

spsolauv:
See previous aborted PR #269 & #367 to get feature description.

:warning: Depends from #433 & #436


einvoicing/class/actions_einvoicing.class.php
```diff
@@ -436,6 +437,12 @@ public function addMoreActionsButtons($parameters, $object, &$action, $hookmanag
 						);
 					}
 
+					if ($object->status == FactureFournisseur::STATUS_DRAFT &&
+						$user->hasRight('facture', 'write') &&
```
atm-maxime, line 445: Use the perm on supplier invoice

einvoicing/class/actions_einvoicing.class.php
```diff
@@ -450,6 +457,103 @@ public function addMoreActionsButtons($parameters, $object, &$action, $hookmanag
...
+				0,	// value_as_key
+				'',	// moreparam
+				0,	// translate
+				0,	// maxlen
+				0,	// disabled
+				'', // sort
+				'minwidth300'	// morecss (CSS class for styling)
+			);
+
+			// Modal allowing to reimport supplier invoice lines
```
atm-maxime, line 521: What was the limitation of the standard formconfim so you had to use a specific modal ?

einvoicing/core/modules/modEInvoicing.class.php
```diff
@@ -601,6 +605,10 @@ public function init($options = '')
 		// 	1
 		// );
 
+		// Societe extrafields
+		// Use a <select> to have the default null value (falling back to default supplier invoice import type parameter set in the module settings)
```
atm-maxime: Why do we need to store the import type ?
The supplier invoice is created the first time by the sync. Then if I want to reimport it with your new feature, I can. What's the goal of having this import type ?

einvoicing/class/actions_einvoicing.class.php
spsolauv: I didn't know it was possible to use this function with specific inputs (as radio buttons), I will try to use it.

einvoicing/class/actions_einvoicing.class.php
spsolauv: Ok

einvoicing/core/modules/modEInvoicing.class.php
spsolauv: This extrafield is added to societe, not to the supplier invoice. It allows to choose a specific import type by supplier, then, when we check if user can reimport supplier invoice lines, we first check the import type of the supplier : 
- if the value of `einvoicing_supplier_invoice_lines_import_type` is not set for the current supplier, then we will use the value of module param `EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE`
- if the value of `einvoicing_supplier_invoice_lines_import_type` is set to manual or auto for the current supplier, then we use this value

We did this to be able to use manual import at supplier level, and use module param as fallback.

einvoicing/class/actions_einvoicing.class.php
spsolauv: Use formconfirm but use type 'other' for input types argument for Dolibarr compatibility reasons

einvoicing/admin/setup_options.php
```diff
@@ -356,6 +356,15 @@
 	$item->fieldAttr['placeholder'] = $langs->transnoentities('Hours');
 	$item->cssClass = 'maxwidth100';
 
+	$item = $formSetup->newItem('EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE');
```
eldy, line 359: Why an option fo this ?
What is goal to have option of if feature exists ?

spsolauv: The option at module level is a fallback value if you did not set the import type on the supplier invoice 'societe'. By example, you can choose to set import type to : 
- manual on 'societe 1' => force to manual import whatever module option value is
- auto on 'societe 2' => force to auto import whatever module option value is
- not set (null) on 'societe 3' => use module option value as fallback

Functionally, you may choose to : 
- enable manual import globally (module option)
- and to set auto just for a few 'societe'
or vice versa

einvoicing/langs/en_US/einvoicing.lang
```diff
@@ -349,6 +349,25 @@ EInvoiceSupplierOrderRefWrongSupplier=A purchase order with reference %s exists
 WarningEinvoicingInvoiceStatusDifferentProvider=Invoice generated or sent for another Access Point %s.
 EinvoicingCantUnvalidateATransmittedInvoice=You try to modify the status of an invoice that is locked once the invoice has been transmitted to the Access Point
 EinvoicingCantModifyATransmittedInvoice=You try to modify a property that is locked once the invoice has been transmitted to the Access Point
+SupplierInvoiceLinesImportType=Supplier invoice lines import type
+EinvoiceImportLines=Reimport e-invoice lines
+SupplierInvoiceExtractLinesWarning=Existing supplier invoice lines will be deleted
+ExtractAllLines=Extract all lines automatically
```
eldy: Can you rename this label into:  
Reset lines and re-import them

Rename: Extract all lines to free lines
Reset lines and re-import them into line without predefined product

einvoicing/core/modules/modEInvoicing.class.php
eldy: Before any PR to add, when you need a new data in database,you must first push a PR that add the structure.
Also you should split your PR in 2:

One to add the feature to select the "reimport line" with the popup of user and the next one will work on saving this information.

einvoicing/langs/en_US/einvoicing.lang
spsolauv: Translations have been updated

daGrumpf-bxp:
Nice to see this one back — the free-line mode in particular fills a real gap (vendors whose products we never want in the catalogue).

:warning: **One thing worth a careful look before the target-product mode is set in stone**, because it is exactly the case we fought in #456 (`ca85051`):

In the `$targetFkProduct > 0` branch, `_findOrCreateProductFromEinvoiceLine()` is never called, so `$res` is never assigned and `$productMatchType = (string) ($res['matchtype'] ?? '')` always ends up empty. The guard right below — `if ($productId > 0 && $productMatchType != 'defaultrouting')` — therefore never fires, and `$supplierPriceEntries` gets filled with the catch-all target product carrying the vendor reference of the line.

On this PR's own path it is harmless today: the `reimportLines` handler drops `$supplierPriceEntries` without persisting it. But `createSupplierInvoiceLinesFromSource()` is also the sync path, where those entries *are* written to `llx_product_fournisseur_price` — and the vendor reference is the **first** matching criteria of the import. Once a vendor reference points at the catch-all product, every later line of that vendor silently routes there, and nothing in the UI says why. That is precisely why the `defaultrouting` exclusion was added. Forcing `$productMatchType = 'defaultrouting'` (or simply skipping the collection) in the target-product branch keeps it safe whatever the caller does.

Two smaller notes:

- The product list of the modal (`SELECT p.rowid ... WHERE cp.fk_categorie = ...`) filters neither on `p.tobuy = 1` nor on `entity`, so it can offer non-purchasable products — and products of another entity — as the target of a supplier invoice line. Same fix as the first commit of #489.
- Design question more than a review remark: #455 and #489 answer the "product not found" case by writing the vendor reference, so the *next* invoice of that vendor resolves on its own. The target-product mode goes the other way — after the collapse per VAT rate the vendor references are gone from the invoice, so there is nothing left to map. The two can coexist, but it would help to state which one is the recommended path, otherwise users end up with four ways of answering "which product for this line" (default routing product, mapping screen, import type, target category).

Happy to help on any of these.

daGrumpf-bxp:
Follow-up on my previous comment: rather than leave it as a reading of the code, I ran it. Here is the evidence and a patch.

**Setup** — Dolibarr 20.0.5 test instance, the `CIIProtocol` of this branch loaded as is, `createSupplierInvoiceLinesFromSource()` called directly with two parsed lines carrying a vendor reference (`MALT-PIL-25`, `HOU-CAS-1`), each run wrapped in a rolled-back transaction (database verified untouched afterwards).

| Scenario | this branch | with the patch below |
|---|---|---|
| Target product mode (`target_fk_product = 3`) | `supplierPriceEntries` = **2 entries, both `productId: 3`, carrying `refFourn: MALT-PIL-25` and `refFourn: HOU-CAS-1`** — and `desc = NULL` on both lines | `[]`, and `desc` = the wording of the XML |
| Auto, product matched by buyer reference | 2 entries, products 3 and 2 | identical |
| Auto, no match, vendor has a default routing product | `[]`, `desc` = wording of the XML | identical |
| Free lines mode | `[]`, `fk_product = NULL` | identical |

So the guard indeed never fires in the target-product mode, and the vendor references of the invoice would be collected onto the catch-all product. Harmless on this PR's own path (the `reimportLines` handler drops the array), but the same function is the sync path, where those entries are written to `llx_product_fournisseur_price`.

The three pre-existing modes come out byte-identical with and without the patch, so the change is confined to the new branch:

```diff
 			if (!$is_deposit_line && !$freeLines) {
 				if ($targetFkProduct > 0) {
+					// A product imposed for the whole invoice is a catch-all, same nature as the default
+					// routing product of the vendor: keep the wording of the XML on the line, and never
+					// glue the vendor reference of the line onto it.
 					$productId = $targetFkProduct;
+					$productMatchType = 'defaultrouting';
 				} else {
 					// Sync or create product
 					$res = $this->_findOrCreateProductFromEinvoiceLine($parsedLine, $flowId);
 					...
 					$productId = $res['res'];
-				}
-				$productMatchType = (string) ($res['matchtype'] ?? '');
+					$productMatchType = (string) ($res['matchtype'] ?? '');
 
-				// Collect supplier price data to be created after invoice is saved.
-				// Not for a default routing product: it is a catch-all, so gluing the vendor reference of
-				// the line onto it would make every next invoice match it instead of the real product.
-				if ($productId > 0 && $productMatchType != 'defaultrouting') {
-					$supplierPriceEntries[] = [
-						'productId' => $productId,
-						'unitPrice' => (float) $parsedLine['netpriceamount'],
-						'refFourn'  => (!empty($parsedLine['prodsellerid']) && $parsedLine['prodsellerid'] !== '0000') ? (string) $parsedLine['prodsellerid'] : '',
-						'tvaTx'     => (float) ($parsedLine['rateApplicablePercent'] ?? 0),
-					];
+					// Collect supplier price data to be created after invoice is saved. Only here: a
+					// vendor reference is recorded when the product was resolved from the line itself.
+					// Not for a default routing product: it is a catch-all, so gluing the vendor reference
+					// of the line onto it would make every next invoice match it instead of the real product.
+					if ($productId > 0 && $productMatchType != 'defaultrouting') {
+						$supplierPriceEntries[] = [
+							'productId' => $productId,
+							'unitPrice' => (float) $parsedLine['netpriceamount'],
+							'refFourn'  => (!empty($parsedLine['prodsellerid']) && $parsedLine['prodsellerid'] !== '0000') ? (string) $parsedLine['prodsellerid'] : '',
+							'tvaTx'     => (float) ($parsedLine['rateApplicablePercent'] ?? 0),
+						];
+					}
 				}
 			}
```

Two things happen there. Moving the collection inside the matcher branch is the part that matters: a vendor reference is then recorded only when the product was actually resolved from the line, so any future "product imposed" mode is safe by construction, without its author having to know this guard exists. Setting `$productMatchType = 'defaultrouting'` is what restores the second half of the behaviour: without it the target-product mode leaves `desc` empty and every line of the invoice displays the label of the catch-all product instead of what the vendor wrote.

Unrelated small note while I was in there: `EINVOICING_IMPORT_AS_FREE_LINES` already exists on main (`CommonProtocol` ~line 1049, a yes/no in the module setup) and imports unmatched lines as free description lines. It may cover part of what the free-line mode of this PR does, and save you some code.

Happy to open this as a PR on your branch if that is easier.

spsolauv:
Import type of a societe why was previously stored in a Dolibarr extrafield is now stored in the table `llx_einvoicing_extrafields` (because Dolibarr extrafields are prohibited from einvoicing module)

einvoicing/core/modules/modEInvoicing.class.php
spsolauv: No need to modify database anymore, import type of a societe is now stored in table llx_einvoicing_extrafields.

daGrumpf-bxp:
Review of the current head (`e5c64ad5`).

> **Updated after running the branch.** The first version of this comment was code reading only. I have since deployed the branch on a bench where Dolibarr 18, 19, 20, 21, 22, 23 and 24 share the same module checkout, and executed the feature over HTTP. Points marked **measured** were run; the rest is code reading. One claim from the first version — data loss on a failed re-import — turned out to be **wrong and is withdrawn** in §6 below.

## 1. The merge of `main` reverted 6 settings in `setup_options.php`

This is unrelated to the feature but it is the most urgent point. The merge commit `e5c64ad5` resolved the conflict on `einvoicing/admin/setup_options.php` by keeping the old side, so `git diff main..HEAD` shows these entries as **removed**:

* `EINVOICING_VAT_POINT_DATE_CODE` (BT-8, issue #419) — gone
* `EINVOICING_REQUIRE_ROUTABLE_RECIPIENT` — back to a yes/no, the `STRICT` third value is gone
* `EINVOICING_SEND_APPROVED_ON_VALIDATION` (status 205) — gone
* `EINVOICING_SEND_PAYMENT_SENT_STATUS` (status 211) — gone
* `is_greyed` handling on protocols — gone, greyed-out protocols become selectable again
* `EINVOICING_TELL_CUSTOMER_PAYMENT_RECEIVED` — gone
* the `header("Location: ...")` + `exit` after `savesyncoptions` — gone, so a refresh re-submits the form

The code reading those constants is still there, they just become unreachable from the admin page. Could you redo the merge so that `setup_options.php` only shows the two new items in this diff?

## 2. Fatal error on the supplier invoice card, Dolibarr 18 to 21 — measured

Same module checkout, a draft supplier invoice loaded on each instance:

| | 18 | 19 | 20 | 21 | 22 | 23 | 24 |
| page complete | no | no | no | no | yes | yes | yes |

On 18-21 the HTML stops in the middle of `llxFooter()`, at `<div id="dialogforpopup">` — no `</body></html>`, no module JS, and every footer script after it is lost:

```
PHP Warning:  Attempt to read property "socid" on null in .../actions_einvoicing.class.php on line 535
PHP Warning:  Attempt to read property "status" on null in .../actions_einvoicing.class.php on line 536
PHP Fatal error:  Uncaught Error: Class "Categorie" not found in .../actions_einvoicing.class.php:545
#0 core/class/hookmanager.class.php(319): ActionsEInvoicing->beforeBodyClose()
#1 main.inc.php(4123): HookManager->executeHooks()
#2 fourn/facture/card.php(4256): llxFooter()
```

Two independent causes, both verified in the core sources:

1. `main.inc.php` calls `executeHooks('beforeBodyClose', $parameters)` **without `$object` and `$action`** on 18, 19, 20 and 21; the two extra arguments only appear from 22 on. So `$object` is `null` in `beforeBodyClose()`: `$object->socid` yields `null`, `isSupplierImportInvoiceLinesAuto(null)` returns `false` because `if ($socId > 0)` never matches, and `$object->status == FactureFournisseur::STATUS_DRAFT` is `null == 0`. All three guards pass, on any supplier invoice, whatever the setting.
2. `new Categorie($db)` (line 545), but `categories/class/categorie.class.php` is only required by `fourn/facture/card.php` **from Dolibarr 22 on**.

Note the module already declares the `all` hook context, so `invoicesuppliercard` is reached without any module re-activation — the new entry in the descriptor is not what enables this.

**Also measured, on 24:** on a supplier invoice that has nothing to do with e-invoicing, the modal HTML and `supplier_invoice.js` are still emitted (the button is correctly absent, being inside the `einvoicing_extlinks` lookup). The modal needs the same guard as the button, and `SupplierInvoiceHelper::isEInvoice()` already exists for that.

## 3. Clicking the button returns HTTP 500 — measured

`SupplierInvoiceHelper::getXmlData()` **throws** when there is no `llx_einvoicing_document` row (`No result found when searching for supplier invoice with id ...`), and again on duplicates. Line 853 calls it with no try/catch, so the `$xmlData === null || $xmlData === ''` guard right below is never reached.

Measured twice on Dolibarr 24:

* a plain supplier invoice draft, `card.php?facid=235&action=reimportLines&token=...` → **HTTP 500**, empty response. The action itself has no e-invoice guard, and the URL is reachable by hand.
* a supplier invoice imported through `createSupplierInvoiceFromSource()`: it has an `einvoicing_extlinks` row, so **the button is displayed**, but no `einvoicing_document` row, so a plain click → **HTTP 500**.

```
PHP Fatal error: Uncaught Exception: No result found ... in SupplierInvoiceHelper.class.php:367
Stack trace:
#0 einvoicing/class/actions_einvoicing.class.php(853): SupplierInvoiceHelper::getXmlData()
#1 core/class/hookmanager.class.php(333): ActionsEInvoicing->doActions()
#2 fourn/facture/card.php(166): HookManager->executeHooks()
```

Note also that `getXmlData($id, true)` calls the Access Point (`fetchFlowXml`) when `xml_data` is empty — a network round trip triggered by a UI button, with no indication to the user.

## 4. The feature switches itself on for every existing install — measured

`EINVOICING_SUPPLIER_INVOICE_LINES_IMPORT_TYPE => AUTO` is declared in `modEInvoicing::$const`, which is only written when the module is **activated**. On an install that simply pulls the new code the constant does not exist, `getDolGlobalInt()` returns 0, which is neither `AUTO(1)` nor `MANUAL(2)` nor `USE_GLOBAL_CONFIG(-1)`, and `isSupplierImportInvoiceLinesAuto()` returns `false`, i.e. manual.

Measured: the constant is absent on all 7 bench instances, and the button and the modal appear with no configuration at all. This is what turns §2 and §3 into an upgrade-day outage rather than an edge case.

## 5. Mode 3 destroys the amounts when a line carries an allowance — measured

Test invoice: 7 CII lines, all at 20%, with **one line-level allowance** (`BasisAmount` 307.20, `ActualAmount` 30.72, so BT-131 = 276.48; document totals 1183.68 / 236.74 / 1420.42).

The **normal import** handles it perfectly: 7 lines, line 1 at `remise_percent = 10` and `total_ht = 276.48`, invoice totals matching the XML to the cent.

Then "reset lines and re-import" in mode 3 onto a target product:

| | before | after mode 3 |
| --- | --- | --- |
| lines | 7 | 1 |
| total HT | 1 183.68 | **27.65** |
| total VAT | 236.74 | **5.53** |
| total TTC | 1 420.42 | **33.18** |

The resulting line is `pu_ht = 30.72, qty = 1, remise_percent = 10`. Cause: the aggregation (lines 905-916) sums `netpriceamount * billedquantity` instead of `lineTotalAmount` — the right variant is there, commented out on line 913 — and the aggregated line inherits `lineAllowances` from the first line of the VAT group while `lineTotalAmount` has been forced to 0. `resolveLineDiscountPercent()` then computes `priceWithoutDiscount = 0 + 30.72`, so `subprice` becomes the discount amount and a 10% discount is applied on top. The invoice silently loses 97.7% of its value, in a draft the user is about to validate.

`EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` would catch it, but that constant is absent on all 7 bench instances — it sits behind `..._AVAILABLE` and is off by default, so in practice nothing catches this.

The same test with no allowance is correct (7 lines → 1 line at 1214.40, totals preserved), so the aggregation only breaks in the presence of line-level allowances or charges — exactly what BT-131 (`lineTotalAmount`) accounts for and `netpriceamount * billedquantity` does not.

Related, same aggregation: the aggregated line also inherits `additionalRefDocs` from its template line, so a deposit line entering the aggregate carries its linked-document handling with it.

## 6. Transaction handling — my earlier "data loss" claim, withdrawn

I first wrote that the deletion loop runs without a transaction and that a failure mid-way leaves the invoice with no lines. **That is wrong**: `doActions()` already opens a transaction at **line 661**, at the top of the method.

Measured: mode 3 submitted with no target product deletes the 7 lines, hits the `EinvoiceReimportLinesMissingTargetFkProduct` error path, and the `rollback()` **restores them** — same line rowids before and after. No data is lost. Sorry for the wrong call.

What remains worth fixing is smaller but real: the new block's `$db->commit()` (line 942) and `$db->rollback()` (lines 858, 875, 926, 937) close a transaction they did not open, so anything the module wrote earlier in the same request is committed or discarded as a side effect of the re-import outcome. Either open a nested `begin()`/`commit()` pair for this branch, or leave the outer transaction to lines 1044/1047 and just set `$this->error` and return.

## 7. Other findings

* **CIIProtocol.class.php:1355** — the `if ($freeLines)` block is redundant: the `elseif (!$is_deposit_line)` branch above already sets exactly the same description when no product is linked. And since it runs **after** the deposit block, it overwrites the `'(DEPOSIT)'` description of a deposit line in free-line mode. (Code reading; not run.)
* **actions_einvoicing.class.php:549** — the product query has no `entity IN (getEntity('product'))` filter (every other product lookup in the module has one) and no `LIMIT`. The whole category is rendered as an HTML `select` **on every card display**, whether or not the user opens the modal; on a large catalogue this is a page killer. The usual idiom is `PRODUIT_LIMIT_SIZE` or an ajax product selector. The `$db->query()` result is also not checked before `fetch_object()`.
* **Destructive action over GET** — measured: the button href is `card.php?action=reimportLines&token=...` and only JavaScript turns it into a modal, so a plain click (or any context where the JS did not load, e.g. §2) deletes and re-imports the lines with no confirmation, silently falling back to auto mode. It answered 302 and replaced all 7 lines.
* **Permission mismatch** — the button and the modal test `hasRight('facture', 'write')`, an alias of `facture->creer`, the **customer** invoice right (`user.class.php`: `if ($permlevel1 == 'write' && !empty($this->rights->$rightsPath->creer))`), while the action requires `fournisseur->facture->creer`. A user with supplier rights only never sees the button; a user with customer rights only sees it and the click silently does nothing.
* **`EinvoiceReimportLinesError`** is declared with a `%s` but used as `trans(...) . $res['message']`, so the literal `%s` shows up in the message.
* The `return -1` paths at lines 858, 875 and 926 do not set `$this->error`, so `card.php` adds an empty `setEventMessages` after the messages already posted.
* **actions_einvoicing.class.php:1027** — the handler also runs on `$action == 'update'` in the `thirdpartycomm` context. When the field is not in the POST, `GETPOSTINT` returns 0, which is not in the whitelist, so the per-thirdparty setting is silently reset to "use global config". A `GETPOSTISSET()` guard would fix it.
* **modEInvoicing.class.php** — const type `'int'` where the core consistently uses `'chaine'`; and `dol_include_once('einvoicing/class/einvoicing.class.php')` at the top of the module descriptor loads the whole class on every page that enumerates modules.
* Leftover commented-out `//if (getDolGlobalString('EINVOICING_FLOWS_SYNC_CALL_LIMIT'))` condition in the setup file.
* UX, mode 3: the aggregated line inherits the description of the first line of its VAT group, so a whole invoice ends up labelled "MALT PILSEN BIO 25KG". A neutral label naming the VAT rate would read better.

**What does work**, for the record, measured on 24: mode 1 (auto) and mode 2 (free lines — 7 lines with `fk_product` NULL, totals preserved) both behave as advertised, and mode 3 is correct on an invoice with no line-level allowance.

## 8. Design question: overlap with what already exists

`main` already offers a full ladder of choices, but only **at synchronisation time**:

| Need | Already in `main` | What this PR adds |
| --- | --- | --- |
| Resolve the product | `findProductFromEinvoiceLine()`, 5 strategies | — |
| Catch-all product per vendor | default routing product on the thirdparty card (`fetchDefaultRouting`, `matchtype='defaultrouting'`) | the same, but picked per invoice |
| Lines with no product | `EINVOICING_IMPORT_AS_FREE_LINES` (CommonProtocol.class.php:1157) | the same, but forced on every line |
| Create missing products | `EINVOICING_PRODUCTS_AUTO_GENERATION` | — |
| Manual mapping before import | `einvoicing/product_mapping.php` | — |

So modes 1 and 2 of the modal largely duplicate existing settings with slightly different semantics (the global option only applies to *unresolved* lines, mode 2 applies to all of them), and mode 3 duplicates the default routing product while adding a second piece of configuration (a product category).

What is genuinely new, and worth having, is:

1. **Replaying the import on an existing draft.** Today the decision is made once at sync time; if it was wrong the only way out is to delete and resynchronise. That is a real gap.
2. **The "one line per VAT rate" aggregation**, which nothing else does.

Would it make sense to narrow the PR to those two, and have the modal pre-select its mode from the existing settings (`EINVOICING_IMPORT_AS_FREE_LINES`, default routing product of the vendor) instead of introducing a parallel configuration path? It would also address the concern raised on #564 about not adding an option for something the module already knows.

spsolauv:
@daGrumpf-bxp Thanks, I reverted useless changes from setup_options.php (certainly a fail when trying to resolve conflicts with latest updates from main branch)

PR #443, opened 2026-07-23
[Einvoicing] Fix supplier invoice consistency verif with discount

spsolauv:
**Supplier invoice check consistency on validation :**
Total VAT excl. was not compared properly (was using wrong amount from the einvoice data).

### DIFF
--- a/einvoicing/class/helpers/SupplierInvoiceHelper.class.php
+++ b/einvoicing/class/helpers/SupplierInvoiceHelper.class.php
@@ -125,9 +125,8 @@ class SupplierInvoiceHelper
 			$details = self::getInvoiceDetailsForComparison($dolSupplierInvoice, $vatComputeMode);
 
 			// VAT excl. total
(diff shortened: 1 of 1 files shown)


Delthair:
Should'nt we use this instead : 
if (!self::areAmountsEqual($details['total_ht'], $parsedHeader['taxBasisTotalAmount'] ?? 0)) {   

in the protocols it's that field that is used : 
CIIProtocol.class.php, ligne 841 :   $supplierInvoice->total_ht = $parsedHeader['taxBasisTotalAmount'] ?? 0;   

 FacturXProtocol.class.php ligne 1470 :  $supplierInvoice->total_ht = $parsedHeader['taxBasisTotalAmount'] ?? 0;

spsolauv:
I was not sure that `taxBasisTotalAmount` was always corresponding to the amount on which taxes are applied if they were duty free lines in the invoice. But it seems `taxBasisTotalAmount` takes in account duty free lines in its total, so I modified as you suggest. Thanks @Delthair

PR #445, opened 2026-07-23
FIX #44 Update SupplierInvoiceHelper.class.php

### LINKED ISSUE #444: CONSISTENCY CHECK does not support Supplier Credit Note
### Bug

When consistency checks for a supplier credit note, the amounts are checked but not in absolute value therefore the checks fails all the time.  
Dolibarr stores credit note (avoir) amounts as negative (see FactureFournisseur::addline()),  but the e-invoice XML keeps them positive : only the document type code (BT-3) marks it as  a credit note per EN16931 / Factur-X. Compare in absolute value in that case.

### Module

einvoicing

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

When consistency checks for a supplier credit note, the amounts are checked but not in absolute value therefore the checks fails all the time.  
Dolibarr stores credit note (avoir) amounts as negative (see FactureFournisseur::addline()),  but the e-invoice XML keeps them positive : only the document type code (BT-3) marks it as  a credit note per EN16931 / Factur-X. Compare in absolute value in that case.

### Attached files

_No response_

Delthair:
fix #444 possibility to check consistency for credit note by using abs amount from dolibarr.

### DIFF
--- a/einvoicing/class/helpers/SupplierInvoiceHelper.class.php
+++ b/einvoicing/class/helpers/SupplierInvoiceHelper.class.php
@@ -121,8 +121,20 @@ class SupplierInvoiceHelper
 
 		$amountErrors = [];
 
+		$isCreditNote = ($dolSupplierInvoice->type == FactureFournisseur::TYPE_CREDIT_NOTE);
+
 		foreach ($calculationRules as $calculationRule => $vatComputeMode) {
 			$details = self::getInvoiceDetailsForComparison($dolSupplierInvoice, $vatComputeMode);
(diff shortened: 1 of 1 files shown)


atm-maxime:
@spsolauv can you confirm this fix since you have worked on the consistency check ?

spsolauv:
In case of a credit note, we can't be sure the amounts are positive or negative (depends from input format and implementation), so taking absolute value for comparison seems to be ok for these values.

PR #455, opened 2026-07-26
[einvoicing] Fix the "Create the product" button and add a vendor product mapping screen

daGrumpf-bxp:
Two changes on the same UX path: what the module offers when a supplier invoice import cannot find a product and the auto-creation of products is disabled.

## 1. Fix: "Create the product" opened a service creation form

The recommended action always displayed **"Create the product"**, but the prefilled link used the type detected by `_detectProductTypeFromEinvoiceLine()`, so it often opened the **service** creation form (`type=1`) — label and destination disagreed.

- The label now follows the detected type (`CreateTheService` / `CreateTheProduct`).
- Fallback on product when the Service module is not enabled.
- Fixed a misleading comment (`// Fallback = service` on a `return 0`).

## 2. Feature: map the vendor products onto existing Dolibarr products

Creating a new product is the wrong answer when the product already exists in Dolibarr under another reference. The only existing alternative was the per-vendor "default routing product", which routes *every* line of that vendor to a single product.

New page `einvoicing/product_mapping.php`:

- lists every line of a flow (vendor ref, buyer ref, EAN, label, qty, unit price, VAT);
- runs the regular matching in **read-only** mode to show which lines are already resolved and which are not;
- lets the user link the unresolved ones to existing products;
- saves the link as a **vendor reference** (`llx_product_fournisseur_price`) — the first criteria of the matching — so the next synchronization finds the products by itself. No new table, no schema change.

All the lines are mapped in one pass, instead of mapping one line then relaunching the sync only to be stopped by the next one.

Implementation notes:

- the matching steps of `_findOrCreateProductFromEinvoiceLine()` are extracted into a public `findProductFromEinvoiceLine()` (read only). No behavior change on the import itself.
- the `PRODUCT_NOT_FOUND` suggested action now offers **"Map to an existing product"** beside "Create the product" — this is the `$createSupplierRefUrl = 'todo'` left in the commented code.
- entry points: button on the synchronization list, plus a left menu entry (needs a module re-activation to appear).
- lines without a vendor product reference are shown as not mappable (there is nothing to store the link on) with an explanation.

Lang keys added to en_US and fr_FR only.

Checked with phan (`--quick`, CI config) and phpcs (Dolibarr ruleset): clean.


Daoud-mohamed:
Nice feature!

PR #456, opened 2026-07-26
[einvoicing] Keep the vendor wording on invoice lines routed to the default product

daGrumpf-bxp:
## Problem

The **default product** of a vendor (thirdparty card → e-invoicing block) is the catch-all used when the product of an invoice line cannot be found. It works, but the resulting invoice is unreadable: a line linked to a product gets no description from the XML (`CIIProtocol::createSupplierInvoiceLinesFromSource()` only fills `desc` for free lines), so **every** unresolved line of the invoice ends up showing the same label — "Misc purchases", "Divers"… The wording sent by the vendor is lost, which is exactly the information you need on those lines.

Second effect, less visible: a vendor price was registered on the catch-all product with the vendor reference of the line. From the next invoice on, the matching finds the catch-all *by reference* (step 1) — the reference is glued to it, and mapping that line to the real product later means deleting the vendor price by hand first.

## Fix

- the matching now reports **how** the product was resolved (`matchtype`, set to `defaultrouting` when the line fell back on the default product of the vendor);
- such a line keeps the **label and description of the XML** on top of the product link;
- and no vendor price is created for it anymore, so the catch-all never hijacks a vendor reference.

Nothing changes for lines resolved normally (vendor ref, buyer ref, `EI-` ref, label) or auto-created.

Also adds the missing fr_FR translation of `EINVOICING_IMPORT_AS_FREE_LINES` (the option and its help text were displayed in English on a French setup).

Checked with phan (`--quick`, CI config) and phpcs (Dolibarr ruleset): clean.


daGrumpf-bxp:
The branch has been rebased on the current main (head `ca85051a`), which resolved the conflict with the changes merged since — #468, #474 and #479 all touched `CIIProtocol.class.php`.

Current state: `mergeable: CLEAN`, and phan / phpcs are green on the head commit. Nothing left to fix on my side, so the "PR to fix (CI or conflict error)" label can be removed whenever convenient.

Issue #410, opened 2026-07-20
(Einvoicing) Sending "Approved" status fails with Super PDP ("invalid CDAR" error) when receiving a supplier invoice


biedseb1:
### Bug

When receiving a supplier invoice, synchronization works correctly. The supplier invoice is then validated (or left as draft) in Dolibarr.

However, when trying to send the "Approved" status via the module to the PA Super PDP, I get the following error:

Failed to send CDAR file to PDP. Status code: 400. Message: invalid CDAR

Expected behavior:
The "Approved" status should be successfully transmitted to the platform.

Current behavior:
Sending the status fails with a 400 error indicating an invalid CDAR.

Comparison elements:
I reproduced the same scenario with the PA Esalink: in that case, the status is correctly sent and properly received by the supplier, with no error. The issue therefore appears to be specific to the CDAR file generation for Super PDP, rather than the status logic on Dolibarr's side.

Steps to reproduce:

Receive a supplier invoice via Super PDP
Send the "Approved" status
Observe the error "Failed to send CDAR file to PDP. Status code: 400. Message: invalid CDAR"

### Module

--Please select a module--

### Steps to reproduce the behavior and expected behavior

Receive a supplier invoice via Super PDP
Send the "Approved" status
Observe the error "Failed to send CDAR file to PDP. Status code: 400. Message: invalid CDAR"

daGrumpf-bxp:
I reproduced this on Super PDP and it comes down to the electronic address of the CDAR recipient (MDT-73), not to the status logic.

On a supplier invoice really received through the platform, sending the "Approved" status (205):

- with the recipient address built from the **vendor SIREN** — what `CdarHandler::generateCdarFile()` does today on `main` — the platform answers `400` and rejects the CDAR;
- with the recipient address built from the **vendor routing ID**, the very same 205 is accepted (`acknowledgement: Ok`).

That also explains why EsaLink works: it does not enforce MDT-73, Super PDP does.

I checked the other candidate at the same time: the referenced document status code (MDT-88) sent as `45` (In process) instead of `1` (Accepted) for a 205. Super PDP accepts `45` without complaining, so that is not what breaks it.

[...]

PR #457, opened 2026-07-26
[einvoicing] Payment lifecycle statuses: conformant cash-in (212) and payment transmitted (211)

daGrumpf-bxp:
## Problem

Recording a payment on an invoice already transmitted to the platform makes the lifecycle message fail with a HTTP 400:

```
[BR-FR-CDV-14/MDT-207] : Si le statut est 'Encaissé' (MDT-105 = 212), ALORS il doit y avoir au moins 1 Bloc MDG-43
avec une valeur de MDT-207 = MEN et tous les blocs MDG-43 avec une valeur MDT-207 = 'MEN' doivent contenir une
valeur MDT-215 (Montant) et une valeur de MDT-224 (pourcentage de TVA).
```

The CDAR built for the status 212 carried no amount at all: `CdarHandler::generateCdarFile()` only knew about the invoice reference and the status code. Fixing that surfaced a second broken rule, `BR-FR-CDV-16` (MDT-124-2, the status detail sequence number), which the module only emitted when a reason code was present — so a 212 never had one. And once the message was schema-valid, the platform answered `no matching invoices found`: the CDAR declared **us as the buyer issuing the status to ourselves as the seller**, which is the mapping of a *supplier* invoice status, so the platform could not find the invoice the status was about.

Beyond the rejections themselves, reviewing the cash-in reporting against the reform showed **three gaps**, all closed here: the amount was missing (1), only fully paid invoices were reported (2), and the scope of the obligation was not taken into account at all (3).

## What this does

- **(1) Cashed amount (MDG-43 / MEN).** `CdarHandler::getCashedAmountCharacteristics()` spreads the cashed amount over the VAT rates of the invoice proportionally to their TTC weight — exact on a full payment, prorata otherwise, since Dolibarr does not attach a payment to given lines. The rounding residual is absorbed by the largest block so the blocks always sum up to what was really cashed. Generation now refuses to build a 212 it cannot fill instead of letting the platform reject it.
- **Sequence number (MDT-124-2).** Emitted for any status detail block, not only for the ones carrying a reason code.
- **Parties of the cash-in.** A 212 is the only status we send on one of our *own* invoices: issuer = our SIREN with role `SE`, recipient = the buyer SIREN with role `BY` and its routing ID (`0225`). Supplier invoice statuses keep the mapping they are already accepted with (us as `BY`, addressed to the vendor as `SE`). Two other details of the reference example follow, again only for the cash-in: referenced document status code `47` (Paid) instead of `45` (In process), and a plain date (format `102`) for its issue date.
- **(2) One status per cash-in.** The status moves from `BILL_PAYED` to `PAYMENT_CUSTOMER_CREATE`: the reform expects the amount of *every* payment, partial ones included, and an invoice that stays partially paid owes its statuses all the same. Write-offs (abandon / bad debt), where nothing is cashed, stay excluded.
- **(3) Scope options.** The payment data are only due for the operations whose VAT is exigible on collection: `EINVOICING_VAT_ON_DEBITS` (seller who opted for the "TVA d'après les débits" scheme: nothing to report at all) and `EINVOICING_PAID_STATUS_SERVICES_ONLY` (restrict to services and down payments). The latter is off by default: sending the optional status on a goods invoice is allowed, whereas missing a mandatory one is not, and a free text line typed as a product would be silently skipped.

## Testing

The generated message was validated against the UN/CEFACT **CDAR D22B XSD** and the full **FNFE `BR-FR-CDV` schematron** (`fnfempe/France_RFE`): 26 rules fired, 0 failed assert. Its structure was diffed against the official **XP Z12-012 annex B** examples (`UC1_F202500003_07-CDV-212_Encaissee.xml` for the seller-issued cash-in, `UC1 205/211` for the buyer-issued statuses). Both the full-payment and the partial-payment breakdowns were checked (blocks summing exactly to the cashed amount).

End to end: the cash-in of a real invoice was accepted by the platform (`acknowledgement: Ok`) after having been rejected on each of the three points above. `phan --quick` clean.

## Also in this PR: "Payment transmitted" (211)

Reading the same reference examples showed the mirror status was missing too. `211` is what a buyer sends to let the vendor know its invoice has been paid — the module knew the code but never sent it, and it was even filtered out of the sendable statuses so it could not be picked by hand either.

- Sent automatically when Dolibarr classifies a supplier invoice as paid, carrying the amount paid and its date in a MDG-43 block (`MDT-207 = MPA`), as in the `UC1 211` example. Only for invoices that really came from the platform, never for a write-off, and once per invoice (`EInvoicing::hasSentStatusMessage()`) since deleting then re-recording a payment classifies the invoice paid a second time. A failure never rolls the payment back.
- Optional status in the reform, and each one costs a platform flow, so it is **off by default**: `EINVOICING_SEND_PAYMENT_SENT_STATUS`.
- **Document status code (MDT-88)** now follows the lifecycle status it goes with — deposited `10`, received `43`, made available `48`, taken over `45`, approved `1`, payment transmitted and cashed in `47` — instead of always announcing "in process" (`45`), and the referenced invoice is dated with a plain date (format `102`). Both come straight from the annex B examples; the statuses those examples do not cover (refusal, dispute, suspension) keep the historical `45`.

Both the 211 and the 212 messages validate against the XSD and the schematron (25 and 26 rules fired, 0 failed assert).

One more defect surfaced while firing those triggers from the command line on dev instances, fixed here: the **electronic address (MDT-73) of a supplier invoice status** was the vendor SIREN, which the platform only accepts when the vendor is registered under that very address; anywhere else the status came back `L'adresse électronique (MDT-73) est invalide` (that is #410).

A status is a reply, so it is now addressed to where its vendor exchanges from, in this order: the routing recorded for that vendor in Dolibarr, then the **BT-34 seller electronic address of the e-invoice it sent us** (parsed on reception since forever, just never used for the reply, and read here from the XML already stored — no extra call), then the address the platform directory declares for that SIREN, then the SIREN as a last resort, with a warning naming the vendor instead of a bare HTTP 400. The comment below details how the three sources disagree on a real invoice, and measures the directory step on 177 production vendors.

The same session also turned up a missing-include fatal in the trigger file, unrelated to the payment statuses — it is sent separately in #459.


daGrumpf-bxp:
## Added: address the status to where the vendor said it exchanges

Follow-up on #410, which is the same lines as the MDT-73 fix already in this PR.

Addressing a status to the vendor routing ID only helps when a routing was recorded for that vendor. Without one, `getBuyerCommunicationURI()` guesses its SIREN, and the platform refuses the status unless the vendor is registered under that very address — an opaque HTTP 400 for something Dolibarr already knows: a lifecycle status is a **reply**, and the invoice being replied to carries the address its sender exchanges from, the **BT-34 seller electronic address**. The module has always parsed it on reception (`sellerCommunicationUri` in `CIIProtocol`); it was simply never used to address the reply.

### Resolution order

1. **a routing recorded in Dolibarr** for that vendor — a deliberate choice of the user, so it wins;
2. **the BT-34 of the e-invoice we received** — the vendor telling us where it exchanges. Read from the XML already stored for that supplier invoice, never fetched from the platform again; an invoice keyed by hand simply yields nothing;
3. **the platform directory** (`checkRecipientDirectory()`, already used by the recipient pre-check) — for a vendor whose e-invoice we do not hold;
4. **the SIREN** guessed by `getBuyerCommunicationURI()` — last resort, now with a warning in the log naming the vendor and saying the platform is likely to refuse the message.

Providers exposing no directory answer `unsupported` without any call, so step 3 costs them nothing.

### Why BT-34 comes before the directory

The directory lists the reception addresses declared for a SIREN and returns the first active one, which can be another address of the same company than the one it exchanges under. On a real received invoice here, the sources disagree:

```
BT-34 of the received CII (the address the vendor put on its own invoice) : 908544638_Status
platform directory, first active entry for the same SIREN                : 908544638
what was sent so far (SIREN guess)                                       : 908544638
```

`908544638_Status` is indeed the address that vendor publishes as its own.

### Step 3, measured on real data

For the vendors whose e-invoice we do not hold, the directory is what is left. Checked read-only on the vendors of two production instances — **177 vendors holding a 9-digit SIREN, none with a routing recorded in Dolibarr**:

| | vendors |
|---|---|
| declare an address that is **not** their bare SIREN → refused today | **8** |
| registered under their SIREN → identical address, no change | 149 |
| absent from the directory → SIREN fallback kept, warning logged | 20 |

```
SA ORANGE                        SIREN=380129866   declared=380129866_ACHATPUB_507807949
SAS METRO FRANCE                 SIREN=399315613   declared=399315613_ACHATSMARCHANDISES
GS1 France                       SIREN=305545675   declared=305545675_GS1FRANCE
AUTO 44 (EUROPCAR)               SIREN=871800637   declared=871800637_SAS_AUTO44
Parthenay Distribution           SIREN=318247863   declared=318247863_3020710001514
SAS DISTRIBUTION SANITAIRE CH.   SIREN=572141885   declared=572141885_57214188500143
SYMI EAUX DE VIENNE - SIVEER     SIREN=200049104   declared=200049104_20004910400017
Communauté de communes           SIREN=200041333   declared=200041333_20004133300069
```

The 149 identical ones are the non-regression: nothing moves for an address that already works. Foreign vendors fall in the "absent" bucket and keep the current behaviour.

Deployed and checked read-only on the instance that received the invoice above: generating the "Approved" status now addresses it to `908544638_Status`.

eldy:
IA is cool but providing PR with tons of files modified make the code completely no under control.

Can you split this in smaller part so we can validate the PR does not bring regressions with what already works.
A PR must be atomic.

daGrumpf-bxp:
Fair enough — this one carried five distinct subjects, that is too much to validate at once. Closing it in favour of five atomic PRs, each with its own reason to exist and its own way of being checked.

**Open now** — the two that are independent of everything else, and that fix statuses which already work today:

- **#476** — the electronic address (MDT-73) of a status sent on a supplier invoice: reply to where the vendor says it exchanges from (routing → BT-34 of the invoice received → platform directory → SIREN), instead of its SIREN which the platform refuses unless the vendor is registered under it. Fixes #410. *1 file + ChangeLog.*
- **#477** — the document status code (MDT-88) follows the lifecycle status it carries (deposited/received/available/taken over/approved/paid) instead of always announcing "in process", as in the XP Z12-012 annex B examples. *1 file + ChangeLog.*

**To come**, one at a time, each rebased once the previous is merged (they all touch `CdarHandler.class.php`, so opening them together would only give you rebase noise):

3. **Conformant cash-in (212)**: the cashed amount broken down by VAT rate (MDG-43/MEN, rule BR-FR-CDV-14), the status detail sequence number (BR-FR-CDV-16), and the parties the right way round (we are the seller of our own invoice, the status goes to the buyer). This is what makes the platform stop answering HTTP 400 / "no matching invoices found".
4. **One status per cash-in**: `BILL_PAYED` → `PAYMENT_CUSTOMER_CREATE` so partial payments are reported too, plus the two options that frame the scope of the obligation (`EINVOICING_VAT_ON_DEBITS`, `EINVOICING_PAID_STATUS_SERVICES_ONLY`).
5. **"Payment transmitted" (211)**: optional status, off by default (`EINVOICING_SEND_PAYMENT_SENT_STATUS`), sent once per supplier invoice classified as paid.

Nothing is dropped, and each of the five keeps its own validation: XSD D22B + FNFE `BR-FR-CDV` schematron for what it changes in the message, and for #476 the measurement on 177 real vendors showing what moves and what does not.

The branch of this PR is kept as-is for reference.

PR #461, opened 2026-07-27
FIX #395 emit the EXTENDED-CTC-FR guideline URN on the native XML path

### LINKED ISSUE #395: [EInvoicing] Le module ne produit jamais de XML conforme EXTENDED-CTC-FR sur le chemin natif
### Bug

Constat technique (issu d'un audit du module einvoicing) :
  - `CIIProtocol::buildXML()` (chemin natif et par défaut de génération XML Factur-X) utilise toujours l'URN de profil générique `urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended` (`EXTENDED`), jamais la variante française `EXTENDED-CTC-FR` (`urn.cpro.gouv.fr:1p0:extended-ctc-fr`). Une ligne commentée dans le code (`CIIProtocol.class.php:1653`) montre qu'une variante CTC-FR avait été envisagée pour le profil EN16931 puis abandonnée.
  - Un chemin de génération XML alternatif et déprécié (`EINVOICING_USE_EXTERNAL_FACTURX_BUILDER`) contient bien un patch capable de forcer l'URN vers `EXTENDED-CTC-FR` (`XmlPatcher::patchGuidelineId()`), mais ce patch est désactivé en commentaire, et de toute façon ce chemin n'est pas celui utilisé par défaut.
  - Pendant ce temps, le module annonce systématiquement à la plateforme (Esalink/SuperPDP) `"flowProfile": "Extended-CTC-FR"` dans les métadonnées d'envoi, quel que soit le contenu XML réel envoyé — désynchronisation entre ce qui est déclaré et ce qui est transmis, jamais contrôlée par le code.

Question ouverte (nécessite une décision métier/conformité) : le profil EXTENDED-CTC-FR doit-il être utilisé pour **toutes** les factures émises par ce module (mandat FR), ou seulement pour des cas spécifiques (avoirs, factures avec ligne d'acompte, qui ont besoin de l'extension `AdditionalReferencedDocument`) ? La réponse détermine si le chemin natif doit être corrigé pour toujours utiliser CTC-FR, ou seulement dans ces cas précis.

### Module

einvoicing

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

Steps to reproduce :
  1. Configurer `EINVOICING_USE_EXTERNAL_FACTURX_BUILDER` à sa valeur par défaut (désactivé) — c'est le chemin natif utilisé en production.
  2. Générer une facture Factur-X pour n'importe quelle facture (fournisseur ou client) via le module.
  3. Ouvrir le XML CII embarqué dans le PDF généré et chercher `ram:GuidelineSpecifiedDocumentContextParameter/ram:ID` : la valeur est
  `urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended` (profil "EXTENDED" générique).
     → Voir `class/protocols/CIIProtocol.class.php:1642-1663` : le tableau `$profileGuidelines` ne contient aucune entrée produisant l'URN `urn.cpro.gouv.fr:1p0:extended-ctc-fr` pour le profil utilisé par défaut.
  4. Comparer avec les métadonnées envoyées à la plateforme lors de la transmission (`class/providers/EsalinkPDPProvider.class.php:432`,
  `class/providers/SuperPDPProvider.class.php:609`) : le champ `flowProfile` déclaré est `"Extended-CTC-FR"`, systématiquement, sans lien avec le contenu XML réellement généré à l'étape 3.

  Expected behavior :
  Le profil déclaré à la plateforme (`flowProfile`) doit toujours correspondre exactement à l'URN de guideline réellement présent dans le XML transmis. Soit :
  - le XML natif doit être généré avec l'URN `EXTENDED-CTC-FR` (si ce profil est requis pour toutes les factures dans le cadre du mandat FR), soit
  - le `flowProfile` annoncé à la plateforme doit refléter le profil réel du document (`EXTENDED` la plupart du temps, `EXTENDED-CTC-FR` uniquement quand le XML le contient effectivement).

### Attached files

_No response_

daGrumpf-bxp:
Fixes #395.

## The problem

The native builder is the only XML path in use (`EINVOICING_USE_EXTERNAL_FACTURX_BUILDER` off), and it never emitted the French URN:

- `CIIProtocol::BUILD_XML_PROFILE = 'EN16931'` → `urn:cen.eu:en16931:2017`
- `FacturXProtocol::BUILD_XML_PROFILE = 'EXTENDED'` → `...#conformant#urn:factur-x.eu:1p0:extended`

while both providers declare `"flowProfile": "Extended-CTC-FR"` to the platform. The `'EXTENDEDFR'` entry @Daoud-mohamed added to the profile table was never selected by anything, so it stayed dead code.

@eldy, on your question — `patchGuidelineId()` does **not** need to be copied into the CII path. It only exists because horstoeko/zugferd writes the URN from its own profile enum, so the deprecated path had no way to change it other than rewriting the string afterwards. The native builder already has a profile table, so pointing the constant at the right key is enough. I left the commented call in `XmlPatcher` untouched and added a comment saying why it stays off.

@Daoud-mohamed — went with your "a single profile for all kind of invoices" proposal. Both protocols now use `EXTENDEDFR`.

## What the switch actually needed

Three conditions in `CIIProtocol` were gated on `$profile === 'EXTENDED'`. Two of them had to accept `EXTENDEDFR` (EXTENDED-CTC-FR is a *conformant* extension built on EXTENDED, so everything EXTENDED allows is allowed there):

- `InvoiceReferencedDocument/TypeCode` (BT-3)
- `InvoiceReferencedDocument/FormattedIssueDateTime` (BT-26)

Without that, moving the CII path off EN16931 would have *silently dropped* both nodes.

The third one is the opposite: the **line-level** `AdditionalReferencedDocument/FormattedIssueDateTime` on a deposit line is **forbidden** by EXTENDED-CTC-FR (`CII-FREXT-DT-027`). It stays gated on `EXTENDED` alone. I only found this by validating — worth flagging, it is the one non-obvious part of the change.

## Validation

Three invoices generated by the module on a dev instance (plain invoice, credit note, final invoice consuming a deposit), checked against the FNFE artefacts: CII D22B uncoupled XSD, `EXTENDED-CTC-FR-CII` schematron, and `BR-FR-Flux2-Schematron-CII`.

| Case | before (EN16931) | after (EXTENDED-CTC-FR) |
|---|---|---|
| plain invoice | XSD OK — 0 failed-assert | XSD OK — 0 failed-assert |
| credit note | XSD OK — **`BR-FR-CO-05/BT-3` KO** | XSD OK — 0 failed-assert |
| final invoice with a deposit line | XSD OK — 0 failed-assert | XSD OK — 0 failed-assert |

So the answer to "*switching will likely require fixing some validation errors*": no new error, and one **pre-existing** French rule failure gets fixed. A credit note used to reference the corrected invoice by number only:

```xml
<ram:InvoiceReferencedDocument>
  <ram:IssuerAssignedID>BQ-FA-2606-0003</ram:IssuerAssignedID>
</ram:InvoiceReferencedDocument>
```

`BR-FR-CO-05` wants the date too (BT-26), which the EN16931 gate was suppressing. Now:

```xml
<ram:InvoiceReferencedDocument>
  <ram:IssuerAssignedID>BQ-FA-2606-0003</ram:IssuerAssignedID>
  <ram:TypeCode>381</ram:TypeCode>
  <ram:FormattedIssueDateTime>
    <qdt:DateTimeString format="102">20260625</qdt:DateTimeString>
  </ram:FormattedIssueDateTime>
</ram:InvoiceReferencedDocument>
```

The whole XML diff is exactly that: the URN on every invoice, plus BT-3/BT-26 on the referenced document for credit notes and deposit-line invoices. Nothing else moves.

## Also in here

`sendSampleInvoice()` in both providers declared `flowProfile: "CIUS"` for a document produced by the same builder — a third value for the same thing. Aligned on `Extended-CTC-FR`.

## Not in here

- **No unit test.** The module has no test fixture for XML generation, and building a usable `$invoiceData`/`$linesData` fixture would have been a much larger diff than the fix. The validation above is reproducible with the FNFE public artefacts; happy to add a test if you would rather have one.
- `einvoicing/assets/samples/SPECIMEN-doli_*.cii.xml` still show the old URN. They are not referenced by any code, so I left them alone rather than add churn — say the word if you want them refreshed.
- The two schematron warnings that show up on the reference sample regardless of profile (`CII-SR-236` FaxUniversalCommunication emitted, `CII-SR-450` both buyer ID and Global ID emitted) are unrelated pre-existing issues, not touched here.


Daoud-mohamed:
@daGrumpf-bxp 
The horstoeko/zugferd library does not support the EXTENDEDFR (extended-ctc-fr) profile.

If this profile is declared, the library fails during XML validation before generating the Factur-X file. Switching to EXTENDEDFR would therefore require fixing additional CII errors/warnings and patching the horstoeko/zugferd library output for Factur-X.

The library is currently used both to generate Factur-X files when EINVOICING_USE_EXTERNAL_FACTURX_BUILDER or to merge the CII XML with the PDF, but it validates the XML in both cases and rejects the unsupported extended-ctc-fr profile.

**NB:** The EXTENDEDFR (extended-ctc-fr) profile has been added for testing purposes to prepare for switching to this profile.

eldy:
> @daGrumpf-bxp 
> The horstoeko/zugferd library does not support the EXTENDEDFR (extended-ctc-fr) profile.
> 
> If this profile is declared, the library fails during XML validation before generating the Factur-X file. Switching to EXTENDEDFR would therefore require fixing additional CII errors/warnings and patching the horstoeko/zugferd library output for Factur-X.
> 
> The library is currently used both to generate Factur-X files when EINVOICING_USE_EXTERNAL_FACTURX_BUILDER or to merge the CII XML with the PDF, but it validates the XML in both cases and rejects the unsupported extended-ctc-fr profile.
> 
> **NB:** The EXTENDEDFR (extended-ctc-fr) profile has been added for testing purposes to prepare for switching to this profile.

@daGrumpf-bxp according to  Mohamed feedback that we can summarize into "en16931 profile seems enough, no need for extended".
If you agree with that, to not loose knowledge and work done for extenended profile, can you add in code a condition on the constant BUILD_XML_PROFILE == 'EXTENDEDFR';

 to switch to code to manage extendefr profile in an optional way ? So we can switch from mode en16931 to extendedfr just by switching one constant and may be in a future we can suggest option if we need.

daGrumpf-bxp:
@Daoud-mohamed you are right, and my "no new error" was wrong. I have pushed 5bc99e1 which does what @eldy asks.

### Where my validation failed

I only ever exercised the CII protocol. Every instance I test on runs `EINVOICING_PROTOCOL=CII`, which writes a standalone XML and never goes near the library — so the whole campaign was blind to the Factur-X path. Reproduced properly now:

```
Dolibarr  : 24.0.0-beta
FPDF/TCPDF: zugferd merger branch
THROWN horstoeko\zugferd\exception\ZugferdUnknownProfileException:
        Cannot determain the profile by urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr
```

And the precise mechanism, for the record: it is not an XML validation, it is `ZugferdProfileResolver::resolve()` matching the guideline URN against `ZugferdProfiles::PROFILEDEF` to obtain three things — the attachment filename, the XMP conformance level and the XMP version. `PROFILEDEF` only knows `urn:factur-x.eu:1p0:extended` and `urn:zugferd.de:2p0:extended`, so CTC-FR throws. That distinction matters, because it is what makes a fix possible without touching the library.

### What 5bc99e1 changes

**Defaults go back to what they were** — `EN16931` for CII, `EXTENDED` for Factur-X. No behaviour change out of the box.

**The profile becomes a variable**, @eldy, as you asked. `getBuildXmlProfile()` returns the class constant unless `EINVOICING_XML_PROFILE` overrides it; an unrecognised value is logged and ignored rather than aborting. So the whole CTC-FR handling is retained and switching to it is one setting, ready to be surfaced in the setup screen when the mandate calls for it:

```
EINVOICING_XML_PROFILE=EXTENDEDFR  →  urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr
EINVOICING_XML_PROFILE=BOGUS       →  urn:cen.eu:en16931:2017   (default, with a syslog warning)
```

**`CtcFrPdfMerger` makes that switch usable on the Factur-X side.** It extends `ZugferdDocumentPdfMerger` and overrides only the three `getXmlAttachment*()` methods, each of which defers to the parent and falls back **only** on `ZugferdUnknownProfileException`. The fallback reuses the library's own EXTENDED definition rather than hardcoded strings. Every profile the library knows keeps its own parameters, and the class becomes a no-op the day `extended-ctc-fr` is added upstream.

The reasoning for reusing EXTENDED: CTC-FR is a *conformant extension* of EXTENDED, and the Factur-X XMP `ConformanceLevel` enumeration has no CTC-FR value. I checked the FNFE XP Z12-012 sample PDFs — `ConformanceLevel` tracks the profile name (`EN 16931`, `EXTENDED`), `DocumentFileName` is always `factur-x.xml`, `Version` always `1.0` — **but none of their sample PDFs carries the CTC-FR guideline**, so this is reasoned, not attested. Worth a confirmation from FNFE or a real PDP acceptance before anyone flips the switch in production.

Result on a Dolibarr 24.0 instance, `EINVOICING_XML_PROFILE=EXTENDEDFR`, Factur-X protocol:

```
embedded guideline  : urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr
XMP ConformanceLevel: EXTENDED | DocumentFileName: factur-x.xml | Version: 1.0
```

**BT-26 is no longer gated on the profile.** The issue date of the corrected invoice is core EN16931, EN16931 accepts it, and `BR-FR-CO-05` requires it on a credit note. Only the referenced-document `TypeCode` stays EXTENDED-only, because EN16931 forbids it (`CII-DT-018`). So that conformance fix now lands on the default profile too, which it would not have done with a plain revert.

### Validation

Plain invoice, credit note, final invoice consuming a deposit, invoice with a line discount — CII D22B XSD + `EN16931-CII` / `EXTENDED-CTC-FR-CII` + `BR-FR-Flux2-CII`:

| | default (EN16931) | `EINVOICING_XML_PROFILE=EXTENDEDFR` |
|---|---|---|
| plain invoice | 0 | 0 |
| credit note | 0 *(was `BR-FR-CO-05`)* | 0 |
| deposit line | 0 | 0 |
| line discount | 0 | 0 |

XML on Dolibarr 23.0.3, merger path on 24.0.0-beta.

### Two things left open

1. **`flowProfile` still does not match.** Both providers hardcode `"flowProfile": "Extended-CTC-FR"` while the default XML is EN16931 / EXTENDED — which is the desynchronisation @atm-maxime opened #395 about, and rolling the default back leaves it in place. The clean answer is to derive `flowProfile` from the profile actually used, so one setting moves both. I have not done it here because it changes what we declare to a live platform (`CIUS` instead of `Extended-CTC-FR` by default) and that deserves a deliberate decision — tell me if you want it and I will add it.

2. **Reading back is still blocked.** `ZugferdDocumentPdfReader::readAndGuessFromFile()` throws the same exception on a CTC-FR Factur-X PDF, and the inbound path calls it unconditionally (`FacturXProtocol.class.php:1170`) even when the module's own parser is used. So under the mandate a partner sending CTC-FR Factur-X cannot be ingested — independently of what we ourselves emit. Happy to open a separate issue; the cheap part of the fix is to build that reader only inside the `EINVOICING_USE_EXTERNAL_FACTURX_READER` branch that actually consumes it.

Longer term the real fix for both is `extended-ctc-fr` landing in `ZugferdProfiles::PROFILEDEF` upstream. I can send that PR to horstoeko/zugferd if you think it is worth it.

daGrumpf-bxp:
I have split this, @eldy — a394f47. Re-reading your comment I had answered it and then some: the branch had grown to five changes for a review that asked for one, and the rollback had left three statements behind that were no longer true (`buildXML()`'s own default still said EXTENDEDFR, the sample-invoice `flowProfile` carried a comment claiming it matched the emitted URN, and `XmlPatcher` claimed the native builder emits CTC-FR directly).

**This PR is now only what you asked for**, +40 lines in one file plus a comment:

- the three places `buildXML()` branches on the profile handle `EXTENDEDFR`, so switching `BUILD_XML_PROFILE` is enough to emit EXTENDED-CTC-FR;
- **no default changes**, no new setting, no library work.

One thing I kept that you did not ask for, because a plain revert would have regressed it: **BT-26 is no longer gated on the profile**. The issue date of the corrected invoice is core EN16931, EN16931 accepts it, and `BR-FR-CO-05` requires it on a credit note — so credit notes generated with the *current default* are refused by the French schematron today. Only the referenced-document `TypeCode` stays EXTENDED-only, since EN16931 forbids it (`CII-DT-018`). Verified: credit note goes from `BR-FR-CO-05` to 0 failed-assert on the default profile.

Everything else moved to **#474**: the `EINVOICING_XML_PROFILE` setting (your "may be in a future" option — drop it there if it is too early) and `CtcFrPdfMerger`, which is what makes the switch usable on Factur-X given the library limitation @Daoud-mohamed pointed out.

The `flowProfile` change I had slipped in is gone entirely. It only made sense while the default was being switched, and the underlying desynchronisation — providers hardcoding `"Extended-CTC-FR"` while the XML says EN16931, which is what #395 actually reports — is still open. It needs its own decision since it changes what we declare to a live platform; happy to take it on once you tell me which way.

daGrumpf-bxp:
Superseded by #474, which you merged as 3961c30.

Everything this PR carries is now in `main`, verbatim:

- `CIIProtocol::isExtendedProfile()` and the `EXTENDEDFR` guideline URN
- `ram:TypeCode` of the referenced document gated on the EXTENDED-level profiles (CII-DT-018)
- `ram:FormattedIssueDateTime` (BT-26) emitted whatever the profile, for BR-FR-CO-05
- the `XmlPatcher` comment on the deprecated external builder path
- the four ChangeLog entries

Checked mechanically: the net diff of this branch reverse-applies cleanly on `main`
(`git apply -R --check`), i.e. there is zero delta left. Rebasing it would produce an empty PR,
so I am closing it rather than leaving a conflicting one in your queue.

Closing — nothing to review here.

PR #465, opened 2026-07-27
FIX #354 : Do not abort the creation of an invoice when its e-invoice status cannot be recorded

### LINKED ISSUE #354: [EInvoicing] erreur à la génération de factures par tache planifié depuis un modèle de facture

daGrumpf-bxp:
Fixes #354.

## Problem

The `BILL_CREATE` trigger records the e-invoice status of the new invoice and returns `-1` when that write fails, which aborts the creation of the invoice itself.

A database left un-migrated after a module update is enough to trigger it: the module SQL then writes a column its own table does not have yet — `Unknown column 'ap_precheck_status' in 'field list'` — and Dolibarr can no longer create **any** invoice, by hand or from a recurring template, until the module is disabled. That is exactly what #354 reports (the migration itself is a separate matter, replaying the module SQL fixes it).

## What this does

Logs the failure with its database error and lets the creation through.

What is recorded at that point is the e-invoice status of a *draft*, and `BILL_VALIDATE` computes it again when the invoice is validated — an unknown or not-generated status is resolved there. Losing that row is not worth losing the invoice, all the more so as the module is the one that cannot write to its own table.

`BILL_VALIDATE` is deliberately left as it is: at validation the status is about to matter.

## Testing

`php -l` and `phan --quick` clean.

### DIFF
--- a/einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php
+++ b/einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php
@@ -136,11 +136,21 @@ class InterfaceEInvoicingTriggers extends DolibarrTriggers
 					$statustouse = $einvoicing->needEInvoiceManagement($object);
 				}
 
-				// When invoice is created
+				// When invoice is created.
+				// A failure here must not abort the creation of the invoice: what is recorded is the
+				// e-invoice status of a draft, which BILL_VALIDATE computes again when the invoice is
+				// validated (an unknown or not-generated status is resolved there). Refusing the
+				// creation instead leaves Dolibarr unable to create any invoice at all -- that is what
+				// a database left un-migrated after a module update does, the module SQL then having a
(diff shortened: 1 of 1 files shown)


Daoud-mohamed:
For me, raising an error and aborting the synchronization is the right behavior, because creating a supplier invoice without recording it in the einvoicing_extlinks table would break the entire workflow and leave the invoice orphaned.

daGrumpf-bxp:
Thanks for the review. Small precision first: this is `BILL_CREATE` on a customer invoice (`Facture`). The trigger has no `BILL_SUPPLIER_CREATE` case, and nothing here goes near the flow synchronization or the supplier invoice import — that path is #479.

On the substance you are right, though: the trigger is the wrong place for this. The write fails because `llx_einvoicing_extlinks` has no `ap_precheck_status` column on the install that reported #354, so the fix belongs to the schema, which is what #407 addressed. Hardening the trigger only makes the symptom quieter.

Closing. The discussion continues on #354.

PR #469, opened 2026-07-27
CHORE remove three stale hand-captured module outputs from assets/samples

daGrumpf-bxp:
Housekeeping, no code change.

`einvoicing/assets/samples/` holds three outputs of the module captured by hand in June — `SPECIMEN-doli_ok.cii.xml`, `SPECIMEN-doli_ko_transfer.cii.xml` and `INVTEST-dol_ok.facturx.pdf` (whose embedded `factur-x.xml` is a fourth copy of the same thing). Nothing in the code, the tests or the documentation references them.

They no longer match what the module produces. All three still carry:

- the `urn:cen.eu:en16931:2017` guideline;
- a `ram:ID` **and** a `ram:GlobalID` on both parties, which c76b5a4 made mutually exclusive in `buildParty()`.

So running a schematron over them today reports `CII-SR-450` against a defect the code no longer has. That is not hypothetical: it happened while auditing #395 and produced a wrong bug report, which I had to retract.

A sample committed as a file has to be refreshed by hand at every change of the generated XML, and it will drift again. Removing the three; git history keeps them for anyone who wants the old shape, and if a reference document is wanted it is better produced by a test than pinned in the tree.

Left in place on purpose: `CrossIndustryInvoice_..._100pD16B.xsd` and `Example-Discount-Line.txt` — a schema and a developer note, neither of which claims to be module output.

Unrelated but noticed while inventorying, not touched here: the four files directly under `einvoicing/assets/` (`00_AdditionalDocument.csv`, `00_ZugferdDocumentPdfBuilder_PrintLayout.pdf`, `03_ZugferdDocumentReader_ImportFile_EN16931.xml`, `03_ZugferdDocumentReader_ImportFile_XRECHNUNG3.xml`) are byte-identical copies of `vendor/horstoeko/zugferd/examples/assets/`, and only the vendor's own example scripts read them — from the vendor directory. Happy to drop them in a separate PR if you want.


eldy:
We need this files as it is used by a lot of user to make some tests currentl. I prefer to cancel this PR completely.

Issue #487, opened 2026-07-28
An invoice classified as paid declares BT-23 = B2 while the amounts say it is not, and is refused (BR-FR-CO-09)


daGrumpf-bxp:
### Bug

An invoice classified as paid in Dolibarr is declared to the platform as "already paid" (BT-23 = `B2`/`S2`/`M2`) even when the generated document says the opposite about the money, which the French CTC-FR schematron rejects with two fatal rules.

`CommonProtocol::getBillingProcessID()` picks the `2` suffix from the invoice status alone:

```php
// einvoicing/class/protocols/CommonProtocol.class.php:104
if ($invoice->status == Facture::STATUS_CLOSED && empty($invoice->close_code)) {
    return $prefix . '2';
}
```

Nothing checks that the amounts back that claim up. On an invoice flagged `paye = 1` with no payment recorded, the module emits:

```xml
<ram:BusinessProcessSpecifiedDocumentContextParameter><ram:ID>B2</ram:ID></ram:BusinessProcessSpecifiedDocumentContextParameter>
...
<ram:GrandTotalAmount>360.00</ram:GrandTotalAmount>
<ram:TotalPrepaidAmount>0.00</ram:TotalPrepaidAmount>
<ram:DuePayableAmount>360.00</ram:DuePayableAmount>
```

**Violated rules** — `BR-FR-Flux2-Schematron-CII` V1.4.0, pattern `BR-FR-CO-09`, both flagged **fatal**:

- `BR-FR-CO-09_BT-23-1` — if BT-23 is `B2`, `S2` or `M2`, the paid amount (BT-113) shall equal the total including VAT (BT-112)
- `BR-FR-CO-09_BT-23-2` — if BT-23 is `B2`, `S2` or `M2`, the amount due for payment (BT-115) shall be 0

**Impact.** The document is refused. This is not an exotic state: "classer payée" without recording a payment is ordinary practice, and the Dolibarr deposit workflow reaches it on its own — a deposit invoice consumed as a discount on the final invoice is marked paid with no payment line.

**Scope observed.** 12 documents out of a 72-document conformance run fail, and they are the 12 declinations of the same invoice: both Dolibarr versions tested (23.0.3 and 24.0.0-beta), all three XML profiles (`EN16931`, `EXTENDED`, `EXTENDEDFR`), both transports (standalone CII and Factur-X). So it depends neither on the version, nor on the profile, nor on the protocol. Every other case of the run passes.

[...]

PR #488, opened 2026-07-28
FIX #485 carry the preceding invoice issue date (BT-26) on every profile

### LINKED ISSUE #485: Missing FormattedIssueDateTime (BT-26) in InvoiceReferencedDocument for credit notes CII output
### Bug

When generating a credit note, the CII XML produced by CIIProtocol is missing the referenced invoice's issue date (FormattedIssueDateTime / BT-26) inside the InvoiceReferencedDocument block.

**Root cause:**
In CIIProtocol.class.php:1719-1729, the TypeCode and FormattedIssueDateTime elements are only added when $profile === 'EXTENDED'.
However, CIIProtocol::generateXML() (line 293) calls buildXML() with profile 'EN16931':
$xmlcontent = $this->buildXML($invoiceData, $linesData, 'EN16931', $outputlangs);
The EXTENDED profile is only used from FacturXProtocol::generateXML() (line 141). For standalone CII (.xml), the TypeCode and date are never included.

**Violated rules:**
- BR-FR-CO-05 (France): the reference BT-25 + date BT-26 pair is mandatory in InvoiceReferencedDocument
- EN 16931: FormattedIssueDateTime is required for referenced documents

**Generated XML (current, invalid):**
```xml
<ram:InvoiceReferencedDocument>
  <ram:IssuerAssignedID>IN2607-0003</ram:IssuerAssignedID>
</ram:InvoiceReferencedDocument>
Expected XML (valid):
<ram:InvoiceReferencedDocument>
  <ram:IssuerAssignedID>IN2607-0003</ram:IssuerAssignedID>
  <ram:FormattedIssueDateTime>
    <qdt:DateTimeString format="102">20260215</qdt:DateTimeString>
  </ram:FormattedIssueDateTime>
</ram:InvoiceReferencedDocument>
```
**Impact:** Sending credit notes to the PDP (Partner Dematerialization Platform) fails validation because BR-FR-CO-05 is not satisfied.
**Proposed fix:**
Remove the $profile === 'EXTENDED' condition so that TypeCode and FormattedIssueDateTime are always included in InvoiceReferencedDocument when available, as these fields are required by EN 16931 and French regulations.

### Module

einvoicing

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

Steps to reproduce:
1. Enable the EInvoicing module in Dolibarr
2. Create an invoice (e.g. IN2607-0003) and validate it
3. Create a credit note (avoir) linked to that invoice
4. Regenerate the CII e-invoice XML via the module (e.g. using 'Regenerate' action)
5. Open the generated _cii.xml file
6. Search for InvoiceReferencedDocument in the XML

Expected behavior:
The InvoiceReferencedDocument block should contain both:
- ram:IssuerAssignedID (BT-25) -> present
- ram:FormattedIssueDateTime (BT-26) -> should be present with format="102" and the invoice date in YYYYMMDD

Actual behavior:
Only ram:IssuerAssignedID is present. The date node is missing entirely, causing PDP validation to reject the credit note with rule BR-FR-CO-05.

### Attached files

_No response_

daGrumpf-bxp:
Fixes #485.

A credit note references the invoice it corrects with an `InvoiceReferencedDocument` carrying that invoice number (BT-25) and its issue date (BT-26). The date was only emitted on the `EXTENDED` profile, so the default output of the module, `EN16931`, sent the reference alone.

### Why it is a rejection, not a detail

`BR-FR-Flux2-Schematron-CII` V1.4.0, pattern `BR-FR-CO-05`, flag **fatal**:

```xml
<let name="headerRefCount"
     value="count($headerRefs[ram:IssuerAssignedID and ram:FormattedIssueDateTime/qdt:DateTimeString])"/>
<assert test="not($typeCode = ('261','381','396','502','503')) or ($headerRefCount ge 1 or $lineRefsValid)"
        flag="fatal" id="BR-FR-CO-05_BT-3">
```

A reference without a date is not counted. The module emits BT-3 = `381` for a credit note, so `headerRefCount` is 0 and the platform refuses the document.

### Why the date belongs to every profile, and the type code does not

Both come from the CEN artefact shipped with the FNFE package:

- `CII-DT-027` (fatal) — `not(ram:FormattedIssueDateTime) or self::ram:InvoiceReferencedDocument`. EN 16931 bans `FormattedIssueDateTime` on a referenced document **and exempts `InvoiceReferencedDocument`**. The date is core, not an extension.
- `CII-DT-018` (fatal) — a `TypeCode` is only tolerated on an `AdditionalReferencedDocument` (50/130/916). Emitting it here under EN 16931 is itself a fatal error, which is why it keeps a profile condition.
- `CII-EXT-DT-018` in `EXTENDED-CTC-FR-CII.sch` adds `or self::ram:InvoiceReferencedDocument`, so the type code is legal on the EXTENDED-level profiles.

The type code condition is therefore kept and widened to `EXTENDEDFR`: that profile, made selectable by #474, is EXTENDED-level too and was losing both nodes. The `isExtendedProfile()` helper spells that out in one place.

### Validation

72 documents generated on two instances — Dolibarr **23.0.3** and **24.0.0-beta** — covering 6 cases (plain invoice, credit note, deposit invoice, final invoice with a deposit line, line discount, document-level discount) × 3 profiles (`EN16931`, `EXTENDED`, `EXTENDEDFR`) × 2 transports (standalone CII, XML extracted back out of the Factur-X PDF). Each file was validated against the CII D22B XSD and against the ruleset matching the guideline URN it carries (`EN16931-CII-validation`, `Factur-X_EXTENDED` or `EXTENDED-CTC-FR-CII`), plus `BR-FR-Flux2-CII`.

All 12 credit note documents pass with 0 failed-assert, on both Dolibarr versions and all three profiles.

Negative control, to show the check is able to fail — same credit note with the date removed, i.e. what the current code produces:

```
avoir, guideline urn:cen.eu:en16931:2017                          BR-FR-CO-05_BT-3 [fatal]
avoir, guideline urn:...urn.cpro.gouv.fr:1p0:extended-ctc-fr      BR-FR-CO-05_BT-3 [fatal]
```

The rest of the campaign is clean apart from one unrelated case, the deposit invoice, which fails `BR-FR-CO-09` on every profile and both versions — reported separately in #487, untouched here.


daGrumpf-bxp:
## Tested on a live instance, against the platform validator

Generated on a Dolibarr instance connected to the SuperPDP sandbox, then submitted to its validation service (FNFE `CII_D22B` XSD + the CEN and `BR-FR-Flux2` schematrons). Two invoices were used, both created by the module's own test helper:

- `BQ-AV-2607-0001` — a **credit note** linked to the invoice it corrects: the case of the issue;
- `BQ-FA-2607-0004` — a final invoice consuming a **deposit invoice**, which carries a referenced document at header level *and* one at line level.

### The credit note, before and after

| code | conformance level | verdict |
|---|---|---|
| main | `urn:cen.eu:en16931:2017` | **invalid** — `BR-FR-CO-05_BT-3` |
| this PR | `urn:cen.eu:en16931:2017` | **valid**, 162 checks, 0 failure |

The rule fires exactly as the issue describes, and its message says why the reference alone is not enough:

```
BR-FR-CO-05/BT-3 : Si le type de facture (BT-3) est un avoir (261, 381, 396, 502, 503), alors :
  - soit au moins une référence à une facture antérieure (BT-25) avec sa date (BT-26) doit être présente
    au niveau entête, [...]
  Références entête trouvées : 0.
```

A reference without its date is not counted at all — which is why every credit note generated with the default profile was refused.

### The three profiles, with this PR

| profile forced | conformance level written | validator |
|---|---|---|
| `EN16931` | `urn:cen.eu:en16931:2017` | valid, 193 checks, 0 failure |
| `EXTENDED` | `urn:factur-x.eu:1p0:extended` | valid, 201 checks, 0 failure |
| `EXTENDEDFR` | `urn.cpro.gouv.fr:1p0:extended-ctc-fr` | valid, 197 checks, 0 failure |

The document-level `TypeCode` stays out of the EN 16931 output (`CII-DT-018`) and is present in the two EXTENDED ones, as intended.

### One thing this PR deliberately does not do

The `FormattedIssueDateTime` of the **line-level** `AdditionalReferencedDocument` (the deposit invoice a line refers to) is still gated on the EXTENDED profile alone. Extending it to EXTENDED-CTC-FR looks symmetrical but is wrong — measured, not assumed:

```
[CII-FREXT-DT-027] - FormattedIssueDateTime should not be present     flag="fatal"
test="not(ram:FormattedIssueDateTime) or self::ram:InvoiceReferencedDocument
      or self::ram:ContractReferencedDocument or self::ram:BuyerOrderReferencedDocument
      or self::ram:DespatchAdviceReferencedDocument"
location=.../IncludedSupplyChainTradeLineItem/SpecifiedLineTradeSettlement/AdditionalReferencedDocument
```

EXTENDED-CTC-FR is *stricter* than Factur-X EXTENDED on that element: the date is allowed at header level (`InvoiceReferencedDocument`, which is what this PR fixes) and forbidden on a line. Removing that single date from the same document makes it valid again, so it is the only cause. The current gating is therefore correct as it stands.

PR #489, opened 2026-07-28
[einvoicing] NEW: a screen to manage the vendor product references created by the mapping

daGrumpf-bxp:
Ok, the feature of #455 is nice, but not ultra simple to maintain in daily use, so we add/introduce this management page which should really simplify the handling of the links between the provider product and the Dolibarr product.

Concretely: #455 writes the mapping as a **vendor reference** on the product (`llx_product_fournisseur_price`), which is the right place - it is the first criteria the import of a supplier invoice uses to find the product of a line. But once written, that link could only be read **one product at a time**, in the "Supplier prices" tab of each product. There was no way to see what had been mapped, and no way to correct a mapping without walking the catalogue product by product.

## The page

`einvoicing/vendorref_list.php`, reachable from Billing > E-invoice synchronization > *Mapped vendor references*, and from a button on the mapping screen of #455.

- lists every vendor reference: vendor, reference, the vendor wording as received, the Dolibarr product, the buying price, creation and last change;
- filters on vendor, reference and product, with the usual sort and pagination;
- **reassigns** a reference to another product, renames it, or deletes the mapping - per line or in bulk.

Nothing is added to the schema: the page reads and writes the table the module already uses.

## Two traps worth mentioning

**Moving a reference is not one call.** `update_buyprice()` deletes the previous line only when the reference stays identical (it deletes by vendor + reference + min quantity before inserting). A *rename plus a move* therefore left the old line behind, and the same vendor reference resolved to two products. `einvoicingRemapVendorRef()` (new `lib/einvoicing_vendorref.lib.php`) updates in place when the product does not change - keeping the price history - and deletes then reinserts when it does. A rename onto a reference already used by the same vendor is refused rather than silently dropping the other mapping, which the unique key `(ref_fourn, fk_soc, quantity, entity)` would otherwise have caused.

**A vendor reference is bought, not sold.** Only products flagged "to buy" are offered, which is also what the first commit fixes on the screen of #455. A mapping whose product is no longer flagged that way is still listed, carries the core "not for purchase" badge, and states what to do: flag the product back on its card, or move the reference onto another product. Same for a mapping left by a deleted vendor, and for an instance holding no purchasable product at all - the screen says so instead of showing an empty combo.

## Commits

1. **FIX** - the picker of the mapping screen filtered on `p.tosell`, so it offered nothing on a purchase-only catalogue, which is exactly the catalogue it is made for (a product the module creates on import has `status = 0`, `status_buy = 1`). Measured on a 294-product instance: 113 products offered before, 150 after.
2. **NEW** - the page itself.

## Checks

`phan` (CI config, `--quick`) and `php -l` clean. Exercised on a copy of a production database (296 products, 12 real mappings) with a harness calling the library directly: rename in place, move, move + rename, duplicate guard, deletion, and the "mapped product flipped to not-for-purchase then remapped then restored" scenario - 24 assertions, database left byte-identical.

Nothing on the outbound side is touched: the page and the library only read and write buying prices, no protocol or document builder is involved.

Follow-up of #455.


eldy:
Can you provide a screnshot of the new screen ?

daGrumpf-bxp:
Sure, here it is, on a dev instance with demo data (Dolibarr 23.0.3).

**The list** - Billing > E-invoice Management > *Mapped vendor references*: every vendor product reference recorded on the products, with the vendor, the reference, the wording the vendor sent, the Dolibarr product it resolves to, the buying price and the dates. Filters on vendor / reference / product, the usual sort and pagination, deletion per line or in bulk.

Mapped vendor references - the list

**Inline edit** - the pencil opens the line: the reference can be renamed and reassigned to another product, then saved.

Mapped vendor references - inline edit

eldy:
Question: On screen "Mapped vendor references", if you delete a line, you delete the record line from llx_prouct_fournisseur_price ?
So it is a direct page to edit supplier prices of a product ?

Also can you provide a screenshot of the other menu entry "Map the product of an einvoice" ?

daGrumpf-bxp:
Answers to both questions, then the screenshot you asked for.

## 1. Yes, deleting a line does delete the record from `llx_product_fournisseur_price`

`vendorref_list.php:185-196` calls `ProductFournisseur::remove_product_fournisseur_price($rowid)`, the same entry point as the "Supplier prices" tab of the product card. Nothing is flagged as deleted or kept anywhere else: the row is gone. Same for the bulk deletion (`:199-217`).

## 2. "So it is a direct page to edit supplier prices of a product?"

Not quite, but you are pointing at something real.

What the page does **not** do: it never edits a price. The buying price and the quantity are shown read-only, only to identify the row. The only actions are on the **vendor reference**: rename it, reassign it to another product, delete it. That reference is the link the module writes on import and uses as the first matching criteria, not a pricing figure.

What is **right in your remark**: the list currently shows *every* `llx_product_fournisseur_price` row carrying a `ref_fourn`, not only the ones the einvoicing mapping created. A reference typed by hand in the product card shows up there too. The screen title promises more scoping than the code applies.

The way forward looks like the one you gave @spsolauv on #491: `llx_einvoicing_extlinks` already tracks what the module created or synchronized (`element_type = 'product'` for the products coming from an e-invoice). Recording the vendor reference mapping there as well - `element_type = 'product_fournisseur_price'`, `element_id` = the rowid, `syncref` = the vendor reference from the XML - would let the screen show only what the module actually created, with a checkbox to display everything else when needed.

I can add it to this PR, or in a separate one if you would rather keep this one about the screen itself. Your call.

## 3. Screenshot of "Map the products of an e-invoice"

Produced from a real sandbox flow rather than from crafted data: an invoice was issued from a second Dolibarr instance to this one through the platform, with five lines whose references none exist on the receiving side, and the automatic creation of products disabled. The synchronization then stops on the first unknown line and points at this screen:

Map the products of an e-invoice

Flow `i_174371`, invoice `TRI-FA-2607-0002`, 5 lines to map. For each line: the vendor product reference, the buyer reference and the global ID when the XML carries them, the label, the quantity with its unit code, the net unit price and the VAT rate - then the Dolibarr product picker. "Save the mapping" writes the vendor references, and the next synchronization finds the products by itself.

And for the record, the list questions 1 and 2 are about:

Mapped vendor references

daGrumpf-bxp:
@eldy @Daoud-mohamed 👋 Rebased on `main`, ready when you are.

I would really like your opinion on the implementation itself: does the approach look right to you, would you have done it another way, or is something missing? 🙂

The one point I am least sure about is `einvoicingRemapVendorRef()` in the new `lib/einvoicing_vendorref.lib.php`: it updates the row in place when the product does not change and deletes + reinserts when it does, because `update_buyprice()` only removes the previous row when the reference stays identical (rename + move otherwise leaves two rows for the same vendor reference). If there is a core helper that already does that properly, I would rather call it than keep my own.

By the way, `phan` is red here but not because of this PR: it fails on `main` too (same job, same single finding) since cec18cb, on a leftover `use Stripe\OAuthErrorObject;` at the top of `einvoicing/class/utils/XmlPatcher.class.php` - `PhanUnreferencedUseNormal`. Happy to send a one-line PR removing it if you want 🧹

daGrumpf-bxp:
Follow-up on the red `phan`: it is fixed by #522 (one line). `phan` is green there ✅ — once it lands, a rerun here goes green too, nothing to change in this PR.

PR #491, opened 2026-07-28
[Einvoicing] Add extrafield to societe (import type)

spsolauv:
Linked to PR #440 but dedicated PR required for database modification (https://github.com/Dolibarr/dolibarr-community-modules/pull/440#discussion_r3665185075)

### DIFF
--- a/einvoicing/core/modules/modEInvoicing.class.php
+++ b/einvoicing/core/modules/modEInvoicing.class.php
@@ -618,6 +618,10 @@ class modEInvoicing extends DolibarrModules
 		// 	1
 		// );
 
+		// Societe extrafields
+		// Use a <select> to have the default null value (falling back to default supplier invoice import type parameter set in the module settings)
+		$result = $extrafields->addExtraField('einvoicing_supplier_invoice_lines_import_type', $langs->trans('SupplierInvoiceLinesImportType'), 'select', 150, '255', 'societe', 0, 0, '', 'a:1:{s:7:"options";a:2:{i:'.Einvoicing::SUPPLIER_INVOICE_LINES_IMPORT_AUTO.';s:11:"automatique";i:'.Einvoicing::SUPPLIER_INVOICE_LINES_IMPORT_MANUAL.';s:6:"manuel";}}', 1, '', '1', '0', '', '', 'einvoicing@einvoicing', '1', 0, 0);
+
 		// Invoice extrafields
 		// Chorus fields
 		// TODO : Remove Chorus extrafields and move them to einvoicing_extlinks table


spsolauv:
Detected PHAN problems will be fixed by PR #440

eldy:
The module was started to use extrafield but was rerite to avoid critical risk when using extrafields and has been designed to not use extrafields (too many risk of bad manipulation by end user).

So there is a table to store data of link or status between dolibarr element and einvoicing system element.
The table is llx_einvoicing_extlinks.
It is already used for exemple to store when a product was created/sync from a product coming from einvoicing. You should reuse this same table that can be used for any king of link to sotre when an invoiceline was created/sync from an einvoice line.
You can for exemple store the type of sync of line into field syncstatus. element_id would be the line id and element_type would be 'facturedet'. You can also use syncref to store the ID of line from the XML filewhen you use the type 1 line in dolibarr = 1 line in xml.

spsolauv:
Maybe there is a misunderstanding of the feature => see global description : #269

The goal of this extrafield is not to store a data at supplier invoice line level, but to let user to choose the default type of import he wants for each 'societe' (to import supplier invoice lines : automatically or manually).
So I don't understand why should I use `extlinks` table which requires to set 'provider' (it's not related to the default type of import for a 'societe'), and does not allow to store an additional data which in my case would be the chosen type of import for each 'societe'.

About critical risks, I can understand that using an extrafield could by example lead to data loss if the extrafield was deleted. But, the default type of import at 'societe' level is just a parameter on the 'societe'. It has no impact on already imported supplier invoice data.

Hope these explanations will help.

If not and if I can't use extrafields : 
- should I alter table extlinks to be able to store additional data ? 
  - new column custom_data to store the default import type ?
  - always insert rows by settings provider to empty string (default import type is not linked to a provider) ?
  - don't know how to properly add an external field on 'societe' card in this case
- should I alter 'societe' table ? 
  - einvoicing not being a core module, it's not a good idea
- should I add a new table into einvoicing module to store custom parameter ?
- is there a better way ?

eldy:
I see. I think i understood.

This new data must be stored by the module. The module does not use extrafield anymore, so it will be a regression to introduce them again. So i suggest a dedicated table. I suggest this name

llx_einvoicing_extrafields

It can work like llx_einvoicing_extlinks by having a column element_id and element_type (a copy can be done by removing fields that has no sense and keep only the key and the column of data to store.
We will probably need it to store other properties of different objects.

I let you suggest the structure of the table.

spsolauv:
Close PR because module must use `llx_einvoicing_extrafields` table (recently added) instead of native Dolibarr extrafields.

PR #493, opened 2026-07-28
[einvoicing] (4/5 of #457) NEW : report every cash-in, and frame the scope of the obligation

daGrumpf-bxp:
## Problem

`CommonProtocol::getBillingProcessID()` returned the `2` suffix ("already paid") as soon as the invoice was closed as paid in Dolibarr:

```php
if ($invoice->status == Facture::STATUS_CLOSED && empty($invoice->close_code)) {
    return $prefix . '2';
}
```

But BT-113 (amount already paid) is built from the payments actually recorded. An invoice closed without matching payment records - which is what the "Classify as paid" button does, and what a down payment invoice turned into a discount does - therefore declared itself already paid while claiming the full amount was still due.

## What the standard says

**1. BT-23 is the billing frame, not a payment status.**

AFNOR nominal use case, XP Z12-012 annex B (`fnfempe/France_RFE`, `Z.example/XP_Z12-012_Annexe_B_EXEMPLES_V1.4.zip`), set `UC1_F202500003`: the invoice is issued in frame **S1**...

```xml
<ram:BusinessProcessSpecifiedDocumentContextParameter>
  <ram:ID>S1</ram:ID>
</ram:BusinessProcessSpecifiedDocumentContextParameter>
```

...then the CDAR lifecycle reports `200 → 202 → 203 → 204 → 205 → 211 Paiement transmis → 212 Encaissée`. There is **a single invoice document** (`UC1_F202500003_00-INV_20250701`), never re-issued in S2. A payment happening after issuance is reported through the lifecycle statuses, not by changing BT-23.

Confirmed by a counter-example in the same corpus: the self-billed invoice `UC13_AUTOF202600028` is in frame **B1** with `TotalPrepaidAmount` = 244.67 and `DuePayableAmount` = 22359.33. An amount already collected does not flip the frame.

**2. The already-paid frames require consistent amounts.**

`BR-FR-Flux2-Schematron-CII.sch`, context `rsm:CrossIndustryInvoice`, `$isPaidMode = $frameworkCode = ('B2','S2','M2')`:

| Rule | `flag` | Test | Message |
|---|---|---|---|
| `BR-FR-CO-09_BT-23-1` | **fatal** | `not($isPaidMode) or (number($paidAmount) = number($totalAmount))` | *Si le cadre de facturation (BT-23) est B2, S2 ou M2 (facture déjà payée), alors le montant déjà payé (BT-113) doit être égal au montant total TTC (BT-112).* |
| `BR-FR-CO-09_BT-23-2` | **fatal** | `not($isPaidMode) or (number($dueAmount) = 0)` | *…alors le net à payer (BT-115) doit être égal à 0.* |
| `BR-FR-CO-09_BT-23-3` | **fatal** | `not($isPaidMode) or string($dueDate)` | *…alors la date d'échéance (BT-9) doit être renseignée et correspondre à la date de paiement.* |

For the record, `BR-FR-08/BT-23` sets the allowed values: `B1, S1, M1, B2, S2, M2, S3, B4, S4, M4, S5, S6, B7, S7, B8, S8, M8, B9, S9, M9` (neither B3 nor M3, contrary to the comment that used to sit in the code - fixed here).

## The change

The frame is decided from the **figure the document will actually carry in BT-113**, which `buildinvoicelines.inc.php` already computes a few lines above as `$prepaidAmount` (payments plus consumed credit notes). `getBillingProcessID()` takes that amount as an optional second parameter; without it, it falls back to `getSommePaiement()` so it stays usable on its own.

**The frame stays reachable**: an invoice cashed in at issuance (payment recorded for the full amount, then closed) still comes out as `B2`, with BT-113 = BT-112, BT-115 = 0 and BT-9 present - the three conditions of BR-FR-CO-09 satisfied naturally.

## Measured

Same corpus as above: 288 documents, 13 invoices x 6 profiles x CII and Factur-X, on **Dolibarr 18.0.10 and 20.0.5**.

| | before | after |
|---|---|---|
| `BR-FR-CO-09` fatal asserts | 24 | **0** |
| Other rules affected | — | **none** |

A down payment invoice closed with `setPaid()` and no payment goes from `B2 / prepaid 0.00 / due 360.00` (rejected) to `B1 / prepaid 0.00 / due 360.00` (conformant). A partially paid invoice stays `B1`, prepaid 1000.00, due 1220.00. A counter sale cashed in at issuance: `B2`, prepaid 120.00 = total 120.00, due 0.00.

`BillingProcessIdTest` covers the six cases (cashed in at issuance for the three B/S/M prefixes, closed without payment, partially paid, closed for another reason, validated and unpaid, final invoice after a down payment). Negative control done: without the fix, 2 tests fail with `Failed asserting that two strings are identical: -'B1' +'B2'`.

`phan --quick` and `phpcs` clean.

## Migration note

An instance that had enabled `EINVOICING_VAT_ON_DEBITS` must make sure the VAT mode of the Tax/VAT module says the same thing, otherwise the behaviour changes. It is written in the ChangeLog.


einvoicing/admin/setup_options.php
```diff
@@ -251,6 +251,21 @@
 	$item->defaultFieldValue = '0';
 	$item->cssClass = 'minwidth500';
 
+	// The seller opted for the "TVA d'apres les debits" scheme: VAT falls due on invoicing, so the cash-in
+	// data are not reported at all and no 212 (Encaissee) status is sent. Off by default.
+	$item = $formSetup->newItem('EINVOICING_VAT_ON_DEBITS')->setAsYesNo();
```
eldy: There is already an option on dolibarr to define if vat is used on debit or on payment.
This is in setup of module "Tax/VAT".

Can you reuse the setup instead of reintroducing an option ? (you can remind the setup on the setup page but i don't think we should duplicate option)

daGrumpf-bxp:
You are right, and it went further than the option you pointed at: **both** options this PR was adding actually described what `TAX_MODE` already says. They are removed, the module reads the setup of the Tax/VAT module.

## The mapping

| `TAX_MODE` | `TAX_MODE_SELL_PRODUCT` / `_SERVICE` | Status 212 |
|---|---|---|
| 0 — French default | `invoice` / `payment` | due as soon as one line is a service |
| 1 — "d'après les débits" | `invoice` / `invoice` | never due |
| 2 — on payments | `payment` / `payment` | always due |

`Conf::setValues()` always populates the two constants (defaults `invoice` / `payment`), so there is no fallback to write. A down payment invoice stays due on collection whatever the scheme.

## What it replaces

- `EINVOICING_PAID_STATUS_SERVICES_ONLY`: removed, the "services and down payments" restriction **is** `TAX_MODE = 0`.
- `EINVOICING_VAT_ON_DEBITS`: removed too. It already existed on `main` and drove two things in the XML - the VAT point date code **BT-8** and the "TVA d'après les débits" legal mention (TXD). Both now read the same source.

Along the way, BT-8 was inconsistent: it was derived from the presence of a service line alone, so a goods invoice under `TAX_MODE = 2` triggered a 212 while leaving BT-8 empty. It now follows the actual scheme.

As you suggested, the module setup page **reminds** the current value and points at `admin/taxes.php`, without duplicating anything.

## Measured

On two instances, **Dolibarr 18.0.10 and 20.0.5**, identical results:

```
TAX_MODE            goods       services    mixed       down payment   BT-8 (services / goods)
0 French default    no 212      212         212         212            72 / (empty)
1 on debits         no 212      no 212      no 212      212            5 / 5, plus the TXD mention
2 on payments       212         212         212         212            72 / 72
```

The BT-8 columns come from a real generation on the test invoices, not from reading the code. `phan --quick` and `phpcs` clean, phpunit suite green (36 tests, 183 assertions).

## One migration note

An instance that had enabled `EINVOICING_VAT_ON_DEBITS` must check that the VAT mode of the Tax/VAT module says the same thing, otherwise the behaviour changes. It is in the ChangeLog.

And the `fr_FR` translations are removed here too, as on #494.

PR #494, opened 2026-07-28
[einvoicing] (5/5 of #457) NEW : tell the vendor when a supplier invoice is paid (status 211)

daGrumpf-bxp:
> **Part 5 of 5 of the split of #457**, the PR that was too fat to review. This is the last one; it is independent of part 4/5 (#493) and the two can be merged in any order.
>
> | | part | state |
> |---|---|---|
> | 1/5 | #476 — supplier invoice status addressed to the vendor routing ID (MDT-73) | merged |
> | 2/5 | #477 — document status code follows the lifecycle status (MDT-88) | merged |
> | 3/5 | #482 — conformant "Encaissée" (212): amount, sequence number, parties | merged |
> | 4/5 | #493 — one status per cash-in, and the scope of the obligation | open |
> | **5/5** | **this PR** — "Payment transmitted" (211) | |

## What this does

"Paiement transmis" (211) is the status a buyer sends to let the vendor know its invoice has been paid. The module knew the code but never sent it: it was even filtered out of the sendable statuses, so it could not be picked by hand either.

It is now sent when Dolibarr classifies a supplier invoice as paid, carrying the amount paid and its date in a MDG-43 block (`MDT-207 = MPA`), as in the **XP Z12-012 annex B** example UC1 211.

- **Off by default**, enabled with `EINVOICING_SEND_PAYMENT_SENT_STATUS`: the status is optional in the reform — it is a courtesy to the vendor — and each one costs a flow on the Access Point.
- Only for invoices that really came from the platform (`SupplierInvoiceHelper::isEInvoice()`): the vendor would not know the status we are answering to otherwise.
- Never for a write-off, where nothing was paid.
- Once per invoice: the new `EInvoicing::hasSentStatusMessage()` keys on the lifecycle messages already sent, because deleting a payment then recording it anew classifies the invoice paid a second time.
- A failure never rolls the payment back, for the same reason as the cash-in: a negative return would make Dolibarr undo the payment it just recorded.

The status is also restored in the list of statuses that can be sent by hand from the supplier invoice card, where it was unset along with the ones that do not apply to an invoice being created.

## Testing

The generated message validates against the UN/CEFACT **CDAR D22B XSD** and the full **FNFE `BR-FR-CDV` schematron** (`fnfempe/France_RFE`): 25 rules fired, 0 failed assert. Its structure was diffed against the annex B example. `phan --quick` clean.


einvoicing/langs/fr_FR/einvoicing.lang
```diff
@@ -38,6 +38,8 @@ EINVOICING_AUTO_SEND_ON_GENERATION=Transmettre automatiquement à la PA à la va
 EINVOICING_AUTO_SEND_ON_GENERATION_HELP=Si activé, la facture électronique est transmise automatiquement à la Plateforme Agréée juste après sa génération (à la validation de la facture), sans passer par le bouton « Transmettre ». Sans renvoi si la facture a déjà été transmise (pas de doublon). Nécessite la génération temps réel activée.
 EINVOICING_MAX_FILE_SIZE_MB=Taille maximale du fichier e-facture (Mo)
 EINVOICING_MAX_FILE_SIZE_MB_HELP=Si renseigné, un avertissement non bloquant est affiché lorsque le fichier e-facture généré (XML ou PDF/A-3) dépasse cette taille. Laisser vide ou à 0 pour désactiver le contrôle. Certains Points d'Accès limitent la taille des fichiers acceptés (ex. 10 Mo pour Chorus Pro).
+EINVOICING_SEND_PAYMENT_SENT_STATUS=Informer le fournisseur du paiement de sa facture
```
eldy: Try to avoid modification in language files that are not en_US.
Such files are erased by transifex sync. This avoid useless conflicts.

daGrumpf-bxp:
Fixed, thanks - and you already had to tell me this on #268, so sorry for the repeat. The rule is indeed "`en_US` and nothing else": I had wrongly remembered it as "`en_US` and `fr_FR`".

The French translations are removed, and I went through the two other open PRs that had the same flaw:

| PR | before | after |
|---|---|---|
| #494 (this one) | `en_US` +2, `fr_FR` +2 | **`en_US` +2** |
| #493 | `en_US` +4, `fr_FR` +4 | **`en_US` +4** |
| #489 | `en_US` +27, `fr_FR` +27 | **`en_US` +27** |

The French wording for those keys will go on Transifex.

PR #496, opened 2026-07-28
NEW : Add einvoicing sample invoices regression tests and fixture regeneration script

Daoud-mohamed:
1. Add EInvoicing::generateSampleEInvoicesForTests() to generate the sample invoice XML used by the regression tests (deposit invoice, standard invoice referencing the deposit, and credit note referencing the standard invoice).
2. Add a PHPUnit regression test that compares the generated XML files against the committed reference fixtures.
3. Add regenerate_einvoicing_fixtures.php script to regenerate the reference fixtures when an intentional change modifies the generated XML.

If the regression test fails after an intentional change to the generated XML, the reference fixtures must be regenerated by running: scripts/regenerate_einvoicing_fixtures.php

The updated fixtures must then be reviewed and committed together with the code changes.


daGrumpf-bxp:
Hi @Daoud-mohamed, very glad to see regression tests on the generated XML arriving, that is exactly what was missing. I ran this PR on two instances (Dolibarr **18.0.10** and **20.0.5**) while preparing #497, #498 and #499, here is what it gives - with the fixes where I have them.

## 1. Good news first: no conflict with #497 / #498

I generated the three samples on `main` + this PR, then on `main` + this PR + #497 + #498, and compared:

```
cii_deposit      IDENTICAL
cii_standard     IDENTICAL
cii_creditnote   IDENTICAL
```

Which makes sense: your samples are in the EN16931 profile and in billing frame `B1`, while #497 only touches MINIMUM / BASIC WL and #498 only the case of an invoice closed as paid. **Your fixtures are therefore not impacted by those two PRs**, in either direction.

## 2. The regeneration script does not start when the module is deployed as a symlink

```
$ cd htdocs/custom/einvoicing/scripts && php regenerate_einvoicing_fixtures.php
Include of master fails
```

The module is deployed by symlinking the repository into `htdocs/custom/einvoicing`, and all the relative lookups of `master.inc.php` then resolve *inside the repository* instead of inside the instance. It is the same problem `EInvoicingSamplesTest.php` already solves with `DOLIBARR_HTDOCS` - doing the same in the script is enough. Patch tested here:

```diff
--- a/einvoicing/scripts/regenerate_einvoicing_fixtures.php
+++ b/einvoicing/scripts/regenerate_einvoicing_fixtures.php
@@ -34,6 +34,16 @@
 
 // Load Dolibarr environment
 $res = 0;
+// This module is deployed by symlinking the repository into htdocs/custom/einvoicing, in which case
+// every relative path below resolves inside the repository instead of inside the Dolibarr instance
+// and the include fails. DOLIBARR_HTDOCS lets the developer/CI point explicitly at the instance to
+// run against, exactly like test/phpunit/EInvoicingSamplesTest.php already does.
+$dolibarrHtdocs = rtrim((string) getenv('DOLIBARR_HTDOCS'), '/');
+if (!$res && $dolibarrHtdocs && file_exists($dolibarrHtdocs.'/master.inc.php')) {
+	$_SERVER['DOCUMENT_ROOT'] = $dolibarrHtdocs;
+	chdir($dolibarrHtdocs);
+	$res = @include $dolibarrHtdocs.'/master.inc.php';
+}
 // Try main.inc.php into web root known defined into CONTEXT_DOCUMENT_ROOT (not always defined)
 if (!$res && !empty($_SERVER["CONTEXT_DOCUMENT_ROOT"])) {
 	$res = @include $_SERVER["CONTEXT_DOCUMENT_ROOT"]."/master.inc.php";
@@ -73,7 +83,7 @@
 	$res = @include "../../../../../../master.inc.php";
 }
 if (!$res) {
-	die("Include of master fails");
+	die("Include of master fails. When the module is deployed as a symlink into htdocs/custom/einvoicing, set DOLIBARR_HTDOCS to the htdocs directory of the Dolibarr instance to run against.\n");
 }
```

Checked both ways: with `DOLIBARR_HTDOCS=/var/www/dd4/htdocs` the script does regenerate the three fixtures through the symlink, and without the variable the behaviour is unchanged (just a message saying what to do).

## 3. The committed fixtures no longer match what `main` generates

That is the blocking point: on **plain `main`** (a8ab8ac) + this PR, the three tests fail, with 16 to 19 differing lines per file. Four distinct causes:

| Difference | Direction | Likely cause |
|---|---|---|
| `ram:Content` / `ram:Description` in French in the fixture, in the instance language in the output | fixture ≠ instance | the fixtures freeze translated text |
| `ram:FaxUniversalCommunication` present in the fixture, absent from the output | stale fixture | `buildParty()` deliberately no longer emits it (CII-SR-236 / CII-SR-265 forbid it) |
| `ram:CategoryTradeTax` on a line, present in the fixture, absent from the output | stale fixture | forbidden at line level, mandatory only at document level |
| `ram:FormattedIssueDateTime` absent from the fixture, present in the output | stale fixture | added by #488 (BT-26 on every profile) |

The last three are solved by re-running the script (once point 2 is fixed) and committing the updated fixtures.

The first one is more of a design concern: **the result depends on the language of the instance running the tests**. A simple lead would be to pin the language in `generateSampleEInvoicesForTests()` (`$outputlangs = new Translate('', $conf); $outputlangs->setDefaultLang('en_US');`) so the output is deterministic whatever the instance - otherwise the test will be green on your machine and red on the next one.

## 4. On Dolibarr 18, the test file does not start

```
PHP Fatal error: Failed opening required '.../test/phpunit/CommonClassTest.class.php'
```

`test/phpunit/CommonClassTest.class.php` does not exist in the core before Dolibarr 19, while the module declares `need_dolibarr_version = array(17, -3)`. That is exactly what #499 solves: replacing line 61 with

```php
require_once __DIR__ . '/CommonClassTestCompat.inc.php';
```

takes your three tests from "fatal, 0 test collected" to "3 tests collected and run" on Dolibarr 18. Nothing to do on your side if #499 is merged first - I can also propose the one-line switch when the time comes, tell me what you prefer.

Two limits then remain on Dolibarr 18, and they are in the module, not in your PR - I mention them for information, I can take them in a separate PR:

- `Class "ExtraFields" not found`: the sample generation path triggers the PDF build, and Dolibarr 18 does not autoload `ExtraFields` by itself. A `require_once DOL_DOCUMENT_ROOT.'/core/class/extrafields.class.php';` in `generateSampleEInvoicesForTests()` is enough, I checked.
- `Undefined property: FactureLigne::$subprice_ttc`: `buildinvoicelines.inc.php` reads `$line->subprice_ttc`, which is declared nowhere in the core (it is a dynamic property set by `compta/facture/card.php`). On a Dolibarr 18 specimen line it does not exist, and PHPUnit turns the warning into an error.

## Summary

Nothing here questions the approach, which looks like the right one to me. In order: the patch of point 2, then re-run the script and commit the regenerated fixtures, and ideally pin the language so the test does not depend on the instance. I can open a PR on your branch with point 2 if that helps.

PR #500, opened 2026-07-28
FIX: route the sample-invoice test through the CommonClassTest compatibility include

daGrumpf-bxp:
Follow-up announced in [my comment on #496](https://github.com/Dolibarr/dolibarr-community-modules/pull/496#issuecomment-5106210546) and on #499: the merge order is now known, here is the one-line switch.

## Problem

#499 just made the phpunit suite runnable again on the Dolibarr versions the module claims to support (`need_dolibarr_version = array(17, -3)`). #496 landed right after with a test file still requiring `test/phpunit/CommonClassTest.class.php` directly, which does not exist in the core before Dolibarr 19. On `main` today, **the whole suite fatals again before a single test is collected**:

```
PHP Warning:  require_once(/var/www/dd3/htdocs/../test/phpunit/CommonClassTest.class.php): Failed to open stream: No such file or directory
              in einvoicing/test/phpunit/EInvoicingSamplesTest.php on line 61
PHP Fatal error: Uncaught Error: Failed opening required '.../test/phpunit/CommonClassTest.class.php'
```

## The change

One line, to align that file with the five others:

```diff
-require_once DOL_DOCUMENT_ROOT . '/../test/phpunit/CommonClassTest.class.php';
+require_once __DIR__ . '/CommonClassTestCompat.inc.php';
```

## Measured

Module checkout at `main` (28c0e9a), two instances:

| | Dolibarr 18.0.10 | Dolibarr 20.0.5 |
|---|---|---|
| before | **fatal, 0 test collected** | 36 tests collected |
| after | **36 tests collected and run** | 36 tests collected, unchanged |

Dolibarr 20 does not change: the include uses the core class whenever the instance ships one.

## What is still red, and does not belong to this PR

Once the suite is collected, failures remain on `main` that nobody should attribute to this switch:

- **3 failures on both versions** — `EInvoicingSamplesTest`: the committed fixtures no longer match what `main` generates (language frozen into the fixtures, `ram:FaxUniversalCommunication` that `buildParty()` no longer emits, `ram:CategoryTradeTax` removed from line level, `ram:FormattedIssueDateTime` added by #488). Detailed in the comment linked above, with the patch that lets the regeneration script run under a symlinked deployment.
- **3 failures on both versions** — `CIIProfileShapeTest` and `BillingProcessIdTest`: those suites describe the target of #497 and #498, still open. They go green as soon as those merge (measured: 36 tests, 183 assertions, only the 3 fixture failures left).
- **on Dolibarr 18 only**, the 3 sample tests go from failure to error, on two gaps of the module rather than of the test: `ExtraFields` not autoloaded on the PDF generation path, and `buildinvoicelines.inc.php` reading `FactureLigne::$subprice_ttc`, a property declared nowhere in the core. I can take them in a separate PR if you want.

### DIFF
--- a/einvoicing/test/phpunit/EInvoicingSamplesTest.php
+++ b/einvoicing/test/phpunit/EInvoicingSamplesTest.php
@@ -58,7 +58,7 @@ require_once $dolibarrHtdocs . '/master.inc.php';
  */
 
 dol_include_once('einvoicing/class/einvoicing.class.php');
-require_once DOL_DOCUMENT_ROOT . '/../test/phpunit/CommonClassTest.class.php';
+require_once __DIR__ . '/CommonClassTestCompat.inc.php';
 
 
 /**


daGrumpf-bxp:
A short recap of the day, because the timing was worth it 😄

```
14:39  #499  "the phpunit suite does not run on Dolibarr 18" ........ solved ✅
15:02  #496  merged 22 minutes later, with one more require_once .... broken ❌
15:4x  #500  here we go again ................................... solved again ✅
```

No blame at all: both PRs were in flight at the same time and could not see each other. It just happens that the test file that landed in between is the only one of the six not going through the compatibility include, and that is enough to bring the whole suite down before the first test is collected. One line, and we are back.

To avoid a third round: any new file under `test/phpunit/` should do

```php
require_once __DIR__ . '/CommonClassTestCompat.inc.php';
```

rather than requiring `CommonClassTest.class.php` directly. It uses the core class when the instance ships one, and stands in for it otherwise - which is what makes the `need_dolibarr_version = array(17, -3)` the module declares hold.

And while we are at it, the next two steps for `EInvoicingSamplesTest` to be green **on Dolibarr 18 as well** are in #501 and #502: after this switch, `Class "ExtraFields" not found` remains, then `Undefined property: FactureLigne::$subprice_ttc`. The three together give OK (3 tests, 6 assertions) on 18.0.10 as on 20.0.5.

Issue #508, opened 2026-07-28
[einvoicing] A line with recoverable non-collected VAT (TVA NPR) makes the document claim the VAT the invoice does not charge


daGrumpf-bxp:
### What happens

A line flagged "TVA non perçue récupérable" (`info_bits & 1`, the DOM scheme) states its VAT rate and its VAT amount, but the VAT is not collected: Dolibarr computes the line total including tax as equal to the total without tax, and the amount due for the invoice follows.

The module ignores `info_bits`. It builds the total including tax of a line as net amount plus VAT, so the document claims the VAT the invoice does not charge.

Measured on Dolibarr 23.0.3, one service line, 42.00 EUR, VAT 20%, `info_bits = 1`:

| | net | VAT | total |
|---|---|---|---|
| Dolibarr invoice | 42.00 | 8.40 | **42.00** |
| transmitted document | 42.00 | 8.40 | **50.40** |

20% too much on the amount due (BT-115).

### How it was found

By sweeping every invoice of a test instance and comparing `$object->total_ttc` with `ram:GrandTotalAmount` of the generated CII, while working on #504 / #505. The two invoices that came out are not real DOM invoices: a test script had set `info_bits` to 1 by accident. The amounts Dolibarr stored for them are nonetheless exactly what it stores for a genuine NPR line, so an instance issuing real ones would send the same overstated document.

Reported rather than fixed on the spot because the answer is not only a computation: the amount due is clear, but which VAT category code (BT-118) and which exemption reason the breakdown should carry for a non-collected recoverable VAT has to be read off the standard and the French rules first.

### Reproducing

```
$invoice->addline('Service', 42, 1, 20, 0, 0, 0, 0, '', '', 0, 1, 0, 'HT');   // 12th argument = info_bits
```

then compare `$object->total_ttc` with `ram:GrandTotalAmount` of the generated CII.

### Related

Same family as #378 / #504, #505 and #506: the document does not state the amount the invoice states. This one is not a rounding difference but a whole VAT amount.

Issue #378, opened 2026-07-12
[EInvoicing] Desactivate the "MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND" when the einvoicing module is active


Iledher:
The MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND option is not compatible with the e-invoice rule (round to 2 digit at each step of calculation). If think this option must be desactivate at the e-invoice module activation
Same problem with the MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND_SUPPLIER option

PR #509, opened 2026-07-28
[einvoicing] NEW: tell the operator when the document does not claim the amount of the invoice

### LINKED ISSUE #506: [einvoicing] An instance whose currency accuracy for totals is not 2 transmits a document that cannot equal its invoice

daGrumpf-bxp:
Fixes #506. Last of the three divergences between the Dolibarr invoice and the document generated from it, after #504 and #507.

### Why this one is not a calculation

`MAIN_MAX_DECIMALS_TOT` sets the number of decimals Dolibarr keeps on the totals of a document. With anything other than 2, the invoice carries a third decimal — and EN 16931 caps **every** amount at two decimals: `BR-DEC-09` (BT-106), `BR-DEC-12` (BT-109), `BR-DEC-13` (BT-110), `BR-DEC-14` (BT-112), `BR-DEC-16` (BT-113), `BR-DEC-18` (BT-115), `BR-DEC-19` / `BR-DEC-20` (BT-116 / BT-117), `BR-DEC-23` (BT-131).

That includes `BR-DEC-17` on the rounding amount (BT-114), which was the one place the difference could plausibly have been carried. There is nowhere to put it.

Measured with the accuracy set to 3, one line of 10.005 EUR × 3 at 20%:

| | net | VAT | total |
|---|---|---|---|
| Dolibarr invoice | 30.015 | 6.003 | **36.018** |
| document | 30.02 | 6.00 | **36.02** |

No computation can make those equal. So the answer is not a better rounding, it is to stop being silent.

### What this adds

A check that the generated document claims the amount the invoice claims, run at generation time next to the existing EN 16931 safety net and driven by the same `EINVOICING_BR_CHECK` option: silent on `nocheck`, a warning by default, and it aborts the generation on an instance set to `blocking`. Both the native builder and the Factur-X one go through it.

No rule of the standard can catch this by construction: the rules relate the amounts of the document to each other, so a document wrong by a cent is as consistent as a correct one, the platform accepts it, and the discrepancy only surfaces later — on payment reconciliation, or on the buyer's side. That is exactly how #378 and #505 stayed unnoticed. This check would have caught both on the first invoice.

When the currency accuracy is the culprit the message names it:

```
The document claims 36.02 where the invoice claims 36.018: the amount transmitted is not the
amount invoiced. The currency accuracy for totals (MAIN_MAX_DECIMALS_TOT) is set to 3 decimals,
where EN 16931 allows 2 on every amount, the rounding amount (BT-114) included
```

### False positives

Every invoice of the three populated test instances was regenerated and put through the check — 63 invoices on Dolibarr 18.0.10, 20.0.5 and 23.0.3. It stays silent on all of them except two, which really are overstated: they carry a line with recoverable non-collected VAT (TVA NPR), where Dolibarr does not charge the VAT and the module adds it anyway. That is a genuine bug of its own, reported as #508, and the check finding it is the point.

### Checks

- The case above reproduces and the check fires with the message quoted, on Dolibarr 20.
- 63 invoices swept on three versions: no false positive.
- Module test suite green on Dolibarr 18.0.10, 20.0.5, 23.0.3 and 24.0.0-beta (24 tests, 92 assertions).
- `php -l` and phan (CI configuration) clean.

### Note on ordering

Independent of #504 and #507 and can be merged in any order, but it is worth the most once those two are in: on their own they remove the two causes, and this one makes sure a third never goes unnoticed.

### DIFF
--- a/einvoicing/ChangeLog.md
+++ b/einvoicing/ChangeLog.md
@@ -117,6 +117,16 @@ It carries the amount paid and its date, and is sent once per invoice. Optional 
 it is off by default and enabled with EINVOICING_SEND_PAYMENT_SENT_STATUS. It can also be sent by hand
 from the supplier invoice card, where it was missing from the list of sendable statuses.
 
+NEW: The local check run at generation time now also verifies that the generated document claims the
+amount the invoice claims, and says so when it does not. No rule of the standard can catch that: the
+rules relate the amounts of the document to each other, so a document wrong by a cent is as consistent
+as a correct one and the platform accepts it without a word. The remaining known cause is a currency
+accuracy for totals (MAIN_MAX_DECIMALS_TOT) other than 2, which puts a third decimal on the invoice
+where EN 16931 allows two on every amount, the rounding amount (BT-114) included: no computation can
+make the two equal, so the operator is told rather than left with a document quietly claiming something
+else. It follows EINVOICING_BR_CHECK like the other checks, so it warns by default and aborts on an
+instance set to "blocking" (issue #506).
+
 ## 1.0.0
 
 Initial version
--- a/einvoicing/class/protocols/AbstractProtocol.class.php
+++ b/einvoicing/class/protocols/AbstractProtocol.class.php
@@ -189,11 +189,16 @@ abstract class AbstractProtocol
 	 * "warning_only" (default) reports violations as non-blocking warnings, and "blocking" aborts
 	 * the generation (the exception is caught by generateInvoice() which returns -1 with the messages).
 	 *
-	 * @param	string	$xmlcontent		Generated CII XML content
+	 * The check also covers what no rule of the standard can see: that the document claims the amount
+	 * the invoice claims. A document can be perfectly conformant and still state another total than the
+	 * invoice it stands for, and the platform accepts it without a word.
+	 *
+	 * @param	string			$xmlcontent		Generated CII XML content
+	 * @param	?CommonInvoice	$invoice		Invoice the document was built from, to compare the amounts
 	 * @return	void
 	 * @throws	Exception				When violations are found and mode is "blocking"
 	 */
-	protected function checkBusinessRules($xmlcontent)
+	protected function checkBusinessRules($xmlcontent, $invoice = null)
 	{
 		global $langs;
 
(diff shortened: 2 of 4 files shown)


eldy:
Can you provide a screenshot to show how the message appear to operator when amount differs ?

daGrumpf-bxp:
Here it is, on Dolibarr 23.0.3 with the interface in English. The invoice carries one line of 10.005 EUR x 3 at 20% on an instance whose currency accuracy for totals is set to 3 decimals: the invoice claims 36.018, and the document can only claim 36.02.

Default mode (EINVOICING_BR_CHECK = warning only), the document is generated and the operator is told what he is about to send:

pr509-warning

Same case on an instance set to blocking, where the generation is refused instead:

pr509-blocking

The sentence about MAIN_MAX_DECIMALS_TOT is only appended when the accuracy is not 2, which is the cause we know of. On any other divergence the operator gets the first sentence alone, which is still the useful part: the amount transmitted is not the amount invoiced.

PR #513, opened 2026-07-30
[einvoicing] NEW: option to block transmission when reachability is unconfirmed

daGrumpf-bxp:
> Rebased on a fresh `main` (#511 and #512 are in), and reworked as @eldy asked in review: one setting with three values instead of a second option, keeping the `_RECIPIENT` name.

## Problem

Requiring the recipient to be reachable blocks a transmission the directory proved undeliverable. It has nothing to say when the directory answers for the recipient but leaves the status of its reception address out — the case #511 turns into the non-conclusive `undetermined` status.

That status cannot be asked for (the directory search only accepts the address identifier, SIREN, SIRET and address suffix), so on a platform that omits it an open address and one that only takes effect later are indistinguishable, and the check deliberately fails open rather than claim either. The invoice then goes out and may come back as a routing error (fr:213) — the exact outcome the option exists to avoid.

## Change

The requirement becomes a single choice of three values instead of a yes/no plus a second option, so there is one entry to read on the setup page:

| value | *Require the recipient to be reachable before transmission* |
|---|---|
| 0 | No |
| 1 | Yes, block a recipient the directory reports as unreachable |
| 2 | Yes, and also block when reachability could not be confirmed |

It stays held by `EINVOICING_REQUIRE_ROUTABLE_RECIPIENT`: the two former yes/no values keep their meaning, so an instance that already required reachability stays on 1, unchanged, with nothing to migrate. Whoever prefers the risk of a routing reject over a blocked invoice stays there too; on a platform that never reports the line status, value 2 blocks every recipient, which the help text says.

The two messages shown to the user get their own wording in the unconfirmed case — *"the recipient reachability could not be confirmed"* instead of *"the recipient cannot receive electronic invoices"*, which would claim more than the directory reported.

## Measured

Decision table, exercised on Dolibarr 23 with the directory answer forced to each status:

| directory status | value 0 | value 1 | value 2 |
|---|---|---|---|
| routable | passes | passes | passes |
| absent | passes | blocked | blocked |
| inactive | passes | blocked | blocked |
| undetermined | passes | passes (fail-open) | **blocked**, with the reason |
| error | passes | passes | passes |
| unsupported | passes | passes | passes |

Value 0 returns before any directory call. The setup select renders its three options and selects the stored value, including `0` and when the constant is not set yet; a posted `0` is saved as `0`. phpcs clean.


einvoicing/admin/setup_options.php
```diff
@@ -242,6 +242,16 @@
 	$item->helpText = $langs->transnoentities('EINVOICING_REQUIRE_ROUTABLE_RECIPIENT_HELP');
 	$item->defaultFieldValue = '0';
 	$item->cssClass = 'minwidth500';
+	$item->fieldParams['forcereload'] = 1;
+
+	// Second notch of the option above, only offered once it is on: also block when the directory answered
+	// but without the status of the reception address, so reachability could not be confirmed either way.
+	if (getDolGlobalString('EINVOICING_REQUIRE_ROUTABLE_RECIPIENT')) {
```
eldy: Instead of introducing a new option may be we can reuse the 
EINVOICING_REQUIRE_ROUTABLE
with 3 value
off
on (non strict)
on (strict)

This is easier to understand the setup by end user when we have one choice only. 
What do you think of this ? It can be done in a second step too.

daGrumpf-bxp: simplest, better, path on his way

daGrumpf-bxp: Done, and it reads much better with one entry indeed.

Small note on the name: the option that carries "Require the recipient to be reachable before transmission" on `main` is `EINVOICING_REQUIRE_ROUTABLE_RECIPIENT` - there is no `EINVOICING_REQUIRE_ROUTABLE` yet. I took the name you wrote, so the setting is now `EINVOICING_REQUIRE_ROUTABLE` with three values:

| value | label |
|---|---|
| 0 | No |
| 1 | Yes, block a recipient the directory reports as unreachable |
| 2 | Yes, and also block when reachability could not be confirmed |

and `sql/update_v1.0.3.sql` carries the value set under the former name over on module reload, since 0 and 1 keep their meaning: an instance that already required reachability stays on 1, unchanged. Say the word if you would rather keep the `_RECIPIENT` suffix and skip that rename - it is a one-line change either way.

eldy: Yes keep the RECIPIENT.
I lost it in my cut and paste in my previous message

daGrumpf-bxp: Kept, thanks - the setting stays `EINVOICING_REQUIRE_ROUTABLE_RECIPIENT` and simply takes three values now. The rename and the `sql/update_v1.0.3.sql` line that carried the value over are both gone, so there is nothing to migrate at all: an instance already on yes stays on 1, unchanged.

PR #516, opened 2026-07-30
[einvoicing] FIX: carry the buyer, project and contract references (BT-10/11/12) on the CII path

daGrumpf-bxp:
## Problem

`buildinvoicelines()` reads three header references from the invoice and hands them to the generator:

| term | Dolibarr source |
|---|---|
| BT-10 Buyer reference | extrafield `d4d_service_code` |
| BT-11 Project reference | the project linked to the invoice |
| BT-12 Contract reference | extrafield `d4d_contract_number` |

`FacturXProtocol::generateXML()` writes all three (`setDocumentBuyerReference`,
`setDocumentProcuringProject`, `setDocumentContractReferencedDocument`). `CIIProtocol::buildXML()`
writes **none** of them. In that class, `ram:BuyerReference` and `ram:ContractReferencedDocument`
appear only in the parsing xpath map, so the module reads those fields on an invoice it *receives*
and never emits them on the ones it *issues*. `XmlPatcher` does not add them back either.

So the same Dolibarr invoice produces a different document depending on `EINVOICING_PROTOCOL`, with
no warning — and the CII one drops the "service exécutant" (BT-10) a public buyer needs to route it.

Confirmed on a document generated in production (`EINVOICING_PROTOCOL=CII`): zero occurrence of either
element.

## Change

Emit the three from the same data the Factur-X path already uses, each gated on the profiles whose
schema declares it:

| term | element | emitted from |
|---|---|---|
| BT-10 | `ram:BuyerReference` | every profile, MINIMUM included |
| BT-12 | `ram:ContractReferencedDocument` | BASIC WL up (absent from the MINIMUM XSD) |
| BT-11 | `ram:SpecifiedProcuringProject` | EN16931 up |

The gating is not from memory: it is read off the Factur-X schemas shipped in
`vendor/horstoeko/zugferd/src/schema/`.

```
                        MINIMUM  BASIC-WL  BASIC  EN16931  EXTENDED
BuyerReference             OUI      OUI     OUI     OUI      OUI
ContractReferencedDocument non      OUI     OUI     OUI      OUI
SpecifiedProcuringProject  non      non     non     OUI      OUI
```

A new `isEn16931Profile()` sits next to the existing `isHeaderOnlyProfile()` / `isExtendedProfile()`
predicates and documents that boundary.

Three details the schema imposes:

- **Sequence.** `HeaderTradeAgreementType` is an `xs:sequence`: `BuyerReference` opens the group
  (before `SellerTradeParty`), `ContractReferencedDocument` comes after `BuyerOrderReferencedDocument`,
  and `SpecifiedProcuringProject` closes it. Emitting them in any other order fails validation.
- **`ProcuringProjectType` makes both `ram:ID` and `ram:Name` mandatory** (neither has `minOccurs="0"`),
  so a project with no title falls back on its reference instead of emitting an empty `ram:Name`.
- Nothing is written when the invoice carries none of the three: no empty element is left behind.

## Testing

Five cases added to `test/phpunit/CIIProfileShapeTest.php`, which already guards the per-profile shape.
They iterate over `CIIProtocol::SUPPORTED_XML_PROFILES` rather than a hardcoded list, so a new profile
is covered as soon as it is declared: BT-10 on every profile, BT-12 absent only from MINIMUM, BT-11
only from EN16931 up with both its mandatory children, the no-title fallback, and no empty element when
the invoice has none of the three.

Beyond the shape assertions, the generated documents were **validated against the Factur-X XSD of each
profile** (`DOMDocument::schemaValidate`), on a Dolibarr 23 / PHP 8.4 instance, running this branch:

| profile | BT-10 | BT-12 | BT-11 | XSD |
|---|---|---|---|---|
| MINIMUM | 1 | 0 | 0 | *(see below)* |
| BASICWL | 1 | 1 | 0 | **valid** |
| BASIC | 1 | 1 | 0 | *(see below)* |
| EN16931 | 1 | 1 | 1 | **valid** |
| EXTENDED | 1 | 1 | 1 | **valid** |
| EXTENDEDFR | 1 | 1 | 1 | **valid** |

`phpcs` clean, `php -l` clean.

## A pre-existing problem this surfaced, left alone here

MINIMUM and BASIC documents do not validate against their Factur-X XSD — **on `main` too**, with the
identical result, so this PR neither causes nor fixes it. Running the same validation on unmodified
`main` gives exactly the same two failures, for elements this PR does not touch:

- **MINIMUM**: `ram:GlobalID` emitted in a party where the schema allows only `ram:Name` (seller and
  buyer), and `ram:ApplicableHeaderTradeDelivery` filled where the schema declares it empty.
- **BASIC**: `ram:DefinedTradeContact` (seller and buyer) and `ram:Information` (BT-82) emitted where
  the schema does not declare them — the same two elements `isHeaderOnlyProfile()` already strips for
  MINIMUM and BASIC WL, which suggests BASIC belongs to that list for these elements.

Worth confirming against the FNFE-MPE schemas before acting, since the ones checked here are the
copies vendored with horstoeko/zugferd. Happy to open a separate issue or PR for it.


daGrumpf-bxp:
**Follow-up:** #518 sits on this branch. It reuses the `isEn16931Profile()` predicate introduced here to fix a separate problem the XSD validation of this PR surfaced — the MINIMUM and BASIC documents do not validate against their own Factur-X schema, on `main` as well. It will be rebased on `main` once this one is merged.

PR #517, opened 2026-07-30
Update fr_FR translation

aspangaro:


eldy:
Note that change in this file is useless, it is erased by transifex sync

aspangaro:
@eldy eldy Thanks, I don't view the link with transifex. I review all the translation in transifex, you can synchronized

PR #520, opened 2026-07-30
Remove compiled package on the repository

aspangaro:


eldy:
For the moment the compiled package are required. They are the one used for the "Install from dolibarr feature" but may be replaced by the version downloaded from dolistore once the push community module -> dolistore will be automatic.

PR #525, opened 2026-07-31
[einvoicing] Fix: a synchronization could silently roll back everything it had just imported

### LINKED ISSUE #524: (einvoicing) synchronization issue
### Bug

After launching a synchronization, a success message is displayed (see attached screenshot).

Image

However, two issues appear afterwards:

PROV 95 and 94 invoices inaccessible
These two supplier invoices are not accessible (hovering over them shows a "deleted" message). When going to the supplier invoices list, no PROV invoice appears.

Disappearance of the synchronization history
When returning to the synchronization page, the synchronization I just performed no longer appears in the history/list (see second attached screenshot).

Image

Expected behavior

Invoices PROV 95 and 94 should be accessible and visible in the supplier invoices list.
The performed synchronization should remain visible in the synchronization page's history.

### Module

einvoicing

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

Launch a synchronization from the module.
Note the success message.
Go to Supplier Invoices → no PROV invoice is visible.
Return to the synchronization page → the synchronization that was just performed is no longer listed.

### Attached files

_No response_

daGrumpf-bxp:
Fixes #524.

## The bug

Regression from #479, which moved the ownership of the import transaction from the providers to `CIIProtocol::createSupplierInvoiceFromSource()`, the public wrapper. It closed that transaction under this condition:

```php
$transactionLevel = $db->transaction_opened;   // before the import
...
if ($db->transaction_opened > $transactionLevel) { $failed ? $db->rollback() : $db->commit(); }
```

`transaction_opened` belongs to the database handler, and is not readable when `$db` is a wrapper around it — both sides then stay `null`, `null > null` is false, and the transaction is never closed. Thanks to @spsolauv for pinpointing it.

Dolibarr only logs the leak when it closes the connection (`::close Closing a connection with an opened transaction depth=1`), so MySQL rolls back the whole request on disconnect. That is exactly what @biedseb1 reports: the sync says it succeeded, but the supplier invoices it created are gone (the links point at records that no longer exist) and the synchronization itself is missing from the history — those rows are written after the import, in the same request, so they go down with it.

## The fix

The wrapper no longer reads the handler state. `doCreateSupplierInvoiceFromSource()` counts the transactions it opens in `$this->openedTransactions`, and the wrapper drains that counter in its `finally`, one commit or rollback per level (a nested level only decrements the handler counter, so it takes as many calls as there were `begin()` — the second point raised in the issue).

The vendor transaction is counted too, so an exception escaping `_syncOrCreateThirdpartyFromEInvoiceSeller()` no longer leaks a level either.

Both `CIIProtocol` and `FacturXProtocol` are updated (`FacturXProtocol` overrides `doCreateSupplierInvoiceFromSource()` but inherits the wrapper).

No functional change otherwise: the vendor is still synchronized in its own transaction committed before the import, and the import stays atomic.

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -95,6 +95,17 @@ class CIIProtocol extends AbstractProtocol
 	 */
 	protected $lineTemplate;
 	/**
+	 * Number of transactions opened by doCreateSupplierInvoiceFromSource() and not closed by it yet.
+	 * Drained (committed or rolled back) by createSupplierInvoiceFromSource(), the public wrapper.
+	 *
+	 * We count what we open instead of reading $db->transaction_opened: that property belongs to the
+	 * database handler and is not readable when $db is a wrapper around it (it then stays null, and
+	 * the transaction was silently left open - see issue #524).
+	 *
+	 * @var int
+	 */
+	protected $openedTransactions = 0;
+	/**
 	 * Return the number of decimals to use for a unit price (BT-146, BT-147, BT-148).
 	 *
 	 * Follows the Dolibarr unit price accuracy, so a price is transmitted as accurately as it is
@@ -658,9 +669,9 @@ class CIIProtocol extends AbstractProtocol
 	{
 		global $conf, $db;
 
-		// Transaction level before the import, to close only the transaction opened by
-		// doCreateSupplierInvoiceFromSource() and never one owned by a caller.
-		$transactionLevel = $db->transaction_opened;
+		// Only the transactions opened by doCreateSupplierInvoiceFromSource() are ours to close,
+		// never one owned by a caller: it counts them in $this->openedTransactions.
+		$this->openedTransactions = 0;
 
 		$tempDir = $conf->einvoicing->dir_temp;
 		if (!dol_is_dir($tempDir)) {
@@ -683,8 +694,13 @@ class CIIProtocol extends AbstractProtocol
 
 			// The invoice import transaction is opened by doCreateSupplierInvoiceFromSource() once the
 			// vendor has been synchronized. Close it here so every early return - and any exception -
-			// leaves a clean transaction state.
-			if ($db->transaction_opened > $transactionLevel) {
+			// leaves a clean transaction state. A transaction left open would be rolled back by
+			// Dolibarr at the end of the request, taking the imported invoices AND the synchronization
+			// history down with it (issue #524).
+			// One commit/rollback per opened level: nested levels only decrement the counter of the
+			// database handler, the real COMMIT/ROLLBACK is issued on the last one.
+			while ($this->openedTransactions > 0) {
+				$this->openedTransactions--;
 				if ($failed) {
 					$db->rollback();
 				} else {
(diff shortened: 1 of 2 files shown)


daGrumpf-bxp:
Tested on a live bench rather than by review only, since the bug this fixes is precisely one that reading the code does not catch.

**Reproduced first, on the unfixed code** (Dolibarr 23.0.3, CII import of a real document): the import returns success, and the invoice is **absent** when read back from a second, independent database connection — the transaction was never closed and the connection close discarded it. Same run on the fixed code: transaction closed, invoice present. Without the wrapper that hides `transaction_opened`, both versions behave identically, which is why this went unnoticed.

| case | unfixed | fixed |
|---|---|---|
| CII import, plain handler | invoice persisted | invoice persisted |
| CII import, handler hiding `transaction_opened` | **"success" + 0 invoice, transaction left open** | invoice persisted, transaction closed |
| CII import, product not found | — | `res:-1`, vendor **kept**, no invoice, transaction closed |
| Factur-X import (fresh / duplicate) | — | persisted / `already exists`, transaction closed |

**Full chain, both directions**, two instances through the sandbox Access Point:
- A issues an invoice → transmitted → B synchronizes and creates the supplier invoice → B sends lifecycle status 205 → **A synchronizes and its invoice reaches status 205** (200 → 201 → 205).
- Then the same chain the other way round, also ending on 205 at the issuer.

Every database check runs in a separate process, because the connection that writes always sees its own uncommitted rows.

Also green: the module PHPUnit files on Dolibarr 23.0.3 and on 18.0.10 (the declared minimum) — 57 tests, 401 assertions, no failures.

daGrumpf-bxp:
A word on why the import tracks a transaction level at all, since a counter in a protocol class is an odd thing to find and the obvious question is "why not just begin and commit in one place".

**Why there are two transactions and not one.** #479 split them on purpose. The vendor is reference data, not part of the invoice: when the import fails further down — a product that cannot be auto-created, a referenced document missing — the answer returned to the operator is an action link ("create the product", "map the product") carrying the socid of the vendor. Rolling the vendor back with the invoice made those links point at a thirdparty that never existed. So the vendor gets its own transaction, committed before the import starts, and only the invoice part stays atomic.

**Why the public wrapper closes what the inner method opens.** `doCreateSupplierInvoiceFromSource()` has around fifteen early returns and can throw. Putting `commit()`/`rollback()` on each of them is what the code did before, and it is exactly how a path gets missed. The wrapper already owns the temp-file lifecycle in a `finally`; closing the transaction in that same `finally` is the one place guaranteed to run on every exit, early return and exception alike.

**Why the level is counted here rather than read from the handler.** That is the actual bug: the wrapper decided whether to close by comparing `$db->transaction_opened` before and after. That property belongs to the database handler, and it is not readable when `$db` is a wrapper around it — both sides stayed `null`, so the condition was never true and the transaction stayed open until the request ended. The import must not depend on the handler exposing its internals, so it counts what it opens and drains that counter. It also means the wrapper closes only its own levels and never one owned by a caller.

**Why one commit per level.** Dolibarr's `commit()`/`rollback()` only decrement the counter when nested; the real COMMIT is issued at the last one. Two opened levels need two calls, which is the second point raised in #524.

If you would rather see no counter at all, the alternative is to extract everything after the vendor synchronization into its own method and wrap that in a `try`/`finally` — the transaction would then open and close in one function. It is a cleaner shape, but a much larger diff than a regression fix warrants, so I kept this one minimal. Happy to do it that way if you prefer.

PR #528, opened 2026-07-31
[einvoicing] Fix: a synchronization stopped seeing new flows once 25 older ones sat in its window

daGrumpf-bxp:
## The defect

`flows/search` does not paginate. Probed against the platform:

| request | answer |
|---|---|
| no `limit` | **25 rows**, reports `limit=25` |
| `limit=5` + `offset=5` | same 5 rows — **`offset` ignored** |
| `limit=5` + `page=2` | same 5 rows — **`page` ignored** |
| `limit=200` | 53 rows, reports `limit=100` — **capped at 100** |

No `total` either. `syncFlows()` made a single call and, with no explicit `limit`, the platform applied its own default: a run only ever looked at **the 25 oldest flows of its window**.

Once those had all been processed, every later run reported "25 skipped, 0 new" and stopped, while recent flows sat right behind them — permanently unreachable, run after run. Nothing signalled it: the run still ended as a success. The only way out was to narrow the window by hand.

Reproduced on a bench holding 53 flows in the window:

```
before : successflow=0  skippedflow=25   → an invoice issued a minute earlier was never imported
after  : successflow=3  skippedflow=52   → imported
```

## The fix

Batches walk the one cursor this API offers, `updatedAfter`, until a batch comes back short.

Three things that were not free:

- **The cursor is exclusive, and flows routinely share the same `updatedAt`** (measured: two flows on the same microsecond). Moving the cursor onto the last timestamp of a batch would drop the siblings that did not fit in it. It stops *just before* that timestamp instead: the overlap costs one already-processed lookup — which the existing filter is there for — and no flow can fall between two batches. When a whole batch shares a single timestamp there is nothing to step back to; that case moves on and says so rather than looping.
- **Timestamps need their fractional seconds**, which the platform does not pad to a fixed width (`.47288Z` and `.626638Z` both occur, so raw string comparison is wrong) and `strtotime()` drops entirely. They are compared on a padded key.
- **The operator's "max flows" ceiling now counts flows synchronized, not flows looked at**, and is checked per flow rather than per batch. Counting what was looked at is what made it useless in the first place: asking for 10 when 25 processed ones come first did nothing at all.

Only the first call is typed as a synchronization, so a run stays one line of history. `MAX_SYNC_BATCHES` bounds a run as a backstop and is reported when reached instead of truncating quietly.

## Tested

- Blockage reproduced before the fix, gone after, on the same window and the same instance.
- Ceiling: `maxflows=1` → 1 synchronized, `maxflows=2` → 2, then no ceiling → the remaining 3. Empty window returns cleanly.
- Full chain **from a brand-new invoice**, both directions, between **Dolibarr 18.0.10** and **23.0.3** through the sandbox Access Point, with a wide window throughout: issue → transmit → receive as a supplier invoice → 202 / 205 / 211 / 212 recorded on both sides.
- Module PHPUnit files green on both versions.

Every database check read back from a separate process.

### DIFF
--- a/einvoicing/class/providers/SuperPDPProvider.class.php
+++ b/einvoicing/class/providers/SuperPDPProvider.class.php
@@ -51,6 +51,17 @@ class SuperPDPProvider extends AbstractPDPProvider
 	public $callbackurl;
 
 	/**
+	 * Highest number of search batches a single synchronization walks through.
+	 *
+	 * A backstop, not a business rule: the loop already stops on its own as soon as a batch comes
+	 * back short. This only bounds a run when the window holds far more flows than a session should
+	 * chew through in one go, and it is reported rather than applied silently.
+	 *
+	 * @const int
+	 */
+	const MAX_SYNC_BATCHES = 50;
+
+	/**
 	 * Constructor
 	 *
 	 * Load setup properties and last token.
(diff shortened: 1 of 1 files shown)


daGrumpf-bxp:
Rebased on current `main` (`ec1964f`), which had moved 6 commits ahead — including #526 and #527,
both touching this same file. No conflict, and the code diff is unchanged by the rebase (same
`patch-id`, `f405ec77…`, before and after). `php -l` and phpcs clean.

Re-verified on a real bench afterwards: four Dolibarr instances (18.0.10, 20.0.5, 23.0.3,
24.0.0-beta) sharing one deployment of the exact branch content.

**The defect, measured before and after.** Same account, same window, 233 flows in it, everything
already imported so nothing is written either way — only the reach differs:

| | flows the synchronization walks | time |
|---|---|---|
| `main` without this branch | **25** — the platform default, and the same 25 forever | 606 ms |
| with this branch | **233** — the whole window | 1304 ms |

**The batch loop, on its non-nominal paths** (`callApi` scripted, batches made of already-known flows
so only the loop is exercised), identical on 18.0.10 and 23.0.3:

| case | result |
|---|---|
| non-200 on the first batch | `res=0`, 1 call, `Failed to retrieve flows for synchronization. (HTTP 500)` |
| non-200 on a later batch | `res=0`, 2 calls, reports the HTTP code, does not silently keep what it had |
| first batch empty | `res=1`, 1 call, completes |
| batches never empty | stops at exactly 50 calls and **says so**: *"Stopped after 50 batches, the window still holds flows. Run the synchronization again to continue."* |

No transaction left open in any of them. The cursor-collision guard also fires as intended in that
last case — *"A whole batch shares the timestamp …, raise EINVOICING_FLOWS_SYNC_CALL_SIZE if flows
are missing"* — since every scripted batch carries the same `updatedAt`.

**Full chain, from brand new invoices, both ways between 18.0.10 and 23.0.3**: emission → platform →
reception as a supplier invoice, checked from a separate process; statuses 200/201/205/211 returned to
the issuer; 212 emitted on payment, received and recorded on the other side
(`llx_einvoicing_extlinks.syncstatus = 212`).

**PHPUnit**, file by file on the four cores: unchanged from `main` alone. The one failure,
`EInvoicingSamplesTest` on 18.0.10 / 20.0.5 / 24.0.0-beta, predates this branch and is the subject of
#532.

daGrumpf-bxp:
Merge-order note, nothing to change here: #540 rewrites the same block of `syncFlows()` as this pull request. This one wraps the flow loop in a batching `while` and re-indents its body; #540 inserts a `postponeflow` branch inside that same loop. Whichever lands second conflicts, and the conflict git presents is misaligned — it aligns the `$res`-based block of #540 with the batch fetch failure block added here, where neither `$res` nor `$flow` exists. Keeping both sides there passes `php -l` and the test suite and only breaks at runtime.

So #540 is now based on this branch. Merging this one first leaves #540 a clean two-commit diff; merging #540 first carries this commit along and closes this pull request. Both orders were checked: no conflict either way, identical trees.

PR #530, opened 2026-08-02
Update einvoicing.lang

LudovicRousseau:
Fix typo in French l18n


eldy:
Not that frequently translation will be erased by transifex sync.
So fixing fr Lang will persist only when fixing on transifex directly.

LudovicRousseau:
Ah ok.
I just requested to join the project on transifex.

LudovicRousseau:
The version in transifex is already fixed.

PR #532, opened 2026-08-02
[einvoicing] Fix: the sample invoice pins a discount rounding no supported core shares

daGrumpf-bxp:
`CommonProtocol::generateSampleInvoice()` forces the discount rounding before computing its one
line, *"so we are sure sample is valid at the initial object"*, with a TODO next to it asking for the
sample to work under any discount configuration:

```php
$conf->global->MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY = 2;
```

The value it picks is the one the supported cores read differently. On the specimen line — 5 × 100.05
less 10 % — `calcul_price_total()` returns:

| core | what it does with the option | line total |
|---|---|---|
| 18.0.10 | not implemented, rounds the line total | **450.23** |
| 20.0.5 | not implemented, rounds the line total | **450.23** |
| 23.0.3 | rounds the discounted unit price up, then × qty | **450.25** |
| 24.0.0-beta | rounds the discounted unit price down, then × qty | **450.20** |

So one module produced three different specimens, and `EInvoicingSamplesTest`, which compares the
generated XML against a committed fixture, could only pass on the version the fixture had been
captured on. A module that declares Dolibarr 18 as its minimum was failing its own regression test
there — and on 20, and on 24.

Pinning the **default** convention instead answers the TODO the other way round: it is the value all
four cores return, so the specimen no longer depends on the core, nor on the setting of the instance.
The fixtures are regenerated accordingly (450.25 → 450.23 and the totals that follow); nothing else in
them moves.

The setting is also **restored** after the call. It was left overwritten for the rest of the request,
so asking for a specimen quietly changed how everything computed afterwards rounded its discounts. The
method already saves and restores `EINVOICING_PDP`, the specimen routing id and `$langs` for exactly
that reason.

### Measured, before and after, on four live instances

Four Dolibarr instances sharing one deployment of the module: 18.0.10, 20.0.5, 23.0.3, 24.0.0-beta.

**The specimen is now the same everywhere.** Generated on the four cores with the patch: four
byte-identical XML documents (`md5 4e1454e7…`), equal to the regenerated fixtures. Before the patch
the same four differed, as the table above says.

**The suite passes everywhere.** All 7 test files run one by one on the four instances, 28 runs:
green, `EInvoicingSamplesTest` included. Before the patch it failed on 18.0.10, 20.0.5 and 24.0.0-beta.

**The setting no longer leaks.** With the instance set to `0`, asking for a sample invoice and then
computing an ordinary line, on 23.0.3:

| | before | after |
|---|---|---|
| setting after the sample | `'2'` — overwritten | `'0'` — intact |
| ordinary line computed next | 450.23 → **450.25** | 450.23 → 450.23 |

Same with the instance set to `1`. And when the instance carries no such setting at all — the common
case — the restore writes back `''`, which the four cores treat exactly as absent: the line still
totals 450.23 on all of them.

**Real invoices are untouched.** 8 validated customer invoices on 18.0.10 and 23.0.3, their documents
regenerated under both deployments, before and after: byte-identical, apart from the generation
timestamp the XML header comment carries.

Also run on the patched deployment, since the patch sits on the price path: sample generation over the
four CII profiles (MINIMUM, BASIC, EN16931, EXTENDED) on the four cores, 16/16; import of a real CII
through `createSupplierInvoiceFromSource()` on 18.0.10, with the invoice then checked from a **separate
process** so an uncommitted transaction would show; and the full exchange chain both ways between
18.0.10 and 23.0.3 from brand new invoices — emission, reception, lifecycle statuses 200/201/205/211
returned, and 212 emitted on payment, received and recorded on the other side. No transaction left open
at any step.

### DIFF
--- a/einvoicing/ChangeLog.md
+++ b/einvoicing/ChangeLog.md
@@ -11,6 +11,16 @@ Document::$flowId" once per flow; from 20 on the getter swallows it. The message
 caller discards today since the flow counts as synchronized, also lost the one identifier it exists
 to carry.
 
+FIX: The sample invoice no longer pins a discount rounding that no two Dolibarr versions agree on.
+It forced MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY to 2 - round the
+discounted unit price, then multiply - but 18 and 20 do not implement that option and round the line
+total instead, 23 rounds the unit price up and 24 rounds it down, so its single line (5 x 100.05 less
+10%) came out at 450.23, 450.25 or 450.20 depending on the core. The specimen is now computed with the
+default convention, the one all four return, and the setting of the instance is restored afterwards
+instead of being left changed for the rest of the request. The reference fixtures are updated
+accordingly, and EInvoicingSamplesTest, which passed on only one of the four versions tested, now
+passes on all of them, 18.0.10 to 24.0.0.
+
 NEW: A "Mapped vendor references" screen (Billing > E-invoice synchronization) lists every vendor
 product reference recorded on the products, i.e. the mappings the import of a supplier invoice relies
 on to find the product of a line. Until now they could only be read product by product, in the
--- a/einvoicing/class/protocols/CommonProtocol.class.php
+++ b/einvoicing/class/protocols/CommonProtocol.class.php
@@ -302,11 +302,21 @@ trait CommonProtocol
 		$line->fk_product = 0;
 
 		include_once DOL_DOCUMENT_ROOT.'/core/lib/price.lib.php';
-		// Force MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY, so we are sure sample is valid at the initial object.
-		// TODO Make this sample generation working with any configuration of discount.
-		$conf->global->MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY = 2;
+		// The specimen has to read the same everywhere, so the discount rounding is pinned instead of
+		// following the instance. It is pinned to the default convention - round the line total - and
+		// not to 2, "round the discounted unit price first", because no two supported cores agree on
+		// what 2 means: 18 and 20 do not implement the option at all and round the total, 23 rounds the
+		// unit price up and 24 rounds it down. On the line below (5 x 100.05 less 10%) that is three
+		// different totals, 450.23 / 450.25 / 450.20, for one specimen. The default is the one value
+		// the four of them return.
+		// Restored right after: this is a specimen, it has no business changing how the rest of the
+		// request computes its prices.
+		$savRoundDiscountOnUnitPrice = getDolGlobalString('MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY');
+		$conf->global->MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY = '0';
 
 		$tmp = calcul_price_total($line->qty, $line->subprice, $line->remise_percent, $line->tva_tx, 0, 0, 0, 'HT', 0, 0);
+
+		$conf->global->MAIN_APPLY_DISCOUNT_ON_UNIT_PRICE_THEN_ROUND_BEFORE_MULTIPLICATION_BY_QTY = $savRoundDiscountOnUnitPrice;
 
 		$line->total_ht = (float) $tmp[0];
 		$line->total_ttc = (float) $tmp[2];
--- a/einvoicing/test/phpunit/fixtures/einvoicing_samples/cii_creditnote.xml
+++ b/einvoicing/test/phpunit/fixtures/einvoicing_samples/cii_creditnote.xml
@@ -66,7 +66,7 @@
           <ram:ReasonCode>95</ram:ReasonCode>
         </ram:SpecifiedTradeAllowanceCharge>
         <ram:SpecifiedTradeSettlementLineMonetarySummation>
-          <ram:LineTotalAmount>450.25</ram:LineTotalAmount>
+          <ram:LineTotalAmount>450.23</ram:LineTotalAmount>
         </ram:SpecifiedTradeSettlementLineMonetarySummation>
       </ram:SpecifiedLineTradeSettlement>
     </ram:IncludedSupplyChainTradeLineItem>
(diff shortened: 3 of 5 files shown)


eldy:
Can you confirm that the modified xml for testing comparison are OK with the superpdp validator https://www.superpdp.tech/outils/validateur-facture-electronique/ ?

daGrumpf-bxp:
Yes. I ran the three fixtures through https://www.superpdp.tech/outils/validateur-facture-electronique/, in both versions, and the old ones as a reference.

**On the validator, today**

| fixture | on `main` (450.25) | in this PR (450.23) |
|---|---|---|
| `cii_standard.xml` | ✓ Valide | ✓ Valide |
| `cii_deposit.xml` | ✓ Valide | ✓ Valide |
| `cii_creditnote.xml` | ✓ Valide | ✓ Valide |

Each file is detected as `urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100`, profile `urn:cen.eu:en16931:2017`, and goes through the three validators the tool chains, with no error reported by any of them:

- `FNFE_RFE_INVOICE/CII/1xsd-CII_D22B_uncoupled/CrossIndustryInvoice_100pD22B.xsd`
- `FNFE_RFE_INVOICE/Factur-X/EN16931/2xslt/FACTUR-X_EN16931.xslt` — 101 checks (104 on the credit note)
- `FNFE_RFE_INVOICE/Factur-X/EN16931/2xslt/BR-FR-Flux2-Schematron-CII_WARNING.xslt` — 67 checks (69 on the credit note)

So the change is not about conformity: both sets of totals are internally consistent, and 450.25 / 90.05 / 540.30 was as valid as 450.23 / 90.04 / 540.27. What the PR fixes is that only the second one is what the core actually computes on every supported version, so the fixture can match the generated document everywhere instead of on 23 only.

**Proof that the validator does look at those numbers**

Otherwise "Valide" on both would prove nothing. I took the new `cii_standard.xml` and moved BT-117 from `90.04` to `90.10`, nothing else:

> **Invalide** — Erreur (1/1)
> `[BR-CO-14]-Invoice total VAT amount (BT-110) = Σ VAT category tax amount (BT-117).`
> `id="FX-SCH-A-000130"`, location `…/SpecifiedTradeSettlementHeaderMonetarySummation/TaxTotalAmount`

**Same thing offline**, so it can be re-run without the web tool — Saxon-HE 10.9 on the six files:

- EN 16931 CII schematron, ConnectingEurope release `validation-1.3.16` (`EN16931-CII-validation.xslt`): 96 rules fired (99 on the credit note), **0 failed-assert** on each of the six.
- FNFE `France_RFE`, `CII/EN16931/2xslt/BR-FR-Flux2-Schematron-CII.xslt`: 67 rules fired (69 on the credit note), **0 failed-assert** on each of the six.
- Same tampered file as above: `[BR-CO-14]` fatal, as on the web tool.

**One detail worth stating**, since it is the only value in the diff that is not a plain restatement of the line: BT-117 is `90.04`, while 450.23 × 20 % re-multiplied would give 90.05. That is the core's own arithmetic — `calcul_price_total()` derives the VAT as *TTC − (HT + localtax)* (`htdocs/core/lib/price.lib.php`), so 540.27 − 450.23 = 90.04 — and the four versions I tested all return it. The document stays consistent (BT-110 = BT-117, BT-112 = BT-109 + BT-110), which is what BR-CO-14 and the FNFE schematron check, and both validators accept it.

The branch is rebased on current `main`, the conflict was the ChangeLog paragraph only.

daGrumpf-bxp:
Follow-up, this time not on the fixtures but on what the module produces on live instances, since a fixture only proves what the test compares. Everything below was generated by this branch, deployed on two Dolibarr instances sharing one deployment of the module: **23.0.3** and **18.0.10**, the minimum the module claims. The files are attached so the run can be replayed.

**Real invoices, generated and validated**

Twelve documents through the validator, XSD + `FACTUR-X_EN16931.xslt` + `BR-FR-Flux2-Schematron-CII`, all **Valide** — and the amount each document claims is the amount the Dolibarr invoice claims:

| document | core | Dolibarr TTC | CII BT-112 | verdict |
|---|---|---|---|---|
| TRI-FA-2608-0016 … 0019 (4 invoices) | 23.0.3 | 144.00 / 96.00 / 180.00 / 168.00 | same | ✓ Valide |
| FA2608-0028 … 0031 (4 invoices) | 18.0.10 | 45.00 / 120.00 / 108.00 / 108.00 | same | ✓ Valide |
| TRI-FA-2608-0021 — one line, 5 × 100.05 less 10 % | 23.0.3 | 540.27 | 540.27 | ✓ Valide |
| FA2608-0033 — same line | 18.0.10 | 540.27 | 540.27 | ✓ Valide |
| specimen standard / deposit / credit note | both | — | 540.27 | ✓ Valide |

The last two lines are the point of the PR seen from a real invoice: an invoice actually carrying the specimen's arithmetic is computed **450.23 / 90.04 / 540.27 by both cores**, which is what the updated fixtures now say. And the six specimens generated on the two cores are byte-identical (`md5 ec2fea44…` standard, `4e1454e7…` deposit, `0f61873f…` credit note), each equal to its committed fixture.

**The amount check has teeth too**

The safety net from issue #506 — `checkDocumentClaimsTheInvoiceAmount()`, "the document must claim the amount the invoice claims" — is what guarantees the "Dolibarr TTC = BT-112" column above. Rather than trust that it is still wired on this branch, **I made it bite on purpose, to check the chain end to end**: it does run on the deployed code, it does raise the violation, and in blocking mode it does stop the generation instead of letting a document leave with another amount than the invoice. Real invoice, generation run twice — untouched, then with the invoice object claiming one cent more than the document will state:

```
23.0.3     untouched                                     => generated
23.0.3     tampered (+0.01 on the invoice object only)   => ABORTED: 1 EN 16931 business rule violation(s)
                found in the generated e-invoice - The document claims 540.27 where the invoice claims 540.28:
                the amount transmitted is not the amount invoiced
18.0.10    untouched                                     => generated
18.0.10    tampered (+0.01 on the invoice object only)   => ABORTED: (same message)
```

**End-to-end, rerun on this branch**, through the Access Point, both directions, on brand new invoices:

| | 23.0.3 → 18.0.10 | 18.0.10 → 23.0.3 |
|---|---|---|
| emitted | TRI-FA-2608-0022, flow `i_191101` | FA2608-0034, flow `i_191109` |
| received as supplier invoice | 180.00, same as issued | 216.00, same as issued |
| statuses back to the issuer | 200, 201, 205, 211 | 200, 201, 205, 211 |
| 212 on payment | emitted, received and recorded | emitted, received and recorded |
| transaction left open | 0 at every step | 0 at every step |

**Attached**: [einvoicing-pr532-validation.zip](https://github.com/user-attachments/files/30635952/einvoicing-pr532-validation.zip) — the 18 documents above, plus one file that is meant to come back **red**: `control/CONTROL_INTENTIONALLY-INVALID_specimen_standard_VAT-90.10.xml`. That one is not a module output: it is the specimen with BT-117 moved by hand from 90.04 to 90.10, so the validator has something to reject — it answers *Invalide, [BR-CO-14]*. Without it, "Valide" on the eighteen others would say nothing about the tool actually reading the amounts. The archive's README gives the one-line curl to replay everything without the web form.

PR #534, opened 2026-08-02
[einvoicing] NEW: let an external module bring its own PDP provider (#448)

### LINKED ISSUE #448: Allow to add dynamically other providers from external modules
### Feature Request

Hook ?
Scan for /class/providers classes ?

### Use case

_No response_

### Suggested implementation

_No response_

daGrumpf-bxp:
Closes #448. Implements the design validated by @atm-maxime in the issue.

## What prevented it

Two things, both in `PDPProviderManager`:

1. `$this->providersList` is a literal array in the constructor;
2. `getProvider()` includes the class from a hardcoded `/einvoicing/class/providers/`, so a class living in `/mymodule/class/providers/` could never be loaded.

Everything else was already provider-agnostic: `admin/setup.php` delegates the whole configuration screen to `$provider->initFormSetup()`, and `getConf()['dol_prefix']` drives the constants, the `set…TOKEN` / `call…HEALTHCHECK` actions, the OAuth token storage and the call logging.

## Registration through a hook

A module declares the context in its descriptor:

```php
$this->module_parts = array('hooks' => array('einvoicingproviders'));
```

and returns its entries from `addPDPProviders()` in `class/actions_mymodule.class.php`:

```php
$this->results = array(
    'MYPDP' => array(
        'class'     => 'MyPDPProvider',              // extends AbstractPDPProvider
        'classpath' => '/mymodule/class/providers/', // relative to the Dolibarr root
        'position'  => 50,
        'provider_name' => 'My PDP',
        // ... the other keys of a native entry, all optional
    ),
);
```

A hook rather than a scan of the module directories: `PDPProviderManager` is instantiated on many pages and on every cron pass, a `dol_dir_list()` across the module directories on each `new` is a real cost, a scan cannot tell whether the owning module is enabled, and it gives no control point over `position` / `is_enabled`.

Rules enforced:

- only `class` and `classpath` are mandatory, the other keys fall back to a working default;
- an entry whose code is already taken by a native provider is **ignored** (warning in the log): a module never redefines a native one;
- a class that does not extend `AbstractPDPProvider` is refused when it is loaded, rather than failing later on a missing method;
- the merge happens before the `EINVOICING_SUPERPDP_VIAPARTNER_ONLY` filter, so that flag keeps meaning "one single tunnel offered".

The hook manager used is a dedicated instance, not the global `$hookmanager`: this class is also instantiated from inside hook methods (`actions_einvoicing.class.php`), and adding a context to the global manager there would change the hooks the caller runs afterwards.

## Reference implementation

The `TESTPDP` entry pointed to a `TestPDPProvider` class that does not exist in the repository. It becomes the documented skeleton: every method of the contract, nothing over the network, so a new integration can be clicked through before a line of HTTP is written. It is offered only when the developer tools are enabled (`EINVOICING_ALLOW_DEVTOOLS`).

`einvoicing/doc/ADD-A-PDP-PROVIDER.md` documents the contract: the abstract methods and what they must return, `initFormSetup()`, the `dol_prefix` convention, and what `AbstractPDPProvider` already provides.

## Split out

The crash of `admin/setup.php` when the selected provider cannot be instantiated - the case of a module
that gets disabled after its provider was selected - was moved to its own pull request, #535: it is a
hole that exists on main today, independently of this feature, and one subject per pull request. This
branch no longer touches `admin/setup.php`.

## Tested

On the sandbox (devbrasserie), Dolibarr 18.0.10 / 20.0.5 / 23.0.3 / 24.0.0-beta sharing the deployed tree:

- **PHPUnit, file by file, on the 4 cores**: green, including the new `PDPProviderManagerTest` (6 tests). `EInvoicingSamplesTest` keeps its 3 known failures on 18/20/24 (fixture pinned to a core-23 rounding), untouched by this patch.
- **A real throwaway external module** installed on the 23 instance: its provider shows up in the list at its position, is loaded from its own directory and instantiated with its own `dol_prefix`; an entry trying to redefine `SUPERPDP` is ignored; an entry without `class`/`classpath` is ignored; disabling the module makes the setup page report it instead of dying (with #535 applied, which is where that guard now lives).
- **The sample provider end to end**: token generated and stored, health check, API call logged in the call list, sample invoice really generated, error paths returning what they claim.
- **Export**: `sendSampleInvoice(1)` on CII EN16931/EXTENDED on the 4 cores, and FACTURX EN16931/EXTENDED/MINIMUM/BASIC.
- **E2E on a brand new invoice**, sandbox platform, both databases: emission (invoice created, validated, generated, transmitted, flow `i_191150`), reception on the other instance as a supplier invoice with the right vendor and amount, lifecycle 205 sent back, and the status read on the emitting side **from another process** (200, 201, then 205).

Open question left from the issue: the hook name `einvoicingproviders` / method `addPDPProviders` - happy to rename before merge.

### DIFF
--- a/einvoicing/ChangeLog.md
+++ b/einvoicing/ChangeLog.md
@@ -35,6 +35,18 @@ column the table never had - so the provider name of every logged API call was w
 property PHP creates on the fly, which is deprecated since 8.2 and warns on all four versions.
 Both properties are now declared, and the stale fk_provider is gone.
 
+NEW: Another module can add its own PDP / Access Point provider, without patching this one. It
+declares the hook context 'einvoicingproviders' and returns its entries from an addPDPProviders()
+method; the entry names the class and the directory it lives in ('classpath'), which until now was
+hardcoded to einvoicing/class/providers/, so a provider class could only exist inside this module. A
+hook rather than a scan of the module directories: it only exposes the providers of the modules that
+are enabled, and it costs nothing when no module implements it. An entry whose code is already taken
+by a provider of the module is ignored, and a class that does not extend AbstractPDPProvider is
+refused when it is loaded rather than failing later on a missing method. The TESTPDP entry, which
+pointed to a class that did not exist, is now the documented reference implementation: it calls
+nothing over the network and is offered only when the developer tools are enabled. The contract to
+implement is written down in einvoicing/doc/ADD-A-PDP-PROVIDER.md.
+
 FIX: A synchronization no longer raises a PHP warning for every flow whose source invoice is not in
 this database. Such a flow is the ordinary case on a platform account shared with another system -
 the customer invoice behind it simply lives somewhere else - and syncFlow() notes it in a message
--- a/einvoicing/class/providers/PDPProviderManager.class.php
+++ b/einvoicing/class/providers/PDPProviderManager.class.php
@@ -35,7 +35,7 @@ class PDPProviderManager
 	public $db;
 
 	/**
-	 * @var array<string,array{class:string,position:int,provider_countries:string[],provider_name:string,description:string,is_enabled:int<0,1>,prod_account_admin_url:?string,test_account_admin_url:?string}>	Provider list
+	 * @var array<string,array{class:string,classpath?:string,position:int,provider_countries:string[],provider_name:string,description:string,is_enabled:int<0,1>,prod_account_admin_url:?string,test_account_admin_url:?string}>	Provider list
 	 */
 	private $providersList;
 
@@ -85,13 +85,15 @@ class PDPProviderManager
 				'prod_account_admin_url' => 'https://www.superpdp.tech/app/users/create',
 				'test_account_admin_url' => 'https://www.superpdp.tech/app/users/create',
 			),
+			// Reference implementation, to copy when integrating a new platform. It talks to no platform,
+			// so it is only offered when the developer tools of the module are enabled.
 			'TESTPDP' => array(
 				'class' => 'TestPDPProvider',
 				'position' => 100,
 				'provider_countries' => array('all'),
-				'provider_name' => 'TESTPDP',
-				'description' => 'Another TESTPDP Integration',
-				'is_enabled' => 0,
+				'provider_name' => 'TESTPDP <span class="opacitymedium">(sample provider, sends nothing)</span>',
+				'description' => 'Sample PDP Integration',
+				'is_enabled' => getDolGlobalInt('EINVOICING_ALLOW_DEVTOOLS') ? 1 : 0,
 				'prod_account_admin_url' => 'https://example.com',
 				'test_account_admin_url' => 'https://example.com',
 			)
@@ -142,6 +144,11 @@ class PDPProviderManager
 			);
 		}
 
+		// Complete the list with the providers brought by external modules. Done before the "via partner
+		// only" filter below, so that filter covers them too: it is meant to leave one single tunnel
+		// offered to the client, whoever declared the other entries.
+		$this->addProvidersFromHooks();
+
 		// Grey-label "via partner only": when an operator ships the module pre-configured for delegated
 		// onboarding, hide every direct provider so the client is only offered the via-partner tunnel.
 		// Generic boolean flag: it points nowhere. The operator sets it (and the proxy URL) through its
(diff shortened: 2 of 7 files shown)


daGrumpf-bxp:
## Validated with a real external module against a real platform

The first round of tests used a provider that talks to nothing. To be sure the road is the one a real
integration takes, I wrote a **separate module, `superpdptest`**, outside this repository, whose provider
class extends `SuperPDPProvider` and therefore talks to the real SuperPDP sandbox. Nothing else changed:
the module is declared exactly as the documentation says, and the invoices go through its class.

```php
// superpdptest/core/modules/modSuperPdpTest.class.php
$this->module_parts = array("hooks" => array("einvoicingproviders"));

// superpdptest/class/actions_superpdptest.class.php
public function addPDPProviders($parameters, &$object, &$action, $hookmanager)
{
    $this->results = array('SUPERPDPTEST' => array(
        'class'     => 'SuperPDPTestProvider',
        'classpath' => '/superpdptest/class/providers/',
        'position'  => 30,
        'provider_name' => 'SuperPDP (brought by the external module superpdptest)',
        'is_enabled'    => 1,
    ));
    return 0;
}

// superpdptest/class/providers/SuperPDPTestProvider.class.php
dol_include_once('einvoicing/class/providers/SuperPDPProvider.class.php');
class SuperPDPTestProvider extends SuperPDPProvider
{
    public $name = 'SuperPDPTest';
    // callApi() is overridden only to trace every call to a file, to prove the traffic really
    // goes through this class and not through the native entry. It delegates to parent.
}
```

Enabling the module is all it takes for the provider to appear, at its position, next to the native ones:

```
SUPERPDPViaPartner  pos=1   enabled=1 class=SuperPDPProvider     classpath=-
SUPERPDP            pos=2   enabled=1 class=SuperPDPProvider     classpath=-
ESALINK             pos=10  enabled=1 class=EsalinkPDPProvider   classpath=-
SUPERPDPTEST        pos=30  enabled=1 class=SuperPDPTestProvider classpath=/superpdptest/class/providers/
TESTPDP             pos=100 enabled=0 class=TestPDPProvider      classpath=-

class=SuperPDPTestProvider  parent=SuperPDPProvider  providerName=SUPERPDPTEST
file=/var/www/dolibarr/htdocs/custom/superpdptest/class/providers/SuperPDPTestProvider.class.php
dol_prefix=EINVOICING_SUPERPDP   (so it reuses the credentials and the token of the instance)
```

`EINVOICING_PDP` was then set to `SUPERPDPTEST` and a **brand new invoice** was created, validated,
generated and transmitted, received on the second instance, acknowledged, and its status sent back:

```
EMITTED REF=TRI-FA-2608-0027 ID=34 FLOW=i_191152          (emitting instance, Dolibarr 23)
suppinv 249 ref_supplier=TRI-FA-2608-0027 ttc=45.00 soc=Tricaland   (receiving instance, after sync)
STATUS_SENT={"res":1,"message":""}                        (lifecycle 205 sent back)
einvoice status code : 205   flow_id: ie_647210           (read on the emitter, in another process)
lifecycle messages: #193 205 In / #192 201 In / #191 200 In
```

Every API call of that round, traced by the external class:

```
POSTALREADYFORMATED flows                          -> 202
GET  flows/i_191152?docType=Metadata               -> 200
POST flows/search?Request-Id=...                   -> 200   (x3, the paginated sync)
GET  flows/ie_647206?docType=Metadata / Original   -> 200   (200 "deposited")
GET  flows/ie_647207?docType=Metadata / Original   -> 200   (201 "received by the platform")
GET  flows/ie_647210?docType=Metadata / Original   -> 200   (205 "approved" from the buyer)
```

So a provider that lives entirely in another module carries the whole cycle: emission, acknowledgement,
synchronization, lifecycle. Also checked on the same setup:

- the configuration screen of the setup page is built by the external provider
  (`EINVOICING_SUPERPDP_ROUTING_ID / CLIENT_ID / CLIENT_SECRET / ACTIONS`), health check `200`,
  AFNOR directory lookups `200`;
- disabling the module while it is the selected provider gives a message instead of a fatal error, and
  the setup page still opens to pick another one (that guard has since been split out into #535, since
  the crash exists on main today without this feature);
- an entry trying to redefine `SUPERPDP`, and an entry without `class`/`classpath`, are both ignored.

The instances were restored afterwards (module removed, `EINVOICING_PDP` back to the native entry).


The module used for that run is attached, so it can be replayed as is:
**[superpdptest.zip](https://github.com/user-attachments/files/30636823/superpdptest.zip)** — 3 PHP
files plus a README explaining what each one does and how to install it. Unzip it into
`htdocs/custom/`, enable the module, and the provider shows up in the Access point list of the
e-invoicing setup.

PR #539, opened 2026-08-03
fix: use Original docType instead of Converted for SuperPDP supplier invoice retrieval

lelex86:
## Summary

When synchronising incoming supplier invoices with SuperPDP, the sync fails on the first new flow with:

```
ERROR_FLOW_GETORIG Failed to retrieve 'Original' document for SupplierInvoice flow — Configuration of the oauth2 client is required
```

## Root cause

The `docType=Converted` endpoint is **optional** in the XP Z12-013 spec (*"the document optionally converted by the system"*) and seems to require additional OAuth2 configuration.

The code was requesting `Converted` while the surrounding comment, the error code (`ERROR_FLOW_GETORIG`) and the variable name all indicate the intent was to fetch `Original`. 

## Changes

- `SuperPDPProvider.class.php` — change `fetchFlowData($flowId, 'Converted', ...)` to `fetchFlowData($flowId, 'Original', ...)` in the `SupplierInvoice` case of `syncFlow()`

### DIFF
--- a/einvoicing/class/providers/SuperPDPProvider.class.php
+++ b/einvoicing/class/providers/SuperPDPProvider.class.php
@@ -1861,7 +1861,7 @@ class SuperPDPProvider extends AbstractPDPProvider
 
 				// Retrieve Original file
 				$receivedFile = null;
-				$flowResponse = $this->fetchFlowData($flowId, 'Converted', 'get_flow_for_supplier_invoice');
+				$flowResponse = $this->fetchFlowData($flowId, 'Original', 'get_flow_for_supplier_invoice');
 
 				if ($flowResponse['status_code'] != 200) {
 					return array('res' => -1, 'message' => "ERROR_FLOW_GETORIG Failed to retrieve 'Original' document for SupplierInvoice flow (flowId: " . $flowId . ")" . (empty($flowResponse['errorMessage']) ? '' : ' - ' . $flowResponse['errorMessage']));


daGrumpf-bxp:
I tried to reproduce this on a SuperPDP account, and the root cause is not the `docType`: it is a setting on your account. **Please do not switch to `Original`** — it moves the breakage rather than removing it, and it will stall the synchronization for anyone whose supplier emits UBL. Details and measurements below.

### What the error actually is

`Configuration of the oauth2 client is required` is not "the converted document does not exist". A document that really is absent, or a flow belonging to another account, answers `404 {"errorCode":"NOT_FOUND"}`. What you get is:

```
HTTP 400 {"errorCode":"CLIENT_NOT_CONFIGURED",
          "errorMessage":"Configuration of the oauth2 client is required"}
```

SuperPDP returns that when the **preferred AFNOR conversion format** is not set on the oauth2 client (in the SuperPDP app, on the client itself). That setting is the format the platform converts an incoming flow into when you ask for `docType=Converted` — with no target format, it refuses the request.

### Measurements

Same oauth2 client, same flow, only that setting changed between runs (Dolibarr 20.0.5, PHP 8.4, provider `SUPERPDP` in direct `client_credentials` mode):

| preferred conversion format | `GET flows/<id>?docType=Converted` | module |
| --- | --- | --- |
| *(empty)* | 400 `CLIENT_NOT_CONFIGURED` | synchronization blocked — your case |
| CII | 200, `rsm:CrossIndustryInvoice` | read and imported |
| UBL | 200, `urn:oasis:…:ubl:schema:xsd:Invoice-2` | synchronization blocked, see below |
| Factur-X | 200, Factur-X PDF | read (XML extracted) |

`Metadata`, `Original` and `ReadableView` answered 200 in all four runs, so the account and the token are fine — only the conversion is unconfigured.

For the Factur-X row I checked the detection and the embedded-XML extraction only, not a full import, and I ran it on Dolibarr 18.0.10, 20.0.5, 23.0.3 and 24.0.0-beta (same PHP 8.4, so this varies the core, not PHP): all four detect `FACTURX` and pull the same 16 KB `CrossIndustryInvoice` out of the PDF. Worth noting since `FACTURX` is declared `protocol_dol_min = 24.0` in `ProtocolManager` — that minimum is never enforced (`getProtocol()` only looks at `is_enabled`) and does not stop the reading path on older cores.

So **setting that option to CII or Factur-X on your account fixes your instance today**, with no code change. It is also worth noting that the format cannot be overridden per request: `&format=CII`, `&conversionFormat=CII`, `&targetFormat=CII` and `&format=FACTUR-X` all return the same 400. It is strictly a property of the oauth2 client, and nothing in the module can see it or set it.

### Why `Original` is not the right fix

`docType=Converted` is precisely what protects the import from a document the module cannot read. UBL is one of the three formats of the French socle, so a supplier may legitimately send an invoice whose `Original` is UBL — and the module refuses UBL:

```
Original    HTTP 200  detected=UBL  getProtocol=NULL  =>  ERROR_FLOW_NOT_SUPPORTED_PROTOCOL
```

`UBL` is disabled in `ProtocolManager::__construct()` (`$ublIsOk = 0`) and `UBLProtocol` does not exist — the `case 'UBL'` branch is commented out. `syncFlow()` then returns `res = -1`, and `syncFlows()` does `break` on the first error, so the whole synchronization stops on that flow and on every flow behind it, run after run. That is the same symptom you are reporting, just triggered by a different account. This is also why #414 moved to `Converted` in the first place, and reverting only half of #414 leaves the `res = 0` → `res = -1` change of that PR in place, which is what makes the failure blocking rather than merely noisy.

I measured that case too: with the account set to convert to UBL, the platform returns a valid UBL document for a flow whose `Original` is CII, the module detects it correctly, and then refuses it. Two of the four possible values of that single account setting brick supplier invoice reception, and nothing in Dolibarr says so.

### Suggestion

Rather than choosing one `docType` and losing on the other configuration, the module should try the documents the flow offers until one is readable, and say what to do when none is:

1. keep `Converted` as the first choice;
2. fall back on `Original`, then on `ReadableView` — on this platform the readable view is itself a Factur-X PDF carrying the same XML, and it read correctly on all four Dolibarr versions above — when the returned document is in a format the module cannot read;
3. map `CLIENT_NOT_CONFIGURED` to a message naming the setting and the values that work, instead of relaying the raw platform text;
4. let the batch carry on past a flow whose document is unreadable, without recording it, so it is retried once the account is fixed instead of blocking every other flow.

I am opening a PR with that, crediting this report as its origin — it fixes your case, and the UBL one as well.

daGrumpf-bxp:
Follow-up: the alternative is now open as #540.

It keeps `docType=Converted` first, falls back on `Original` then `ReadableView` when the returned document is in a syntax this module cannot read, turns `CLIENT_NOT_CONFIGURED` into a recommended action naming the setting, lets the batch carry on past such a flow without storing it (so it is retried, never dropped), and states the requirement on the SuperPDP setup page.

Tested end to end on the sandbox, invoices really transmitted, in both configurations: with the account converting to UBL the fallback imports the invoice that #539's error currently blocks, and with the account converting to CII the module makes the exact same single call as today.

Your instance should also be unblocked right away by setting the preferred AFNOR conversion format of your SuperPDP application to CII, without waiting for either PR.

eldy:
I agree with @daGrumpf-bxp 
We must use "converted" format so retrieve invoice will always be CII format.
Original will return different format that depends on sender (that we can control).

So i close this pr in favor of #540

PR #540, opened 2026-08-03
fix(einvoicing): read a received supplier invoice through whichever document the access point can give

daGrumpf-bxp:
> **Stacked on #528 — please merge #528 first.** This branch now carries the commit of #528 on top of `main`, in addition to its own two.
>
> Both pull requests rewrite the same ~40 lines of `syncFlows()`: #528 wraps the flow loop in a batching `while` and re-indents its body, this one inserts the `postponeflow` branch inside that same loop. Whichever lands second conflicts — and the conflict git presents is misaligned: it offers the `$res`-based block of this pull request against the *batch fetch* failure block of #528, where neither `$res` nor `$flow` exists and which ends on a `return`. Resolving that by keeping both sides passes `php -l` and the whole test suite, and only breaks at runtime, on the first batch call that does not answer 200.
>
> Rather than leave that trap to whoever merges second, this branch is based on #528. Merging #528 first leaves this pull request a clean two-commit diff. Merging this one first carries #528 along and closes it. Both orders were checked: no conflict either way, and identical trees. The stack assumes a merge commit, as on #529 / #530 / #531 - squashing #528 would change the commit and bring the conflict back.

Origin: #539, which reports that synchronizing received supplier invoices fails on the first new flow with `ERROR_FLOW_GETORIG ... Configuration of the oauth2 client is required`, and proposes switching `docType=Converted` back to `Original`.

I could reproduce the report, and the cause is not the `docType`: it is the **preferred AFNOR conversion format** of the SuperPDP application, which the setup page never mentions. Switching to `Original` would not fix it either — it only moves the breakage onto anyone whose supplier emits UBL.

## What the access point actually does

Measured on one SuperPDP account, on one flow, changing nothing but that setting between runs:

| preferred conversion format | `GET flows/<id>?docType=Converted` | module |
| --- | --- | --- |
| *(empty)* | `400 CLIENT_NOT_CONFIGURED` | synchronization blocked — the case of #539 |
| CII | 200, `rsm:CrossIndustryInvoice` | imported |
| UBL | 200, `urn:oasis:…:ubl:schema:xsd:Invoice-2` | synchronization blocked |
| Factur-X | 200, Factur-X PDF | imported |

`Metadata`, `Original` and `ReadableView` answered 200 in all four runs, so the account and the token are fine. The format is not overridable per request either: `&format=CII`, `&conversionFormat=CII`, `&targetFormat=CII` all still return the same 400.

The UBL row is the one that matters for the fix proposed in #539: the document is valid, the module recognizes it, and then refuses it, because `UBL` is disabled in `ProtocolManager::__construct()` (`$ublIsOk = 0`) and `UBLProtocol` does not exist. UBL belongs to the French socle, so an issuer may legitimately send one — which is exactly what `docType=Converted` is there to absorb, and why #414 moved to it.

So two of the four values of a setting that lives outside Dolibarr stop supplier invoice reception, and in both cases `syncFlow()` returns `-1`, `syncFlows()` breaks on the first error, and every flow behind it stays stuck run after run.

## What this PR changes

1. **Read the flow through whichever of its documents can be imported** — `Converted` first, since it is what shields the import from an unsupported syntax, then `Original`, then `ReadableView` (on this platform the readable view is itself a Factur-X PDF carrying the same data). Each rejection is recorded with its reason.
2. **When none is readable, say what to configure** instead of relaying the raw platform text: `CLIENT_NOT_CONFIGURED` becomes a recommended action naming the setting and the values that work.
3. **Do not stall the batch on it.** The flow returns `-1` *before* the document is stored, so it stays pending and comes back on the next run — a received invoice is never dropped — but the run carries on with the other flows and reports the postponed ones apart. Only this specific failure is treated that way; every other error still aborts, since flows are processed in order and a business error needs a manual action first.
4. **State the requirement on the setup page**, where the SuperPDP application is created: the "Preferred AFNOR conversion format" must be CII (Factur-X also works), never empty and never UBL.

`docType=Converted` stays the first choice, so what #414 fixed is preserved.

## Before / after

Same instance, same account, conversion format set to UBL, syncing the same flows.

Before — the run dies on the first unreadable flow, and the only advice is to enable debug mode:

before, synchronization aborted on an unreadable document

After — the unreadable `Converted` document is stepped over, the readable one is imported, and the run completes:

after, the run completes

## Tests

End-to-end on the SuperPDP sandbox, two Dolibarr instances, invoices really transmitted through the platform:

- **Emission Dolibarr 23 → reception Dolibarr 20 with the conversion format set to UBL.** Invoice `TRI-FA-2608-0028` (1457.28 TTC) sent, received and imported with the same total. The module's own call log shows the mechanism: `flows/i_…?docType=Converted` (UBL, unusable) immediately followed by `flows/i_…?docType=Original` (CII, imported). Before the patch the same run ended `res=-1`, 0 flow imported, aborted on the very first flow.
- **Emission → reception on the instance whose account converts to CII.** Invoice `BQ-FA-2608-0034` sent, received and imported. A single `docType=Converted` call, no fallback: when the setting is right the behaviour is unchanged.
- **Same instance, same code, the account setting flipped from UBL to CII in between**, which is the clearest read of the change:

  | account setting | calls the module makes | outcome |
  | --- | --- | --- |
  | UBL | `docType=Converted` then `docType=Original` | fallback, imported |
  | CII | `docType=Converted` only | unchanged path, imported |

- **A hole in the middle of a batch, then the hole fixed** — the point being that a flow left unread must come back, not be lost. Three invoices emitted in a row, the middle one forced unreadable on the receiving side:

  | | flow | run 1 | run 2 (after the fix) |
  | --- | --- | --- | --- |
  | 1 | `i_193006` | imported | already processed |
  | 2 | `i_193007` | **postponed** | **imported** |
  | 3 | `i_193009` | imported | already processed |

  Run 1 ends `res=1`, 4 flows synchronized, 1 postponed, and `llx_einvoicing_document` holds flows 1 and 3 but **not** flow 2 — which is what makes it eligible again. Run 2 imports it: 1 new, 18 already existing, and the third supplier invoice is created with the same total as the other two. The lifecycle message that referenced the postponed invoice was skipped gracefully in run 1 (`No supplier invoice matching the vendor reference`) rather than failing.

- **Factur-X emission, CII reception.** With the account converting to CII, an invoice emitted as Factur-X arrives as `Original` = the Factur-X PDF the issuer really sent, `Converted` = `<CrossIndustryInvoice>`. Imported from the converted one, in a single call.
- Factur-X, the other value that works, was checked on **Dolibarr 18.0.10, 20.0.5, 23.0.3 and 24.0.0-beta** (same PHP 8.4): all four detect `FACTURX` on the converted PDF and extract the same 16 KB `CrossIndustryInvoice` from it. Worth noting since `FACTURX` is declared `protocol_dol_min = 24.0`, a minimum `getProtocol()` never enforces and which does not stop the reading path.

Unrelated to this PR but found on the way: in `syncFlows()`, `$res['actiondata']` holds sub-arrays, so the `THIRDPARTY_NOT_FOUND` branch concatenating `$dataval` raises `Array to string conversion` on PHP 8. Happy to send that separately.


daGrumpf-bxp:
Force-pushed: the two commits of this pull request are unchanged in content, they now sit on top of the commit of #528 (`a4980ec` → `c7631e2`). The reason is in the note at the top of the description: the two pull requests rewrite the same lines of `syncFlows()`, and the conflict git shows to whoever merges second is misaligned in a way that a "keep both sides" resolution turns into a runtime error rather than a visible one.

Both merge orders were replayed on a clean worktree from `main`: no conflict either way, identical resulting trees, `php -l` clean, and the module test suite green on 18.0.10, 20.0.4, 23.0.3 and 24.0.0-beta (the three `EInvoicingSamplesTest` failures on 18 and 20 are the ones #532 fixes; they are already there on `main` alone).

PR #544, opened 2026-08-04
[einvoicing] Say which VAT regime the invoice declares, and let the seller state it

### LINKED ISSUE #419: [Einvoicing] Manage VAT regime of main company in generated einvoice

daGrumpf-bxp:
Closes #419.

## What BT-8 means in the French socle

The issue asks for the VAT regime of the seller to reach the generated document, in `ram:DueDateTypeCode` (BT-8). Reading the normative material first was worth it, because BT-8 is narrower than its EN 16931 definition suggests. **XP Z12-012 annexe A** (the normative semantic model) introduces it as the

> « Champ permettant de spécifier l'option pour le paiement de la taxe d'après les débits »

and reads its three values as « **5** : date de la facture (TVA sur DEBITS) », « **29** : date de livraison (TVA sur DEBITS) », « **72** : date de paiement (TVA sur ENCAISSEMENTS) ». Two rules hang off it:

- **G1.43** (applies to flux 2): « Le BT-8 ne sera obligatoire que si l'entreprise a opté pour la TVA sur les débits et le spécifie au moyen du code 5 (CII) »
- **BR-FR-MAP-03**: « BT-8 est obligatoire pour les factures de service dès lors que l'assujetti Vendeur a opté pour les débits »

So `5` declares an option rather than a due date, and a seller who did not take that option has nothing to declare on a goods invoice. **XP Z12-014 annexe A** confirms that silence is meaningful, describing a VAT due on collection as « avec BT-8 absent, **ou bien** présent et signifiant à l'encaissement (72) ». The derivation the module already had follows all of that and is left untouched.

## What this changes

**FIX — a down payment invoice now declares 72.** XP Z12-014 annexe A settles that case against every other rule:

> « La TVA est exigible à l'encaissement de l'acompte pour les livraisons de biens comme pour les prestations de service, **même avec option sur les débits** »

That is already why the module reports the cash-in of a down payment to the platform with the status 212. But Dolibarr builds every down payment line as a goods line (`createDepositFromOrigin()` hardcodes `product_type = 0`), so under the default VAT mode the document declared nothing at all while its own lifecycle declared a collection. It now declares 72, whatever the VAT mode and whatever regime the seller declared.

**NEW — the regime can be declared explicitly**, with the setup option `EINVOICING_VAT_POINT_DATE_CODE`, for a seller whose regime the VAT mode of the Tax/VAT module cannot express. That mode still decides by default and **nothing changes for an instance that leaves the setting alone**. An explicit value applies to every document and carries the "TVA d'après les débits" legal mention and the scope of the status 212 with it, so the document, its mention and its lifecycle cannot contradict each other. It is also the only way to declare the exigibility at the delivery (29): the automatic derivation never produces that code, which says the same thing as 5 while **BR-FR-MAP-29** states that « le PPF attend uniquement 5 » — translating a 29 is the platform's duty on flux 1 / 10.1, not the issuer's, and 29 stays legal on flux 2.

The three questions — which code the document declares, whether the legal mention is due, whether the cash-in is reported — now read their answer from three functions in `einvoicing.lib.php` instead of three open-coded copies of the VAT mode.

## Decision table

| VAT mode | document | BT-8 | "VAT on debits" mention | status 212 |
|---|---|---|---|---|
| TAX_MODE 0 (French default) | goods only | *(none)* | no | no |
| TAX_MODE 0 | services / mixed | 72 | no | yes |
| TAX_MODE 0 | down payment | **72** | no | yes |
| TAX_MODE 1 (debits) | any | 5 | yes | no |
| TAX_MODE 1 | down payment | **72** | yes | yes |
| TAX_MODE 2 (all on payment) | any | 72 | no | yes |
| setting = 5 / 29 | any | 5 / 29 | yes | no |
| setting = 72 | any | 72 | no | yes |

## Validation

- **phpunit 71/71** on Dolibarr **18.0.10, 20.0.4, 23.0.3 and 24.0.0-beta**, including a new `VatPointDateCodeTest` covering the whole table. `phan --quick` clean.
- **Regression measured, not assumed**: 756 documents regenerated (22 invoices × 6 profiles × CII and Factur-X × the three VAT modes) and compared byte for byte against this branch's parent. The only document whose output differs is the down payment invoice.
- **1512 documents** (the same corpus under six VAT configurations) validated against the Factur-X / CII D22B schemas and the FNFE schematrons: 0 XSD error and no BT-8-related assertion failure on any profile, the pre-existing failures of that test corpus being identical across the six. Negative controls confirm the rules do fire: an out-of-list code trips `BR-CL-06`, adding BT-7 trips `BR-CO-03`, two distinct codes trip `CII-SR-462`.
- **End to end against a real Access Point**, both directions between two Dolibarr instances: 48 invoices covering {goods, services, mixed, down payment} × {TAX_MODE 0/1/2, setting 5/29/72}. All 48 acknowledged `Ok` by the platform, all 48 valid against the XSD, the CEN EN16931 schematron and BR-FR-Flux2 with **zero failed assertion**, and all received on the other side with BT-8 intact — including 29, which the platform does not rewrite on flux 2.

The specimen behind the reference fixtures now pins the VAT mode it generates under, like it already pins the language and the discount rounding, so those fixtures no longer depend on the setup of whoever regenerates them.


daGrumpf-bxp:
The new setting on the module setup page, above the reminder of the VAT mode it defaults to:

bt8-setup

PR #553, opened 2026-08-06
New: EINVOICING_PREFER_ORIGINAL — import issuer's Original (Factur-X) before Converted

Pichinov-Jose:
# New: `EINVOICING_PREFER_ORIGINAL` — import the issuer's Original document (its Factur-X) instead of the Converted one

## Problem
On import, `SuperPDPProvider::fetchImportableFlowDocument()` fetches the flow documents in the order `Converted → Original → ReadableView` and imports the first readable one.

When the recipient's access-point account is configured to **convert** incoming invoices to CII, the `Converted` document is a CII XML — so even when the issuer sent a **Factur-X** (carrying the human-readable PDF), the created supplier invoice ends up with only the XML attached. `FacturXProtocol` already attaches the PDF as `<ref_supplier>_einvoice.pdf`, but the CII `Converted` document is imported first and that path is never reached.

## Fix
Add an opt-in constant **`EINVOICING_PREFER_ORIGINAL`**. When set, the fetch order becomes `Original → Converted → ReadableView`, so the issuer's **Original** document (its Factur-X, with the readable PDF) is imported when available. The existing "skip a document whose syntax is not supported and try the next" logic is unchanged, so a UBL-only `Original` still falls back to the `Converted` document.

Default behaviour is unchanged (`Converted` first).

## Tested (live SuperPDP flow)
Issuer sending Factur-X, recipient account converting to CII. With the constant set, the log shows:

```
fetchImportableFlowDocument No usable 'Converted' document for flowId ... reading the 'Original' one instead
FacturXProtocol::createSupplierInvoiceFromSource ...
```

and the created supplier invoice now carries `<ref>_einvoice.pdf` (the issuer's Factur-X, byte-for-byte the emitted file) instead of only the CII XML.

## Notes
- Backward-compatible, opt-in via a single constant.
- Lets a recipient keep the emitter's Factur-X PDF on the created supplier invoice without changing the access-point account's conversion syntax.

---

### DIFF
--- a/einvoicing/class/providers/SuperPDPProvider.class.php
+++ b/einvoicing/class/providers/SuperPDPProvider.class.php
@@ -2,6 +2,7 @@
 /* Copyright (C) 2025       Laurent Destailleur         <eldy@users.sourceforge.net>
  * Copyright (C) 2025       Mohamed DAOUD               <mdaoud@dolicloud.com>
  * Copyright (C) 2026		MDW							<mdeweerd@users.noreply.github.com>
+ * Copyright (C) 2026		Jose Martinez				<jose.martinez@pichinov.com>
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -2446,7 +2447,12 @@ class SuperPDPProvider extends AbstractPDPProvider
 			'client_not_configured' => false
 		);
 
-		foreach (array('Converted', 'Original', 'ReadableView') as $docType) {
+		// EINVOICING_PREFER_ORIGINAL: fetch the issuer's Original document (its Factur-X, which carries
+		// the human-readable PDF) before the Converted one, so the created supplier invoice keeps the PDF.
+		$docTypeOrder = getDolGlobalString('EINVOICING_PREFER_ORIGINAL')
+			? array('Original', 'Converted', 'ReadableView')
+			: array('Converted', 'Original', 'ReadableView');
+		foreach ($docTypeOrder as $docType) {
 			$flowResponse = $this->fetchFlowData($flowId, $docType, 'get_flow_for_supplier_invoice');
 
 			if ($flowResponse['status_code'] != 200) {


eldy:
Having all einvoicing received and stored in the same format is currently the expected behaviour.
The original format can also be saved but as a second data, but this isbpart of another feature.
Note that if you always want to receive factur-x, you must set this option from your 
Pa dashboard that defines what is the "converted" format.
Saving into a format that will depends on sender will probably generate a lot of trouble, forcing the module to always switch between format but this may wotk if finding facturx and cii only.
Bit what happen with your option if sender isbusing ubl format ?

daGrumpf-bxp:
Some context on why the current order is a chain of three and not just `Converted`, since it is the thing your constant reverses.

### Why not `Converted` alone

`Converted` on its own is not something the module can rely on, because whether it exists at all is decided **outside Dolibarr**, on the access-point oauth2 client. That is what issue #539 turned out to be: the synchronization died on the first new flow with *"Failed to retrieve 'Original' document - Configuration of the oauth2 client is required"*, and every flow queued behind it stayed stuck.

Measured on a SuperPDP account, same flow, changing only the preferred conversion format of the client:

| conversion format set on the PA | result of `docType=Converted` |
|---|---|
| unset | HTTP 400 `CLIENT_NOT_CONFIGURED` — no document at all |
| CII | a document this module reads |
| Factur-X | a document this module reads |
| UBL | a valid document this module **refuses** (UBL is part of the French socle but has no implementation here) |

So two of the four possible values of a setting that is invisible from Dolibarr stop supplier invoice reception completely. Neither of those two failures says anything about the *other* documents of the same flow, which is the whole point: rather than betting on one shape and failing the flow on the first miss, the module tries `Converted`, then `Original`, then `ReadableView`, and imports the first one it can actually read (PR #540).

`Converted` stays first because it is the one that shields the import from an issuer emitting a syntax the module does not read — that is exactly what PR #414 fixed when it moved away from `Original`. And when none of the three is readable, nothing is stored for that flow, so it comes back on the next run instead of blocking the batch: a received invoice is never dropped.

### Where this is written down

The reasoning is in the docblock of `SuperPDPProvider::fetchImportableFlowDocument()` (`einvoicing/class/providers/SuperPDPProvider.class.php`), right above the `foreach (array('Converted', 'Original', 'ReadableView') as $docType)` loop — the three shapes, what each one is, and why the order is what it is. Worth reading before changing the order, since the fallback also carries the #539 fix. (It is not in the README today; if it should be user-facing documentation rather than a code comment, that is a fair separate point.)

### The conversion has to be configured on the PA

This is the part that matters for your use case, and it is what @eldy is pointing at: **`Converted` only exists if the preferred conversion format is set on the access-point client**. It is a setting on the PA dashboard, not a Dolibarr one.

If what you want is to always receive Factur-X, setting the conversion format to **Factur-X** on the access-point client gives you exactly that through the normal `Converted` path — with the readable PDF, without flipping the order, and without the received format depending on what each sender chose to emit. That is the difference: the PA setting gives one format for everybody, the constant gives whatever the issuer sent.

### On the UBL question

To answer it factually for your patch: with `EINVOICING_PREFER_ORIGINAL` set and an issuer emitting UBL, the loop rejects the `Original` document (`Original: UBL is not supported`) and falls through to `Converted`, so it does not break — the existing skip-and-try-the-next logic covers it. The objection stands on the other ground though: the format of the created supplier invoice then depends on the sender rather than on your own configuration.

eldy:
@daGrumpf-bxp i think convertion format is always set on AP. If it was not in past for some superpdp account, I think it is because the feature (mandatory for all AP) was not yet implemented, but now, for new superpdp account, this setup exists so should be always set i suppose...

daGrumpf-bxp:
@eldy i've recently add a new oauth2 app on superpdp converstion isn't select by default, cust have to select it else superpdp send 400 on converted request as explain on recent PR

PR #555, opened 2026-08-07
FIX #554 : a Factur-X built on Dolibarr < 24 is not a Factur-X file

### LINKED ISSUE #554: Factur-X generation on Dolibarr < 24 yields a non-compliant PDF (no /AF, no PDF/A-3) unless MAIN_DISABLE_TCPDI is set
## Summary
On Dolibarr **< 24.0**, generating a Factur-X invoice with this module can produce a **non-compliant** PDF: the `factur-x.xml` is embedded, but the document lacks the PDF/A-3 `OutputIntent` and the document-level `/AF` (Associated Files) entry that a Factur-X reader/PDP validator expects.

## Root cause
`FacturXProtocol::generateInvoice()` embeds the XML through two code paths, selected by the `class_exists('FPDF', false) && is_subclass_of('FPDF', 'TCPDF')` test:

- **≥ v24**: horstoeko/zugferd `ZugferdDocumentPdfMerger` (via `CtcFrPdfMerger`) → proper PDF/A-3 with `/AF`.
- **< v24 (fallback)**: a TCPDF `Annotation(FileAttachment)` path.

On Dolibarr < 24, `htdocs/includes/tcpdi/tcpdi.php` defines `class FPDF extends TCPDF {}`. As soon as any PDF is rendered in the request — e.g. the invoice PDF produced at validation, right before the e-invoicing hook runs — this `\FPDF` is loaded. The zugferd/setasign FPDI merger needs the *real* `\FPDF`, so it fails with `Call to undefined method horstoeko\zugferd\ZugferdPdfWriter::_getpagesize()`, and the module falls back to the TCPDF path — which does **not** emit `/AF` nor the PDF/A-3 `OutputIntent`.

## Observed (Dolibarr 23.0.x)
A Factur-X generated in the normal validation flow:

| Marker | Result |
|---|---|
| `/Type /EmbeddedFile` (XML embedded) | present |
| `/AF [ … ]` (Associated Files) | **absent** |
| `OutputIntent` (PDF/A-3) | **absent** |

Setting **`MAIN_DISABLE_TCPDI=1`** (so `pdf_getInstance()` never loads `tcpdi.php`/`\FPDF`) makes the zugferd path run on v23 → the same invoice then carries `/AF` + PDF/A-3 and the underlying invoice model still renders correctly.

⚠️ A **scoped** set/restore of the constant around the generation does **not** work: once `\FPDF` is loaded in the PHP request it cannot be unloaded, and the module keys off `class_exists('FPDF')`. The constant has to be effective **before the first `pdf_getInstance()` of the request**, which only a persistent global value guarantees.

## Suggestion (from lightest to most complete)
1. **Document** the `MAIN_DISABLE_TCPDI=1` workaround for Dolibarr < 24 on the module setup page — with the caveat that it disables TCPDI-based PDF import/merge globally (PDF backgrounds, PDF concatenation, mass PDF merge…), so it is only safe when those features are not used.
2. **Warn** on < v24 when the produced Factur-X is missing `/AF`/PDF/A-3, instead of silently producing a degraded file that a PDP may reject.
3. Longer term, make the < v24 path produce a compliant file — e.g. via the `atgp/factur-x` library already noted as a `TODO` in `FacturXProtocol`, or by running the merge in an isolated subprocess where Dolibarr's `\FPDF` is not loaded.

Happy to help with a PR for (1) and/or (2).

## Environment
Dolibarr 23.0.x · PHP 7.4 · einvoicing 1.0.3 · horstoeko/zugferd v1.0.123.

daGrumpf-bxp:
Fixes #554. Thanks @Pichinov-Jose for the analysis, it is accurate down to the root cause.

## What was wrong

Below Dolibarr 24 the module produced a plain PDF with the XML attached to it: no document level `/AF`, no PDF/A-3 output intent, and the XML embedded twice. Nothing reported it, and the file is refused further down the line. The public validator on a file produced by `main` on Dolibarr 23.0.3:

```
POST https://api.superpdp.tech/v1.beta/validation_reports  -F file=@BEFORE.pdf
{"is_valid": false, "format": "", "conformance_level": "",
 "error": "Le fichier ne contient pas un et un seul fichier factur-x.xml"}
```

The cause is a class name collision. `htdocs/includes/tcpdi/tcpdi.php` declares its own `class FPDF extends TCPDF {}` with no guard, and the horstoeko/zugferd writer needs the real `\FPDF` underneath. The first declaration in a PHP request wins and the other one can never be loaded, so a request that has already rendered a PDF — the invoice PDF produced at validation, right before our hook runs — makes the zugferd writer inherit from TCPDF and die on `_getpagesize()`. The code caught that with a fallback that silently degraded the output.

Two things follow from that, both verified rather than assumed:

* the zugferd path itself works perfectly on Dolibarr 23 **when the shim is not loaded** (a CLI run produces a fully conformant file). The version was never the problem, the load order is.
* the collision is not resolvable in-process, whatever the order: preloading the real FPDF makes the later `require_once` of `tcpdi.php` fatal on a redeclare.

## What this does

The way out is not to need `\FPDF` at all. setasign/fpdi also ships a **namespaced TCPDF flavour** of its page importer (`setasign\Fpdi\Tcpdf\Fpdi`), which the shim cannot touch, and TCPDF supports PDF/A-3 natively. `FacturxTcpdfMerger` puts the two together and produces the same structure as the v24 path, which keeps running unchanged and is untouched by this PR.

Three TCPDF gaps are filled in `_out()`, the single point every PDF object goes through (object offsets are recorded when an object starts, so lengthening a body there shifts nothing):

* the catalog gets the document level `/AF` array — TCPDF writes the `/Names /EmbeddedFiles` tree, which makes the file downloadable, but a Factur-X reader looks up `/AF`;
* the file specification gets `/AFRelationship /Data` instead of TCPDF's hardcoded `/Source`, which is what the v24 path emits;
* the embedded stream gets its `/Filter` back: in PDF/A-3 mode TCPDF *overwrites* the filter entry with the attachment subtype instead of adding it, so a compressed attachment would be written deflated and declared as plain. TCPDF never meets the case because it also refuses to compress in PDF/A mode, which we turn back on (PDF/A bans LZW and encryption, not Flate).

Three smaller things came with it:

* **the produced file is checked before being handed over**, whichever merger ran, so an incomplete Factur-X is reported as a warning instead of being sent (suggestion 2 of the issue);
* **a carrier PDF that cannot be read now aborts with an explicit message**. Before, TCPDI printed `<strong>TCPDI_PARSER ERROR ...</strong>` and `die()`d mid-request, leaving a 34-byte `_facturx.pdf` behind — a file that looks like the e-invoice and is not one;
* **the sample invoice of the setup page** was hit by the same collision from a page that had already rendered a PDF: a PHP fatal error, uncatchable, blank setup page. It now uses the same merger.

Since Factur-X now produces the same file on every supported version, the "(Dolibarr 24.0+)" hint on the protocol selector is gone — say the word if you would rather keep it.

I did not take suggestion 1 of the issue (documenting `MAIN_DISABLE_TCPDI=1`): it works, but it disables TCPDI-based PDF import and merge instance-wide (PDF backgrounds, concatenation, mass merge), which is a high price for something the module can do on its own.

## Tests

All of it run for real on a version matrix (Dolibarr **18.0.10 / 20.0.4 / 23.0.3 / 24.0.0-beta**, PHP 8.4), never deduced.

**Unit** — the 9 test files, one PHP process each, on the 4 cores: 36 runs, all green.

**Generation, real invoices** — the 8 most recent validated invoices of each instance regenerated the way the web flow does it (invoice PDF first, so the shim is loaded), 32 files: all carry the embedded XML, `/AF`, the output intent and the Factur-X XMP, with no warning. Standard invoices, credit notes, down payments and a replacement invoice among them.

**Generation, sample invoice** — 2 protocols × 4 profiles × 4 cores = 32 runs, all produce a file. On `main`, the Factur-X rows are a PHP fatal error on 18 / 20 / 23 as soon as a PDF was rendered first.

**External validation** — five representative outputs through the public SuperPDP validator, XSD + `FACTUR-X_EXTENDED.xslt` + `BR-FR-Flux2-Schematron-CII`:

| file | valid | checks | failures |
|---|---|---|---|
| standard, Dolibarr 18 | yes | 168 | 0 |
| replacement, Dolibarr 20 | yes | 337 | 0 |
| down payment, Dolibarr 23 | yes | 165 | 0 |
| 4-page invoice (60 lines), Dolibarr 23 | yes | 1697 | 0 |
| credit note, Dolibarr 24 | yes | 171 | 0 |

Negative control: the same generation on `main` / Dolibarr 23 is the rejected file quoted at the top, so the tool does say no.

**Fidelity** — the produced Factur-X rendered page by page (`pdftoppm`) against the invoice PDF it was built from: byte-identical images, on a 1-page and on a 4-page invoice, geometry preserved. The embedded stream extracted back: declared `FlateDecode`, inflates, well-formed XML.

**Error path** — carrier PDF replaced by garbage: `generateInvoice()` returns `-1` with `"The e-invoice XML cannot be embedded into the PDF ...: Unable to find pointer to xref table"` and no file is left behind. Same scenario on `main`: `TCPDI_PARSER ERROR` printed as HTML, request killed, stub left on disk.

**End to end, both directions, through the real platform** — two sandbox instances in Factur-X, fresh invoices, SuperPDP:

* **Dolibarr 18 → Dolibarr 23**: emitted (`flowSyntax: Factur-X`, `acknowledgement: Ok`), imported as a supplier invoice on the other side — checked from a separate process and connection — lifecycle back to the sender `200 → 201 → 205`, then payment recorded and `212` sent.
* **Dolibarr 23 → Dolibarr 18**: same chain, `200 → 201 → 205 → 212`. The imported supplier invoice carries the amounts of the source invoice exactly (210.00 / 42.00 / 252.00, 1 line).

## One thing noticed, left alone

`generateSampleInvoiceOld()` does `require __DIR__ . "/ExampleHelpers.php"` rather than `require_once`, so generating two samples in a single PHP request is a fatal redeclare. It does not bite a real page (one click, one request) and it has nothing to do with this issue, so I left it out rather than widen the diff. Happy to send it separately.

## File size

On Dolibarr < 24 the output is heavier than the v24 one, by a fixed amount, not a factor: TCPDF embeds a PDF/A copy of its default font (~26 KB) plus the sRGB profile and its own XMP block. A one-page invoice goes from 12 KB to ~44 KB, a four-page one from 17 KB to 54 KB. Dropping the font is not safe — TCPDF references it from the page resources even when no glyph is drawn — and correctness seemed the better trade. `EINVOICING_MAX_FILE_SIZE_MB` still covers the case where an Access Point caps the size.


einvoicing/class/protocols/ProtocolManager.class.php
```diff
@@ -58,8 +58,7 @@ public function __construct($db)
 			),
 			'FACTURX' => array(
 				'protocol_name' => 'FACTURX',
-				'protocol_dol_min' => '24.0',
```
eldy, line 61: I 'm afraid,we can experience a lot of other problem if we try to use facturx on version < 24, due to the fact that dolibarr load first it's version of fpdf before the lib horstoeko do it. For example, when using external modules that also manipulate lib and pdf.

For this reason, I think we must continue to discourage use of the awful facturx format in favor of cii as much as possible and above all on version < 24.

daGrumpf-bxp: Thanks for looking at it — and the concern is the right one to raise, because it is precisely the one this PR is built around. Let me show why it no longer applies to this path.

**Below v24 the Factur-X merge does not go through `\FPDF` at all anymore.**

`FacturxTcpdfMerger` extends `setasign\Fpdi\Tcpdf\Fpdi`, which is `class Fpdi extends \TCPDF` — the namespaced TCPDF flavour of the page importer, which the shim cannot reach. From horstoeko/zugferd it uses only `ZugferdProfileResolver` and `ZugferdProfiles` to resolve the guideline URN, plus the XMP asset file: metadata, no PDF writer. `ZugferdDocumentPdfBuilder` / `ZugferdPdfWriter` — the classes that descend from `setasign\Fpdi\Fpdi` and therefore from `\FPDF` — are never instantiated on that branch. So "Dolibarr loads its own FPDF before the horstoeko lib does" no longer has a consequence: nothing asks horstoeko to write a PDF.

**And the branch is taken on the actual state of the request, not on the version:**

```php
if (class_exists('FPDF', false) && is_subclass_of('FPDF', 'TCPDF')) {
```

which is why your own example — an external module that also manipulates the PDF libs — is an argument *for* this shape rather than against it. I checked on a 24.0.0-beta: `htdocs/includes/tcpdi/tcpdi.php` is still shipped, with line 22 `//class FPDF extends TCPDF {}` commented out. So v24 is safe by default, but the problem is not version-bound, it is shim-bound. If anything in a request does hold an `FPDF extends TCPDF` — a module bundling its own tcpdi, say — then on current `main` that request produces a silently degraded Factur-X (`generateInvoice`, the annotation branch) or an uncatchable PHP fatal (`generateSampleInvoiceOld`, which has no branch at all), on **any** Dolibarr version, 24 included. With this PR it falls back to TCPDF and produces a conformant file, and the structure check added here reports the case instead of handing the file over, whichever merger ran.

**On continuing to discourage Factur-X: no argument from me.**

One thing worth knowing before you decide, though: `protocol_dol_min` is display-only. Its single use is `admin/setup_options.php:123-124`, which appends a grey `(Dolibarr 24.0+)` to the label in the selector. It never disabled the protocol and never stopped anyone from selecting it — which is exactly how a < 24 instance ends up emitting a file the platform refuses, with nothing said about it. So the hint and this fix are complementary rather than alternatives: the hint steers people towards CII, the fix means those who pick Factur-X anyway do not send a broken document.

Say the word and I put the line back in this PR.

LuckyRoussel98: @daGrumpf-bxp , @eldy ,

**Independent field confirmation on Dolibarr 22.0.5 — and why this PR is the right layer for the fix**

We hit the exact issue of #554 on a stock Dolibarr 22.0.5 (protocol FACTURX): the fallback branch produced a plain PDF with the XML as a mere annotation — no `/AF` entry, no PDF/A-3 identification, no Factur-X XMP — rejected by validators.

To confirm the root cause, we fixed it from the other side: we backported the v24 core change into our 22 tree (`includes/tcpdi/`) — commented out the `class FPDF extends TCPDF {}` shim and re-parented `class FPDF_TPL extends TCPDF` directly. That alone was enough: with the global name freed, the module's existing `CtcFrPdfMerger` branch runs as designed, and the binary output checks out on real invoices — `pdfaid:part=3` + `conformance=B`, OutputIntent, full Factur-X XMP extension schema, exactly one embedded `factur-x.xml`. TCPDI's inheritance
chain is unchanged (the shim was an empty link), and nothing else in a stock 22 tree references the global `FPDF`.

Two teams, two different fixes, one identical diagnosis: **the only obstacle below v24 is the `\FPDF` name collision.**
The two fixes don't compete — they operate at different layers:
- **A core-side fix heals the platform.** Once the shim stops squatting the global name (which is exactly what v24 already does), *any* module shipping unscoped FPDF-based libraries works. But it only helps instances whose core is patched — and this module runs on thousands of stock 18–23 cores that nobody can patch retroactively.
- **This PR makes the module immune wherever it lands.** By going through the namespaced `setasign\Fpdi\Tcpdf\Fpdi`, the merge below v24 never touches `\FPDF` at all. As far as we
  can see, this fully answers the load-order concern: load order cannot matter for a class that is never referenced.

Since the module cannot assume a fixed core, the module-side fix is the only one that helps every user now. The integrity check before hand-off is a welcome bonus: a wrong file failing loudly beats a silent transmission to a PDP.
We run several 22.0.5 instances and are happy to test this branch on them (we would temporarily revert our core backport first, to test against stock 22 behavior). 
----InfraS----

daGrumpf-bxp:

Factur-X, emitted from **Dolibarr 23.0.3** — the version range this branch is about — and exchanged for
real through the SuperPDP sandbox. One new invoice, `EINVOICING_PROTOCOL=FACTURX`:

```
emitter  dolidev 23.0.3 -> TRI-FA-2608-0134, flow i_214968, TRI-FA-2608-0134_facturx.pdf produced
Access Point   : acknowledgement Ok
receiver dd23 23.0.3 -> supplier invoice 1035, ref_supplier TRI-FA-2608-0134, ht 100.00
statuses back on the emitter : 200 (ie_733582), 201 (ie_733583)
```

And the carrier itself, same instance, same invoice shape, before and after the branch:

| | `main` | this branch |
|---|---|---|
| document level `/AF` | yes | yes |
| PDF/A-3 `OutputIntent` | **no** | yes |
| XMP `pdfaid:part` | **absent** | `3` |
| XMP `fx:ConformanceLevel` | **absent** | `EXTENDED` |
| XMP `fx:DocumentType` | **absent** | `INVOICE` |
| size | 67 247 bytes | 98 324 bytes |

The produced PDF passes the reference validator as a Factur-X document: `is_valid true`, conformance
`urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended`, 0 failure over the XSD + 102 EXTENDED
checks + 65 BR-FR checks.

PR #564, opened 2026-08-08
FIX #560 : a seller that charges no VAT does not identify itself, and the invoice is refused (BR-E-02)

### LINKED ISSUE #560: BR-E-02 when company is configured as VAT exempt
### Bug

Generated invoices are rejected with BR-E-02 when the company is configured as "Non assujetti à la TVA".

The generated XML contains the SIREN as BT-47 (SpecifiedLegalOrganization) but seems to miss the identifier required by BR-E-02 when no VAT number exists.

The invoice is rejected by SuperPDP validator and transmission platform.

### Module

einvoicing

### Dolibarr Version

22.0.3

### Environment PHP

8.1.34

### Environment Database

11.4.12-MariaDB

**Steps to reproduce**
1. Setup a company with: VAT management -> "Non assujetti à la TVA"
2. Create a customer invoice with 0% VAT
3. Validate the invoice
4. Generate the e-invoice through EInvoicing
5. Submit the generated file to SuperPDP validator or transmission platform
6. The invoice is rejected with BR-E-02

**Current behavior**
The generated XML contains:
<ram:SpecifiedLegalOrganization>
    <ram:ID schemeID="0002">_SIREN HERE_</ram:ID>
</ram:SpecifiedLegalOrganization>
The seller SIREN is therefore present as a legal identifier (BT-47).
However, the generated invoice is rejected with BR-E-02.

**Expected behavior**
The generated invoice should pass EN16931 validation and be accepted by the PDP for companies configured as "Non assujetti à la TVA".
According to AFNOR XP Z12-012, when no VAT number exists, the SIREN may need to be provided as the seller tax registration identifier (BT-32).

### Attached files

_No response_

daGrumpf-bxp:
Fixes #560. Thanks @IngeniousDavid for the analysis and the proposed patch — the element, the scheme and the value are the right ones, and this builds on it. Two things made me widen it, both checked against the schematron and the core rather than assumed.

## What was wrong

A company set as "VAT is not used" has no VAT number, and the CII writer only ever emitted one:

```php
if (!$minimal && !empty($data[$prefix . 'vatnumber'])) {   // schemeID 'VA' — BT-31
```

so the seller party carried **no `ram:SpecifiedTaxRegistration` at all**. Every exempt line then broke BR-E-02, which wants BT-31, BT-32 *or* BT-63. Reproduced on Dolibarr 23.0.3 in the exact setup of the report (`FACTURE_TVAOPTION=0`, no `MAIN_INFO_TVAINTRA`), the public validator gives one single failure out of 92 checks:

```
[BR-E-02] ... shall contain the Seller VAT Identifier (BT-31), the Seller tax registration
identifier (BT-32) and/or the Seller tax representative VAT identifier (BT-63).
```

and the platform refuses the flow: `acknowledgement: Error`, `flowProfile: Undefined`.

Recording an exemption reason code, as the wiki suggests, does not help — @xylphe's message says so and the reproduction confirms it. That code answers **BR-E-10** (the *reason*); BR-E-02 is about the *identifier*. The generated document carried `VATEX-FR-FRANCHISE` and failed all the same.

## What this does

The seller declares whichever identifier its VAT regime calls for: its VAT number under `VA` (BT-31) when it charges VAT, its SIREN under `FC` (BT-32) when it does not.

**The regime is read from the core, not re-derived.** `Societe::setMysoc()` already turns `FACTURE_TVAOPTION` — what the "VAT is used / is not used" radio of `admin/company.php:257` writes as `1`/`0` — into `->tva_assuj`, and `getCategoryRate()` already decides from that same `->tva_assuj` whether a line is exempt. Reading the property the core computed is what keeps the two from disagreeing: a document declaring an exempt line while claiming a VAT registration, or the reverse, is not something that should be possible.

**And no option of this module states it.** An earlier revision of this branch offered the choice in the setup; @eldy pointed out the risk in review and he is right, so it is gone. A second place to declare the regime is a second place for it to disagree with Dolibarr, and since the VAT category of every line comes from the same `->tva_assuj`, that disagreement produces exactly the document nobody can explain: exempt lines next to a claimed VAT registration. A constant left over from that revision is ignored, and a test pins it in both directions.

Worth noting for anyone reading the core: `->tva_assuj` is not the same type on every supported version, and this module declares Dolibarr **17** as its minimum. Checked in the upstream sources of 17 through 24, not only on the cores available here:

| Dolibarr | `setMysoc()` | `->tva_assuj` |
|---|---|---|
| 17 / 18 / 19 | `$conf->global->FACTURE_TVAOPTION` | the **string** `'0'` / `'1'` |
| 20 → 24 | `getDolGlobalInt('FACTURE_TVAOPTION')` | the **int** `0` / `1` |

So the reading here is the one `get_default_tva()` itself uses, which covers both — there is no version test anywhere in this diff. Both forms are pinned in the unit test rather than left to whichever core runs the suite, since 17 and 19 are not among the four I can execute on. `->idprof1` needed the same check and gives the same answer everywhere: it is `MAIN_INFO_SIREN` from 17 to 24, only the syntax of the read changed with `getDolGlobalString()`.

**Two cases deliberately left out of the fallback.**

* A seller *subject to* VAT that simply left the field empty keeps getting the `BADVATNUMBER` message naming what to fill in. Falling back on its SIREN would produce a document that passes BR-E-02 while stating the company is not registered for VAT, and would hide a missing field instead of reporting it.

* **An exportation and an intracommunity supply cannot use BT-32 at all.** This is the one point on which I had to depart from the proposed patch, and it comes straight out of the schematron we vendor:

  | rule | category | assertion accepts |
  |---|---|---|
  | BR-E-02, BR-Z-02, BR-AE-02, BR-S-02 | E, Z, AE, S | `@schemeID = ('VA', 'FC')` |
  | **BR-G-02** (export) | G | `@schemeID = 'VA'` **only** |
  | **BR-IC-02** (intracom) | K | `@schemeID = 'VA'` **only** |

  So emitting `FC` whenever no VAT number is present turns a clear failure into a silent one for those two categories: the document is still refused, but nothing in the module connects the refusal to a missing field. Those two paths now stop with an explicit message, the same way BR-S-02 already does. This also fixes a pre-existing case: a seller with an incomplete company setup was already producing invalid export documents, quietly.

## A second defect found on the way, and fixed here

Once BR-E-02 was gone, the document was still returned as **invalid** by the reference validator, on `BR-FXEXT-E-08`. That rule reconciles the taxable amount of an exempt VAT breakdown (BT-116) with the sum of the net amounts of the lines it covers, and its selector only counts a line whose own reason code and reason text equal those of the breakdown. Both writers put them on the breakdown alone, so it counted zero lines:

```
Code : E, rate : 0, Exemptcode : VATEX-FR-FRANCHISE, basisAmount : 100, SumBT131 : 0, NBlines : 0
```

It is worth being explicit about why this is in the same PR: without it, an exempt invoice is still refused, which is what the issue is about. Measured on Dolibarr 24, same instance, same EXTENDED profile:

| Factur-X | `is_valid` |
|---|---|
| exempt seller, BT-32 only | **false** |
| exempt seller, BT-32 + line reason | **true** |
| VAT-liable seller (control) | true |

The reader of `CIIProtocol` already expected both elements at line level, so this closes a loop rather than opening one. On the CII side they are written in the D22B sequence of `TradeTaxType`; on the Factur-X side they are the fifth and sixth arguments of `addDocumentPositionTax()`, which were simply left unset. A line with nothing to declare is unchanged.

## Tests

Run for real on the version matrix (Dolibarr **18.0.10 / 20.0.4 / 23.0.3 / 24.0.0-beta**, PHP 8.4), never deduced.

**Unit** — the 10 test files, one PHP process each, on the 4 cores: **40 runs, all green**, including a new `SellerVatRegimeTest` (7 tests) covering the derivation, the override, the literal forms of `tva_assuj`, and the two "declare nothing" cases.

**Non-regression** — the same standard 20 % invoice generated before this branch and after it: the two XML files differ only by the invoice date, which is the day they were created on. No structural difference.

**Schematron, locally with Saxon-HE 10.9** against the XSLT of the EUPL reference repository (`fnfempe/France_RFE`), including the two an online validator does not chain: the error-level `BR-FR-Flux2-Schematron-CII.xslt`, and `EXTENDED-CTC-FR-CII.xslt`, the schematron of the profile this module emits.

| document | EN16931-CII | BR-FR-Flux2 | EXTENDED-CTC-FR |
|---|---|---|---|
| before the fix (Dolibarr 23) | **1 failed** `BR-E-02` | 0 / 65 | **2 failed** |
| after, Dolibarr 18.0.10 | 0 / 92 | 0 / 65 | **0 / 94** |
| after, Dolibarr 20.0.4 | 0 / 92 | 0 / 65 | **0 / 94** |
| after, Dolibarr 23.0.3 | 0 / 92 | 0 / 65 | **0 / 94** |
| after, Dolibarr 24.0.0-beta | 0 / 91 | 0 / 65 | **0 / 93** |

The four also come back `is_valid: true` from the public validator; before the fix the same generation is refused.

**Factur-X** — the other writer this touches, generated on Dolibarr 24 with a seller not subject to VAT. The PDF carries `/AF`, `/EmbeddedFile`, the output intent and `factur-x.xml`; the embedded XML carries `<ram:ID schemeID="FC">` and the line-level reason, validates 0 / 91, 0 / 65, 0 / 93, and the PDF itself is returned **`is_valid: true`**.

**End to end through a real Access Point** — a fresh invoice, seller not subject to VAT, 0 % line, Dolibarr 23 → SuperPDP: `acknowledgement: Ok`, `flowProfile: CIUS`. The same invoice before the fix: `acknowledgement: Error`, `flowProfile: Undefined`.

**Error path** — seller subject to VAT with no VAT number recorded, 0 % invoice to a Swiss customer: generation stops with

```
BADVATNUMBER[BR-G-02]: The VAT number of the seller ... is mandatory on an exportation outside
the EU: that rule accepts the seller VAT identifier (BT-31) only, not the tax registration
identifier (BT-32) a seller charging no VAT would declare.
```

and the same invoice with the VAT number restored generates normally, so the guard has no false positive.

`phpcs` clean on every file touched.

## One thing left open

No rule in any of the three schematrons constrains the *content* of BT-32: `FC` appears only inside the `schemeID = ('VA', 'FC')` assertions of the BR-*-02 family, and the French schematron does not mention `FC` or BT-32 at all — no format check, unlike BR-FR-10 / BR-FR-32 on BT-30. So putting the SIREN there is a semantic choice rather than a rule-checked one. It is what BT-30 already carries under scheme 0002, it is what @IngeniousDavid reported as accepted up to status 203, and it passes end to end here — but if someone has the normative wording of BT-32 at hand and it points elsewhere, I will follow it.


eldy:
Just one comment: if user is not subject to vat, i think he is not qualified to use einvoincing and can disbale the module, right ?

casta:
> Just one comment: if user is not subject to vat, i think he is not qualified to use einvoincing and can disbale the module, right ?

Hello,

No, the module is still needed:
* you MUST receive einvoices from your suppliers
* you CAN send einvoices to your clients, even without VAT. It’s just not mandatory, but I think this is a plus. For example, to public clients, I manually upload PDF to chorus. Using the module, I can integrate with chorus through superpdp.
* In 2027 you will MUST do ereporting through the PDP (which is in the roadmap I think).


And just tested, with this patch, einvoinces are now valid ! Thank’s !!

daGrumpf-bxp:
@casta answered it from the field better than I could, and I agree with the three points.

To add only what I measured: on the four cores, a seller not subject to VAT emits and receives normally once BT-32 is there, and the reference validator goes from refused to accepted. Being exempt changes *which* identifier the document carries — BT-32 instead of BT-31 — not whether the company exchanges e-invoices at all. Reception is not optional for anyone, and e-reporting will not be either.

einvoicing/admin/setup_options.php
```diff
@@ -283,6 +283,18 @@
 	$item->defaultFieldValue = 'auto';
 	$item->cssClass = 'minwidth500';
 
+	// Which tax identifier the seller declares, BT-31 or BT-32. Left to the "Sales tax type" of the
+	// company setup by default; an explicit value is for a seller whose regime it cannot express
+	// (issue #560).
+	$item = $formSetup->newItem('EINVOICING_SELLER_VAT_REGIME')->setAsSelect(array(
+		'auto'      => $langs->transnoentities('EINVOICING_SELLER_VAT_REGIME_AUTO'),
```
eldy: Allowing to set franchise when user has set Use vat seems dangerous.
The litigious case is only if user has set not use vat (it may be because franchise is used for french user). There is a risk of bad consistency between option. For exemple if user has set "vatvis used" and set here franchise.

daGrumpf-bxp: You are right, and the option is gone (a748d18). No choice is left to the user: the regime is read from
the company setup and from nowhere else.

The inconsistency you describe is not even hypothetical, because the same setting decides two things at
once. `getCategoryRate()` derives the VAT category of every line from `$mysoc->tva_assuj`, so with "VAT
is used" and "franchise" selected side by side the document would have carried standard-rated lines
together with a seller declaring no VAT registration - or the reverse - and no operator could have
diagnosed that from the two screens. One place to state it, and it is Dolibarr's.

So `einvoicingSellerVatRegime()` now reads the `->tva_assuj` the core computed, nothing else. A constant
left over from this branch is ignored, and a test pins that in both directions.

Run on the bench (Dolibarr 17 → 24, same clone):

- the module test suite, file by file, on 23.0.3 and 18.0.10: all green (10 files); `SellerVatRegimeTest`
  passes on the eight versions 17.0.4 → 24.0.0-beta;
- the drift case for real on 23.0.3 - `FACTURE_TVAOPTION` = 0 with the leftover constant set to
  `standard` → regime `franchise`, `[{"type":"FC","value":"000000002"}]`; the opposite pair → `VA`. The
  setting wins in both directions;
- a real customer invoice generated by a seller set "Non assujetti à la TVA", with the leftover constant
  still saying `standard`: the document carries `<ram:ID schemeID="FC">000000002</ram:ID>` for the seller
  and its line is `CategoryCode E` with `VATEX-FR-FRANCHISE`. Your validator answers **valid** on it, 0
  failure out of 93 + 65 checks + the XSD;
- negative control on the same file, seller `SpecifiedTaxRegistration` removed and nothing else:
  **invalid**, `[BR-E-02]` - which is the report of #560 reproduced;
- the setup page still builds its 30 items and renders on 23.0.3 and 18.0.10, without that row.

daGrumpf-bxp:

The case of the issue, exchanged for real through the SuperPDP sandbox rather than validated locally: a
**seller set "Non assujetti à la TVA"** emits a new invoice to a customer on another instance.

```
emitter  dd23 23.0.3, mysoc Burger Queen, FACTURE_TVAOPTION=0 -> tva_assuj=0
  invoice BQ-FA-2608-0110, one service line at 0%, flow i_214991
  document: <ram:SpecifiedTaxRegistration><ram:ID schemeID="FC">000000002</ram:ID>   <- BT-32, what this branch adds
            breakdown CategoryCode E, ExemptionReasonCode VATEX-FR-FRANCHISE
Access Point   : acknowledgement Ok - accepted and routed (this is what BR-E-02 refused before)
receiver dolidev 23.0.3 -> supplier invoice 246, ref_supplier BQ-FA-2608-0110, ht 100.00, line vat 0%
statuses back on the emitter : 200 (ie_733663), 201 (ie_733664)
```

The seller setting was restored to "assujetti" after the run.

eldy:
Can you help me to solve conflict ?

daGrumpf-bxp:
Done — `main` merged in, conflict resolved, CI green, **MERGEABLE/CLEAN**.

The conflict was only about where new code lands, not about what it does: this branch and #575 (BT-73/BT-74,
merged since) each appended a function at the end of `einvoicing/lib/einvoicing.lib.php`, so git saw one
block replacing another. Both are kept, side by side:

| | added by |
|---|---|
| `einvoicingSellerVatRegime()` / `einvoicingSellerTaxRegistrations()` | this branch |
| `einvoicingInvoicingPeriodFromLines()` | #575 |

The rest merged on its own, and the two places where the two changes meet in the same function are worth a
look since git decided them alone - I checked both by hand:

- `CIIProtocol::buildLineNode()`: this branch writes `ram:ExemptionReason` / `ram:ExemptionReasonCode`
  inside the line `ram:ApplicableTradeTax`, #575 replaced the inline line-period block with a call to the
  shared `buildBillingPeriodNode()`. Result keeps both, in the D22B sequence order.
- `buildinvoicelines.inc.php`: `sellerTaxRegistations` (this branch) and `invoicingPeriodStart` /
  `invoicingPeriodEnd` (#575) both feed `$invoiceData`, from two independent derivations.

Re-checked on the merge commit rather than on what each branch proved separately:

- the module test suite, file by file, on 18.0.10 and 23.0.3 - **11 files green on both**, including the
  `SellerVatRegimeTest` of this branch and the `InvoicingPeriodTest` and `SupplierInvoiceHelperTest` that
  came with main;
- one full sandbox exchange, new invoice, seller "Non assujetti à la TVA": `BQ-FA-2608-0111` carries
  `<ram:ID schemeID="FC">000000002</ram:ID>` and `VATEX-FR-FRANCHISE`, the Access Point acknowledges `Ok`,
  the invoice is received on the instance in front, and 200 / 201 / **205** come back on the customer
  invoice - the 205 being sent by the receiver on validation, which is the behaviour merged yesterday.
  Nothing left in an open transaction on either side.

PR #566, opened 2026-08-09
FIX: the setup sections lose their title from Dolibarr 20 (generateOutput signature threshold)

daGrumpf-bxp:
Found while auditing every `DOL_VERSION` branch of the module against the core sources of 17 to 24, rather than against the versions that happen to be installed.

## What is wrong

`admin/setup.php` renders each section of the Access point setup through a small wrapper:

```php
if ((float) DOL_VERSION >= 20) {
    // Native multi-arg signature available since Dolibarr 20.0.0
    return $formSetupInstance->generateOutput(true, false, $title, 'titlefieldmiddle');
}
```

That comment is not right. `FormSetup::generateOutput()` grew its arguments in two steps, not one:

| Dolibarr | signature |
|---|---|
| 17 / 18 / 19 | `$editMode` |
| 20 / 21 / 22 | `$editMode, $hideTitle` |
| **23 / 24** | `$editMode, $hideTitle, `**`$title`**`, `**`$cssfirstcolumn`** |

PHP discards the surplus arguments of a user function without a word, so from 20.0 on the page asked for a title the core never looked at, and printed the default "Parameter / Value" header instead — the opposite of what the wrapper exists to do. Nothing failed, which is why it went unnoticed.

## What this does

The threshold moves to 23, where the signature actually is, and the rewrite that supplies the title covers 18 to 22.

That rewrite needed one change to reach 22. It required the second header cell to hold the "Value" label:

```php
'#<tr class="liste_titre">\s*<td>Parameter</td>\s*<td>Value</td>\s*</tr>#'
```

and 22.0 renders that cell **empty**, so the substitution silently missed on the very version where the native call had just stopped working. It is matched on its shape now.

## Tests

Measured on the bench, **one instance per supported version, 17 through 24** — 17.0.4, 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and 24.0.0-beta — not deduced from the sources alone.

The probe reproduces both branches of the wrapper against a real `FormSetup`, reads the signature back through reflection, and reports whether the produced HTML carries the title:

| Dolibarr | `generateOutput()` seen by reflection | native 4-arg call | rewrite, before | rewrite, after |
|---|---|---|---|---|
| 17.0.4 | 1 arg: `$editMode` | title lost | title present | title present |
| 18.0.10 | 1 arg: `$editMode` | title lost | title present | title present |
| 19.0.4 | 1 arg: `$editMode` | title lost | title present | title present |
| 20.0.4 | 2 args: `$editMode, $hideTitle` | title lost | title present | title present |
| 21.0.4 | 2 args: `$editMode, $hideTitle` | title lost | title present | title present |
| 22.0.5 | 2 args: `$editMode, $hideTitle` | title lost | **title lost** | title present |
| 23.0.3 | 4 args: `…, $title, $cssfirstcolumn` | **title present** | title present | title present |
| 24.0.0-beta | 4 args: `…, $title, $cssfirstcolumn` | **title present** | title present | title present |

The reflection column is the point: the four-argument form the page was calling only exists from 23, and the two versions where it does are exactly the two where the native call works. Below that the title was dropped in silence — and 22 was worse still, since the rewrite that was supposed to cover the gap missed there too.

After this change every version produces the title: 17 to 22 through the rewrite, 23 and 24 through the native call, which is untouched.

Unit suite green on all eight instances, 9 test files each, one PHP process per file. `phpcs` clean.


daGrumpf-bxp:

This branch only changes how the setup page calls `FormSetup::generateOutput()`, so the exchange chain is
what needed proving unchanged. One **new** invoice through the SuperPDP sandbox:

```
emitter  dolidev 23.0.3 -> TRI-FA-2608-0132, flow i_214941
Access Point   : acknowledgement Ok
receiver dd23 23.0.3 -> supplier invoice 1034, ref_supplier TRI-FA-2608-0132, ht 100.00
statuses back on the emitter : 200 (ie_733489), 201 (ie_733490)
```

And the point of the branch itself, measured again on the eight versions (the probe reproduces both
branches of the wrapper on the real core):

| Dolibarr | `generateOutput()` signature | native 4-argument call | the rewrite used below 23 |
|---|---|---|---|
| 17.0.4 / 18.0.10 / 19.0.4 | `($editMode)` | title **lost** | title present |
| 20.0.4 / 21.0.4 / 22.0.5 | `($editMode, $hideTitle)` | title **lost** | title present |
| 23.0.3 / 24.0.0-beta | `($editMode, $hideTitle, $title, $cssfirstcolumn)` | title present | title present |

The setup page also builds and renders its whole option list on 18.0.10 and 23.0.3 (29 and 30 items).

PR #574, opened 2026-08-11
FIX: EINVOICING_PROFILE is ignored in Factur-X generation — profile a…

LuckyRoussel98:
…lways forced to EXTENDED

## Problem

When the exchange protocol is `FACTURX`, the generated PDF always embeds a Factur-X XML with the **EXTENDED** profile (guideline `urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended`, XMP `fx:ConformanceLevel = EXTENDED`), even when the hidden constant `EINVOICING_PROFILE` is set to `EN16931`.

## Root cause

In `class/protocols/FacturXProtocol.class.php` (around line 311), the profile switch has two defects:

1. the `case 'EN16931':` branch assigns `ZugferdProfiles::PROFILE_EXTENDED` instead of `ZugferdProfiles::PROFILE_EN16931`;
2. the branch has no `break` (there is even a `// no break` marker), so execution falls through into `default:`, which re-assigns EXTENDED and calls `ZugferdDocumentBuilder::createNew()` a **second time**.

Net effect: whatever the configured value, `$used_profile` is always EXTENDED, and when `EINVOICING_PROFILE=EN16931` the document builder is instantiated twice. The syslog line just above ("create new XML document based on PROFILE_EN16931 (CIUS-FR)") makes debugging harder because it does not reflect what is actually produced.

## Why it matters

- EN 16931 is the profile of the European standard — the level every Access Point / Plateforme Agréée and every receiver must support. EXTENDED is a superset intended for complex use cases; claiming EXTENDED for ordinary invoice data only reduces interoperability (stricter/different schematron on the receiving side, extended fields the receiver is not required to process), with no benefit since the generator does not write any EXTENDED-only field.
- Inconsistency between protocols: the CII protocol already produces EN 16931 (`urn:cen.eu:en16931:2017`), so the same invoice currently gets a different conformance level depending on the transport format.

## Fix

Assign the correct constant in each branch, add the missing `break`s, and move the `createNew()` call after the switch so the builder is created exactly once. The default behaviour (EXTENDED when `EINVOICING_PROFILE` is not set) is preserved.

## How to test

1. Set `EINVOICING_PROTOCOL=FACTURX` and `EINVOICING_PROFILE=EN16931` in `llx_const`.
2. Generate an invoice PDF and extract the embedded `factur-x.xml` from `<ref>_facturx.pdf`.
3. Before the fix, the guideline reads `urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended`; after the fix it reads `urn:cen.eu:en16931:2017`, and the XMP `fx:ConformanceLevel` switches from `EXTENDED` to `EN 16931`.
4. Without `EINVOICING_PROFILE` set, the output remains EXTENDED (default unchanged).

The defect is version-independent (the switch is executed the same way on every Dolibarr version supported by the module, including 24). Verified on a real invoice: with the fix, the embedded XML validates with the EN 16931 guideline and a single builder instantiation.

### DIFF
--- a/einvoicing/class/protocols/FacturXProtocol.class.php
+++ b/einvoicing/class/protocols/FacturXProtocol.class.php
@@ -310,13 +310,13 @@ class FacturXProtocol extends CIIProtocol
 			$profile = getDolGlobalString('EINVOICING_PROFILE');
 			switch ($profile) {
 				case 'EN16931':
-					$used_profile = ZugferdProfiles::PROFILE_EXTENDED;
-					$facturxpdf = ZugferdDocumentBuilder::createNew($used_profile);
-					// no break
+					$used_profile = ZugferdProfiles::PROFILE_EN16931;
+					break;
 				default:
 					$used_profile = ZugferdProfiles::PROFILE_EXTENDED;
-					$facturxpdf = ZugferdDocumentBuilder::createNew($used_profile);
+					break;
 			}
+			$facturxpdf = ZugferdDocumentBuilder::createNew($used_profile);
 			dol_syslog(get_class($this) . '::executeHooks create new XML document based on ' . $used_profile);
 
 			// Get the type of invoice in FacturX nomenclature


daGrumpf-bxp:
Reviewed and run on the bench (Dolibarr 17.0.4, 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3,
24.0.0-beta, one instance per version on the same clone). The switch itself is wrong exactly as you
describe, and your rewrite is the right shape: the `EN16931` case assigned `PROFILE_EXTENDED`, the
`// no break` fell through into `default:`, and the builder was instantiated twice. Verified that the
constant you now assign produces what it should, straight from the vendored horstoeko on five versions:

| constant | `GuidelineSpecifiedDocumentContextParameter/ID` |
|---|---|
| `PROFILE_EN16931` | `urn:cen.eu:en16931:2017` |
| `PROFILE_EXTENDED` | `urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended` |

Two things I could not reproduce, and I think they change what this PR claims rather than whether it
should be merged.

**1. That branch does not run on a default install.** `FacturXProtocol::generateXML()` starts with:

```php
if (!getDolGlobalInt('EINVOICING_USE_EXTERNAL_FACTURX_BUILDER')) {
    return parent::generateXML($invoice, $outputlangs);
} else {
    // deprecated, use the native $CII->buildXML method instead !
```

Everything below - including the profile switch - is the deprecated horstoeko builder. Unless
`EINVOICING_USE_EXTERNAL_FACTURX_BUILDER` is set, generation goes to the native CII writer, and that
writer does not read `EINVOICING_PROFILE` at all: its knob is **`EINVOICING_XML_PROFILE`**
(`CIIProtocol::getBuildXmlProfile()`, accepted values `MINIMUM`, `BASICWL`, `BASIC`, `EN16931`,
`EXTENDED`, `EXTENDEDFR`, default `FacturXProtocol::BUILD_XML_PROFILE` = `EXTENDED`).

Measured on a real customer invoice, protocol `FACTURX`, with this PR deployed - `EINVOICING_PROFILE`
makes no difference because the branch is never entered:

| instance | `EINVOICING_PROFILE` | guideline of the embedded XML |
|---|---|---|
| 18.0.10 | unset | `...#conformant#urn:factur-x.eu:1p0:extended` |
| 18.0.10 | `EN16931` | `...#conformant#urn:factur-x.eu:1p0:extended` |
| 23.0.3 | unset | `...#conformant#urn:factur-x.eu:1p0:extended` |
| 23.0.3 | `EN16931` | `...#conformant#urn:factur-x.eu:1p0:extended` |
| 24.0.0-beta | unset / `EN16931` | `...:extended`, XMP `EXTENDED` |

So the observation "the generated PDF always embeds EXTENDED even when `EINVOICING_PROFILE=EN16931`" is
right, but the cause on a default install is that the constant is not the one that path reads. Setting
`EINVOICING_XML_PROFILE=EN16931` is what changes the profile today, and it does so on the CII protocol
too.

**2. In the mode where the branch *is* reached, generation dies before producing anything.** With
`EINVOICING_USE_EXTERNAL_FACTURX_BUILDER=1` on an ordinary invoice (one product line, HT price base,
no discount):

```
PHP Warning:  Undefined array key "grosspriceamount" in .../FacturXProtocol.class.php on line 464
PHP Fatal error:  Uncaught TypeError: ZugferdDocumentBuilder::setDocumentPositionGrossPrice():
Argument #1 ($amount) must be of type float, null given, called in .../FacturXProtocol.class.php on line 464
```

`buildinvoicelines.inc.php` only sets `grosspriceamount` when a unit price **including tax** is known
(`if ($line_unit_price_ttc)`), which is the rare case its own comment describes, so the key is simply
absent for a normal line. Reproduced on 18.0.10 and 23.0.3, and identical on `main` without this PR -
**pre-existing, not something you introduced**. On 24.0.0-beta the same run ends with no Factur-X file
produced.

Which means the "How to test" recipe cannot have been run in the external-builder mode, and in the
default mode the profile does not move. Worth aligning the description with one of the two, so whoever
reads this later is not sent looking for a behaviour change that the code cannot produce yet.

None of that is an argument against merging: the switch is broken, this is the correct fix, and it is
strictly better than what is there. Three small things if you want to go further:

- a `ChangeLog.md` entry under `## 1.0.4` - the repo documents user-visible fixes there;
- `MINIMUM`, `BASICWL`, `BASIC` and `EXTENDEDFR` still land in `default:` and come out EXTENDED, so the
  same defect remains for four of the six profiles the module names. Reading `EINVOICING_XML_PROFILE`
  and mapping it through `ZugferdProfiles` would make the two writers answer to one setting;
- and the `dol_syslog()` line above the switch still announces `PROFILE_EN16931 (CIUS-FR)`
  unconditionally, which is the misleading log your description mentions - it is one line away.

daGrumpf-bxp:
For completeness, a full sandbox cycle with this branch deployed, to be sure the exchange chain is untouched by it: a new invoice emitted from dolidev (23.0.3) — `TRI-FA-2608-0136`, flow `i_214998` — accepted and routed by the Access Point, received on the instance in front as supplier invoice 1036, statuses 200 (`ie_733683`) and 201 (`ie_733684`) back on the emitter, no transaction left open. As expected: the branch only rewrites a switch in the deprecated external-builder path, so nothing in the default chain moves.

PR #575, opened 2026-08-11
FIX #572 : the invoicing period of the document (BT-73 / BT-74) is never written

### LINKED ISSUE #572: BT-73 and BT-74 not written to xml document
### Bug

The BT-73 and BT-74 fields (Invoicing period start / end dates ) are omitted from the e-invoice generation.

### Module

einvoicing

### Dolibarr Version

23.0.2

### Environment PHP

_No response_

### Environment Database

_No response_

I'm just doing some grep in the codebase, but my understanding:

In the file `vendor/horstoeko/zugferd/src/ZugferdDocumentBuilder.php: ` , there is a function to set the BT-73 and BT-74 fields

```
/**
 * Get detailed information on the billing period
 *
 * @param  DateTimeInterface|null $startDate   __BT-73, From BASIC WL__ Start of the billing period
 * @param  DateTimeInterface|null $endDate     __BT-74, From BASIC WL__ End of the billing period
 * @param  string|null            $description __BT-X-264, From EXTENDED__ Further information of the billing period (Obsolete)
 * @return ZugferdDocumentBuilder
 */
public function setDocumentBillingPeriod(?DateTimeInterface $startDate, ?DateTimeInterface $endDate, ?string $description): ZugferdDocumentBuilder
{
    $period = $this->getObjectHelper()->getSpecifiedPeriodType($startDate, $endDate, null, $description);

    $this->getObjectHelper()->tryCall($this->headerTradeSettlement, "setBillingSpecifiedPeriod", $period);

    return $this;
}
```

The function is never actually called, except in the sample e invoice with hard-coded dates.

```
 grep -ri setDocumentBillingPeriod class

class/protocols/FacturXProtocol.class.php:              $documentBuilder->setDocumentBillingPeriod(DateTime::createFromFormat("Ymd", "20250101"), DateTime::createFromFormat("Ymd", "20250131"), "01.01.2025 - 31.01.2025");
```

### Attached files

_No response_

daGrumpf-bxp:
Closes #572, with the option @eldy chose: BT-73/BT-74 are derived from the periods of the lines, because
"a lot of software are not as accurate as Dolibarr and will manage only the period on invoice level".

## What was wrong

BG-14 was never written, by either path, although the data was already collected and then thrown away.
`buildinvoicelines.inc.php` accumulated every line period into `$billing_period` and nothing ever read it
back, while the header array hard-coded the two fields:

```php
'invoicingPeriodStart' => null,
'invoicingPeriodEnd'   => null,
```

So an invoice stated a period on its lines (BT-134/BT-135, emitted since #573) and denied having one in
its header. On the Factur-X side `setDocumentBillingPeriod()` was only ever called from
`generateSampleInvoiceOld()`, with hard-coded dates, which is what @Drinausaur found by grep.

## What this does

The document covers what its lines cover: **the earliest start date and the latest end date** of the
lines that carry one. Dolibarr has no invoicing period at invoice header level, so there is nothing else
to read it from, and EN 16931 does not say how to derive BG-14 from BG-26 — this is the decision, taken
in one function so it is testable and so both writers share it:

```php
einvoicingInvoicingPeriodFromLines(['start' => [1 => ts, 2 => ts], 'end' => [...]]) // -> ['start' => ts, 'end' => ts]
```

**Nothing is invented for a side that is missing.** BR-CO-19 asks for the start date *or* the end date,
so a line billed from a date with no end gives a header period with BT-73 alone, and no empty BT-74.

**A period that would start after it ends is not emitted at all.** One line open from March next to
another closed in January derives `start > end`, which BR-29 refuses — and a whole document rejected
because of a header nobody filled in would be worse than not deriving one. The lines keep their own
periods in that case, which is exactly what happens today. Verified rather than assumed: taking the
document produced here and swapping the two header dates, and nothing else, the reference validator
answers `[BR-29]` and `Invalide`.

**Where the node goes is not decoration.** `HeaderTradeSettlementType` is a sequence, so
`ram:BillingSpecifiedPeriod` sits after every `ram:ApplicableTradeTax` and before
`ram:SpecifiedTradeAllowanceCharge` / `ram:SpecifiedTradePaymentTerms`. Read off the schemas this module
vendors, which are the FNFE-MPE ones:

| profile | declares `BillingSpecifiedPeriod` under the header settlement |
|---|---|
| MINIMUM | no — the type holds `InvoiceCurrencyCode` and the monetary summation, nothing else |
| BASIC WL, BASIC, EN16931, EXTENDED, EXTENDED-CTC-FR | yes, between `ApplicableTradeTax` and `SpecifiedTradeAllowanceCharge` |

The enclosing condition already excluded MINIMUM, so nothing special was needed for it, and BASIC WL —
which carries no lines — now carries the period the lines would have carried.

The line-level writer was rewritten to call the same `buildBillingPeriodNode()`, so the two levels
cannot drift apart. Same node, same `format="102"`, one place to fix.

## Tests

`InvoicingPeriodTest` (new, 6 tests) pins the derivation: the period covering several lines, a single
line, one side alone on each side, no line with a date, the contradictory pair, and the one-day period
that must not be mistaken for one. `CIIProfileShapeTest` gets 5 more (15 tests, 294 assertions) for the
generated XML: the node present on every profile but MINIMUM, its two dates and their format, its
position in the sequence, XSD validation of every profile *with* a period, one side alone, and no empty
group when the invoice has no period at all.

Run file by file on the eight versions the module supports, one instance per version on the same clone:

| | 17.0.4 | 18.0.10 | 19.0.4 | 20.0.4 | 21.0.4 | 22.0.5 | 23.0.3 | 24.0.0-beta |
|---|---|---|---|---|---|---|---|---|
| `InvoicingPeriodTest` | OK | OK | OK | OK | OK | OK | OK | OK |
| `CIIProfileShapeTest` | OK | OK | OK | OK | OK | OK | OK | OK |
| the other 8 test files | | OK | | | | | OK | |

On real invoices — two service lines with different periods, validated and generated:

| instance | lines | header BG-14 | validator |
|---|---|---|---|
| 23.0.3 | 06-01→06-30, 05-15→07-31 | BT-73 `20260515`, BT-74 `20260731` | valid, 0 failure / 200 checks |
| 18.0.10 | same | same | valid, 0 / 203 |
| 24.0.0-beta | same | same | valid, 0 / 200 |

And the three other shapes on 23.0.3, each generated for real and validated: a start alone → BT-73 only
(valid, 0/164); the contradictory pair → no header period, lines keep theirs (valid, 0/191); no period
anywhere → nothing emitted (valid, 0/158). Plus the negative control above, which is the same document
with its header dates swapped and nothing else: `Invalide`, single failure `[BR-29]` — so the validator
does bite on exactly what the guard prevents.

Also run, since generation is shared code: the sample matrix (CII standard / down payment / credit note
on 18, 23 and 24 — all OK; the Factur-X samples fail below 24 on the FPDF/TCPDF collision of #554, which
is unchanged by this branch and reproduced identically on `main`), and an import of one of the generated
documents as a supplier invoice, read back from a second process to be sure nothing was left in an open
transaction.

## One adjacent gap, deliberately left alone

The import path does not carry BT-134/BT-135 into the supplier invoice lines: `updateline()` is passed
`$supplierInvoice->lines[$i]->date_start`, which the parser never fills from `linePeriodStart`, so the
imported lines land with `date_start`/`date_end` NULL. Measured on the import above. That is a reading
gap and this issue is about writing, so it is not touched here — happy to open a separate issue for it.


daGrumpf-bxp:
Adjacent gap noted at the end of the description is now #576 — the import side dropping BT-134/BT-135. Independent of this branch, which stays about writing the header period.

daGrumpf-bxp:

Real exchange through the SuperPDP sandbox, one **new** invoice with two service lines billed over
different periods:

```
emitter  dolidev 23.0.3  ->  TRI-FA-2608-0131, flow i_214933
  line 289  2026-06-01 .. 2026-06-30
  line 290  2026-05-15 .. 2026-07-31
  header BG-14 : BT-73=20260515 BT-74=20260731      <- what this branch adds
Access Point   : acknowledgement Ok, the document is accepted and routed
receiver dd23 23.0.3  ->  supplier invoice 1033, ref_supplier TRI-FA-2608-0131, ht 200.00
statuses back on the emitter : 200 (flow ie_733460), 201 (flow ie_733461)
```

So the derived header period does not make the platform refuse the document, and the whole chain -
generation, transmission, routing, reception, statuses - behaves as before with it present.

For the record, the received lines land without their own period on this branch (`date_start=NULL`): that
is the reading gap of #576, which is a separate branch, and this run is what makes the two visibly
independent.

PR #582, opened 2026-08-13
FIX #581 : the instance hash in the XML comment depends on the button that generated the document

### LINKED ISSUE #581: differences in CII generation depending on the trigger
### Bug

Using current main branch (ae41c5ba2716f96218d0351d687d760413c1edb5)

I noticed that when I generate the CII e-invoice using the « Generate » button in the attached files of a bill, The cii headear contains the instance public hash.
When I generate the CII e-invoice using the « Facture Électronique -> Régénérer » button, this information is not embedded:

```
$ diff -u FA2607-0167_*
--- FA2607-0167_2_cii.xml       2026-08-13 08:28:05.968461305 +0200
+++ FA2607-0167_cii.xml 2026-08-13 08:27:54.324864687 +0200
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <rsm:CrossIndustryInvoice xmlns:rsm="urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100" xmlns:qdt="urn:un:unece:uncefact:data:standard:QualifiedDataType:100" xmlns:ram="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100" xmlns:xs="http://www.w3.org/2001/XMLSchema" xmlns:udt="urn:un:unece:uncefact:data:standard:UnqualifiedDataType:100" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
-  <!--Einvoice XML generated by Dolibarr 23.0.3 - 2026-08-13T06:27:56Z-->
+  <!--Einvoice XML generated by Dolibarr 23.0.3 - 2026-08-13T06:27:46Z - Instance public hash 2a6fd2182eccf4511883591232b615f95488ce55b681c9b3ea3a43e6ff6f2302-->
   <rsm:ExchangedDocumentContext>
     <ram:BusinessProcessSpecifiedDocumentContextParameter>
       <ram:ID>S1</ram:ID>
```

It seems due to the fact that when using the « Facture Électronique -> Régénérer » button, the `getHashUniqueIdOfRegistration` function from blockedlog module seems not exposed.

I don’t know if it’s hide some issue. This should not be a problem regarding the generated invoice since it’s a comment in the XML, but I’m not qualified to asses that.

Thank’s!

### Module

einvoicing

### Dolibarr Version

23.0.3

### Environment PHP

8.4

### Environment Database

postgresql

Please see description

### Attached files

_No response_

daGrumpf-bxp:
Closes #581.

## What @casta saw

The same invoice, generated twice, gives two different XML headers:

```
-  <!--Einvoice XML generated by Dolibarr 23.0.3 - ...-->
+  <!--Einvoice XML generated by Dolibarr 23.0.3 - ... - Instance public hash 2a6fd218...-->
```

The hash is there when the document is produced by the **Generate** button of the attached files, and
gone when it is produced by **Facture Électronique → Régénérer**. Their diagnosis was right, and the
core says why.

## Why

`CIIProtocol::buildXML()` guards the stamp with `function_exists('getHashUniqueIdOfRegistration')`, and
never loads the file that declares it. That function lives in `blockedlog/lib/blockedlog.lib.php`, and
in a Dolibarr 23 checkout the only place that pulls the library into an invoice request is the PDF
builder:

```php
// htdocs/core/lib/pdf.lib.php:785, pdfWriteBlockedLogSignature()
include_once DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php';
```

`main.inc.php:3786` includes it too, but only on `index.php` or with `&forceregistration=1`.

So the **Generate** button goes through `generateDocument()` → PDF model → `pdf.lib.php` → library
loaded → `function_exists()` true → hash written. **Régénérer** (`action=generate_einvoice`) calls
`$protocol->generateInvoice()` directly, which never touches the PDF builder → `function_exists()`
false → no hash. Nothing was ever version- or setup-dependent: it is purely which include chain the
request happened to walk.

## The fix

Load the library where the comment is written, then keep the existing guard.

Both guards stay on purpose, and no `DOL_VERSION` test replaces them: the library file already ships
with Dolibarr 17 to 22, it is the *function* that appeared in 23, so `function_exists()` is the only
test that means what it says. Below 23 nothing is stamped, exactly as today. Nothing at include time in
that library executes anything — it declares functions only, in every version checked.

## Measured, on the eight versions

Same invoice on each instance, generated through the **Régénérer** path
(`$protocol->generateInvoice()`, the code the button runs):

| Dolibarr | before | after |
|---|---|---|
| 17.0.4 | no hash | no hash |
| 18.0.10 | no hash | no hash |
| 19.0.4 | no hash | no hash |
| 20.0.4 | no hash | no hash |
| 21.0.4 | no hash | no hash |
| 22.0.5 | no hash | no hash |
| 23.0.3 | **no hash** | **hash** |
| 24.0.0-beta | **no hash** | **hash** |

Control on the same eight, through the PDF path (`generateDocument()`): unchanged, and now identical to
what the regenerate path produces on the same instance.

`EInvoicingSamplesTest`, `CIIProtocolTest` and `CIIProfileShapeTest` pass on all eight (3 + 5 + 15 tests).
`phpcs` clean with the repository ruleset.

## Note for @casta

Your instinct that it should not matter is right — it is a comment, no rule of EN 16931 or of the
CTC-FR schematron reads it, and no receiving software does. What it did hide is the asymmetry itself:
two code paths building the same document out of two different include states. That is worth removing
even when the visible symptom is cosmetic.

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -1899,6 +1899,14 @@ class CIIProtocol extends AbstractProtocol
 		$doc->appendChild($root);
 
 		// Add comment
+		// getHashUniqueIdOfRegistration() lives in the blockedlog library, which only the paths going through
+		// the PDF builder happen to include (pdf.lib.php). Load it here as well, otherwise the very same invoice
+		// is stamped with the instance hash or not depending on the button that triggered the generation.
+		// The library ships with older versions too, but the function itself only appeared in Dolibarr 23,
+		// so both guards stay: nothing is stamped below 23, exactly as before.
+		if (!function_exists('getHashUniqueIdOfRegistration') && file_exists(DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php')) {
+			include_once DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php';
+		}
 		$hash_unique_id = '';
 		if (function_exists('getHashUniqueIdOfRegistration')) {
 			$algo = 'sha256';
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -1902,9 +1902,9 @@ class CIIProtocol extends AbstractProtocol
 		// getHashUniqueIdOfRegistration() lives in the blockedlog library, which only the paths going through
 		// the PDF builder happen to include (pdf.lib.php). Load it here as well, otherwise the very same invoice
 		// is stamped with the instance hash or not depending on the button that triggered the generation.
-		// The library ships with older versions too, but the function itself only appeared in Dolibarr 23,
-		// so both guards stay: nothing is stamped below 23, exactly as before.
-		if (!function_exists('getHashUniqueIdOfRegistration') && file_exists(DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php')) {
+		// The library ships with every supported version, the function itself only from Dolibarr 23, so the
+		// function_exists() below still decides: nothing is stamped under 23, exactly as before.
+		if (!function_exists('getHashUniqueIdOfRegistration')) {
 			include_once DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php';
 		}
 		$hash_unique_id = '';
--- a/einvoicing/ChangeLog.md
+++ b/einvoicing/ChangeLog.md
@@ -267,6 +267,14 @@ checks both on any update and reports only the last of the two, which made the c
 behind the vendor error. Marking a thirdparty as a vendor also stores its new code now, which passing
 the code alone never did: update() writes the code columns only when it is allowed to modify them.
 
+FIX: The same invoice no longer produces two different documents depending on the button that generated
+it. The comment opening the XML names the instance it was produced on, and that name comes from
+getHashUniqueIdOfRegistration(), a function of the blockedlog library that only the paths going through
+the PDF builder happen to load - so generating from the attached files carried the hash, while
+regenerating from the e-invoicing menu, which calls the writer directly, silently dropped it. The
+library is now loaded where the comment is written. Nothing changes below Dolibarr 23, where that
+function does not exist yet, and nothing changes on the PDF path (issue #581).
+
 ## 1.0.3
 
 FIX: The totals of the generated document now follow the rounding convention of the instance. Dolibarr


daGrumpf-bxp:
End-to-end cycle run on the sandbox with this branch, and a control cycle run on `main` right after, for
the record:

| | branch | flow | acknowledgement | received on the other instance |
|---|---|---|---|---|
| this PR | `75c9a15` | `i_322069` | `Pending` | no |
| control | `main` `ae41c5b` | `i_322518` | `Pending` | no |

Emission works on both (`code=15`, right recipient, document written), and the document produced by the
cycle carries the hash as intended:

```
<!--Einvoice XML generated by Dolibarr 23.0.3 - ... - Instance public hash 846d21b711596d3620111715f5584a1f631d93a638dc7361068bfdf3d078e9e5-->
```

The routing acknowledgement stayed `Pending` on both flows and neither was delivered, so the receiving
half of the cycle could not be exercised today. Since `main` behaves identically, it is a sandbox-side
condition and not something this branch introduces — which is also what the change makes structurally
impossible: it only decides whether an XML comment carries the instance hash.

daGrumpf-bxp:
Put back to draft: the end-to-end acceptance is not done, and I would rather say so than let the
earlier note read as a pass.

The sandbox has stopped processing submitted flows this morning. Everything sent today is stuck, while
what was sent yesterday went through in a second:

| flow | submitted | acknowledgement | flowProfile |
|---|---|---|---|
| `i_218995` | 12 Aug 09:19 | `Ok` (+1.4 s) | `CIUS` |
| `i_322069` | 13 Aug 09:51 | `Pending` | `Undefined` |
| `i_322518` | 13 Aug 09:56 | `Pending` | `Undefined` |
| `i_323547` | 13 Aug 10:08 | `Pending` | `Undefined` |

`updatedAt` still equals `submittedAt` on all three, so the platform never even classified them. It is
not the document: a full diff of the document that went through against the one that is stuck shows
only the reference, the dates and the generation timestamp - same guideline `urn:cen.eu:en16931:2017`,
same business process `S1`, same routing address `0225:315143296_2342`. A control cycle run on `main`
behaves identically, so it is not this branch either.

Everything that does not depend on the platform is done and reported above. I will run a complete cycle
and mark this ready as soon as the sandbox routes again.

daGrumpf-bxp:
Follow-up on why the cycle could not be completed, now settled from the platform's own interface rather
than inferred: **the sandbox stopped analysing submitted documents this morning.** Nothing on this
branch, nor in the documents it produces, is involved.

The platform's event log tells it plainly:

| flow | events |
|---|---|
| `i_218995` (12 Aug, this exact document) | `Téléversée` 11:19:50.103 → **`Déposée (validée)` 200** 11:19:51.471 → `Émise` 201 → `Approuvée` 205 |
| `i_211026` (8 Aug, a document it did not like) | `Téléversée` 11:54:26.630 → **`Rejetée` 213** 11:54:27.437 |
| `i_322069`, `i_322518`, `i_323547`, `i_324266` (13 Aug) | `Téléversée` — **and nothing after** |

The analyser normally answers in under a second, whether it accepts (200) or rejects (213). Today it
does not answer at all: the four flows carry no invoice number and no totals in the invoice list, their
detail page is empty of seller, buyer and lines, and the download button returns
`{"http_status_code":400,"message":"invoice has not been analyzed yet"}`. That is also what the API
metadata was saying with `flowProfile: Undefined` and an `updatedAt` still equal to `submittedAt`.

Two controls rule out this side:

- **The document that the platform accepted yesterday, resubmitted today**, byte for byte apart from
  its invoice number — which had to change, the platform keying duplicates on BT-1 and not on the
  trackingId — comes back `Undefined` / `Pending` like the rest (`i_324266`). Same file, same declared
  `flowProfile: CIUS`, opposite outcome, one day apart.
- **SuperPDP's own validator** (`api.superpdp.tech/v1.beta/validation_reports`) reports the four
  documents valid with **zero failures**, at conformance level `urn:cen.eu:en16931:2017`. Negative
  control on the same document, with `ram:CalculatedAmount` moved from 40.00 to 40.06 and nothing else:
  invalid, `[BR-CO-14]`. The tool bites, so the four passes mean something.

This branch stays in draft until a complete cycle runs. Everything that does not depend on the platform
is done and reported above.

einvoicing/class/protocols/CIIProtocol.class.php
```diff
@@ -1899,6 +1899,14 @@ public function buildXML(array $invoiceData, array $linesData, $profile = '')
 		$doc->appendChild($root);
 
 		// Add comment
+		// getHashUniqueIdOfRegistration() lives in the blockedlog library, which only the paths going through
+		// the PDF builder happen to include (pdf.lib.php). Load it here as well, otherwise the very same invoice
+		// is stamped with the instance hash or not depending on the button that triggered the generation.
+		// The library ships with older versions too, but the function itself only appeared in Dolibarr 23,
+		// so both guards stay: nothing is stamped below 23, exactly as before.
+		if (!function_exists('getHashUniqueIdOfRegistration') && file_exists(DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php')) {
```
eldy: Do we need the "file_exists(DOL_DOCUMENT_ROOT.'/blockedlog/lib/blockedlog.lib.php')" ?

daGrumpf-bxp: Good catch, removed - it was defensive for no reason.

Checked rather than assumed, on the eight versions the module supports: `blockedlog/lib/blockedlog.lib.php` is present in all of them, 17.0.4 through 24.0.0-beta. Only the function is recent, and `function_exists()` already carries that. The core itself includes the same file with no guard of its own, in `pdf.lib.php` - which is the very path this change mirrors.

Re-measured through the regenerate path after the change: identical to before, nothing stamped on 17 to 22, hash on 23 and 24.

daGrumpf-bxp:
The sandbox came back around 11:00 UTC and the cycle now completes end to end on this branch.

| step | result |
|---|---|
| emission (dolidev, Dolibarr 23.0.3) | `TRI-FA-2608-0145`, flow `i_325651`, `code=15` |
| platform | `Téléversée` → **`Déposée (validée)` 200** → **`Émise par la plateforme` 201** in 1.4 s, `flowProfile: CIUS`, acknowledgement `Ok` |
| reception on the other instance | supplier invoice 1153, `ref_supplier=TRI-FA-2608-0145`, 200.00 net / 240.00 gross, both line periods preserved |
| statuses back to the issuer | 200 and 201, attached to the invoice |

Taking it out of draft. Full state of the acceptance for this change:

- the eight supported versions measured through the regenerate path, before and after: nothing stamped on 17 to 22, hash on 23 and 24, unchanged on the PDF path;
- `EInvoicingSamplesTest`, `CIIProtocolTest`, `CIIProfileShapeTest` green on the eight (3 + 11 + 15 tests);
- the four documents produced today validated by the SuperPDP validator, zero failures, with a negative control (`BR-CO-14`) proving the tool bites;
- `phpcs` clean, and the `file_exists` guard dropped as you asked.

daGrumpf-bxp:
The red phan check on this pull request is not from this branch: it is the
`PhanTypeArraySuspiciousNullable` on `actions_einvoicing.class.php:1387`, which `main` itself carries
since f5157454 and which every open pull request inherits through its merge ref. This branch touches
`CIIProtocol.class.php` and the ChangeLog only.

#584 fixes it, measured before and after on the eight supported versions. This check should go green
here on its own once that lands.

PR #583, opened 2026-08-13
FIX #548 : approving a received invoice locks the statuses that come after it

### LINKED ISSUE #548: [einvoicing] Option to send payment status (211) when is fully paid
### Bug

When a supplier invoice is validated and a full payment is recorded, all e-invoice status actions disappear. As a result, it is no longer possible to send status 211.

Since the invoice is fully paid at that stage, status 211 could be sent automatically when the invoice reaches the paid status. This would keep the e-invoice lifecycle consistent without requiring a manual action that is no longer available.

Alternatively, the status action could remain available manually after the invoice has been paid, allowing users to decide when to send it.

Note: reopening the invoice does not restore the e-invoice status buttons

### Module

einvoicing

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

set a supplier invoice fully paid

### Attached files

_No response_

daGrumpf-bxp:
Closes #548.

@camlafit reported that once a supplier invoice is validated and fully paid, every e-invoice status
action is gone, so the "Payment transmitted" (211) can no longer be sent. The automatic 211 and the
manual one both landed on `main` earlier, but the report also pointed at something that was not fixed,
and that is what this PR is about.

## Why the actions disappear

The `einvoice` button group of the supplier invoice card was hidden as soon as an **Approved (205)** or
a **Refused (210)** status had been accepted by the platform, on the reading that either of them closes
the lifecycle.

Only the refusal does. An invoice sent back to its vendor is not going to be paid, so nothing more is
owed on it. An approved invoice, on the contrary, is precisely the one that gets paid — and the 211 is
what reports that payment. Since the normal order of things is *approve, then pay*, the button group
was already gone by the time anyone would want the 211. That also explains the last remark of the
report: re-opening the invoice does not bring the buttons back, because the condition never looked at
the Dolibarr status of the invoice at all, only at what had already been sent.

## What changes

The card now offers what the exchange still allows, instead of closing it on the first final status:

- a status the platform accepted is never proposed a second time;
- a **refusal** leaves nothing — the group disappears, as before;
- an **approval** only takes the refusal away with it, since an approved invoice cannot then be refused.

The decision moved out of the hook into `EInvoicing::getSendableStatusesForReceivedInvoice()`, so it can
be tested, and `hasSentStatusMessage()` gained an `$onlyAccepted` flag to express "already accepted by
the platform" rather than "already attempted" — a send the platform rejected has to stay retryable.

One thing found on the way: the inline query being replaced compared `direction = 'out'` and
`lc_validation_status = 'Ok'` against the case actually stored, which is `Out` and `Ok`. MySQL lets that
pass on a case-insensitive collation; PostgreSQL matches nothing, so on Postgres the gate never fired at
all. Both comparisons are lowercased now, like the direction test already in `hasSentStatusMessage()`.

## Measured

On real invoices of the bench, comparing the old gate with the new list on the same records:

| instance | invoice | already accepted | before | after |
|---|---|---|---|---|
| dd23 | SI2608-0052 | 205 | *no button* | **211** |
| dd23 | SI2608-0050 | 205 | *no button* | **211** |
| dd23 | SI2608-0049 | 205 | *no button* | **211** |
| dd24 | SI2606-0001 | 205 | *no button* | **211** |
| dd18 | (PROV479) | 205 | *no button* | **211** |
| dd23 | SI2608-0006 | 205, 211 | *no button* | *no button* |
| dd18 | SI2608-0003 | 205, 211 | *no button* | *no button* |
| dd24 | SI2607-0002 | 211, 210, 205 | *no button* | *no button* |

The 211 thus made reachable was then really sent, by hand, on `SI2608-0052` (Dolibarr 23, SuperPDP
sandbox): `sendStatusMessage(211) -> res:1`, flow `ie_915345`, `validation=Ok` — accepted by the
platform. The card closes afterwards, the invoice having nothing left to send.

New test `test/phpunit/ReceivedInvoiceStatusesTest.php`, 5 cases including "a rejected send stays
retryable": **OK on the eight supported versions** (17.0.4, 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5,
23.0.3, 24.0.0-beta). `phpcs` clean with the repository ruleset.

A full sandbox cycle was run on this branch: emission fine (`code=15`, flow `i_323547`), but the routing
acknowledgement stayed `Pending` and the document was not delivered — a control cycle run on `main`
today behaves identically, so it is the sandbox and not this branch. Nothing here touches emission or
routing in any case.

The customer side is untouched, and was checked rather than assumed: recording a payment on a
transmitted customer invoice still sends the "Cashed in" (212) through the `PAYMENT_CUSTOMER_CREATE`
trigger — flow `ie_916003`, `validation=Ok`.

### DIFF
--- a/einvoicing/ChangeLog.md
+++ b/einvoicing/ChangeLog.md
@@ -267,6 +267,18 @@ checks both on any update and reports only the last of the two, which made the c
 behind the vendor error. Marking a thirdparty as a vendor also stores its new code now, which passing
 the code alone never did: update() writes the code columns only when it is allowed to modify them.
 
+FIX: Approving a received invoice no longer takes away the statuses that come after it. The einvoice
+button group of the supplier invoice card disappeared as soon as an "Approved" (205) or a "Refused"
+(210) status had been accepted by the platform, on the assumption that either of them closes the
+lifecycle. Only the refusal does - an invoice sent back to its vendor is not going to be paid - while
+an approved one is, and "Payment transmitted" (211) is what reports it. Since the normal order of
+things is to approve an invoice and then pay it, the manual 211 was already unreachable by the time
+anyone would want it, and re-opening the invoice did not bring it back: the condition never looked at
+the Dolibarr status of the invoice, only at what had been sent. The card now offers what the exchange
+still allows - a status the platform accepted is not proposed a second time, a refusal leaves nothing,
+and an approval only takes the refusal away with it. The query it replaces also compared the direction
+and the validation status against their stored case, which matches nothing on PostgreSQL (issue #548).
+
 ## 1.0.3
 
 FIX: The totals of the generated document now follow the rounding convention of the instance. Dolibarr
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -445,40 +445,30 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 			$resql = $db->query($sql);
 			if ($resql && $db->num_rows($resql) > 0) {
 				$db->free($resql);
-				// Check if a final status (approved or rejected) has already been sent and validated
-				// → in this case, the lifecycle is complete, so we hide the button
-				$sqlFinal = "SELECT rowid FROM " . $db->prefix() . "einvoicing_lifecycle_msg";
-				$sqlFinal .= " WHERE element_id = " . (int) $object->id;
-				$sqlFinal .= " AND element_type = '" . $db->escape($object->element) . "'";
-				$sqlFinal .= " AND direction = 'out'";
-				$sqlFinal .= " AND lc_status IN (" . (int) EInvoicing::STATUS_APPROVED . ", " . (int) EInvoicing::STATUS_REFUSED . ")";
-				$sqlFinal .= " AND lc_validation_status = 'Ok'";
-				$sqlFinal .= " LIMIT 1";
-				$resqlFinal = $db->query($sqlFinal);
-				$hasFinalLifecycle = ($resqlFinal && $db->num_rows($resqlFinal) > 0);
-				$db->free($resqlFinal);
+				// Offer what the exchange still allows, rather than closing it on the first final status:
+				// the button group used to disappear as soon as an "Approved" (205) was accepted, while
+				// the payment - and the "Payment transmitted" (211) that reports it - necessarily comes
+				// after the approval (issue #548).
+				$availableStatuses = $einvoicing->getSendableStatusesForReceivedInvoice($object->id, $object->element);
 
-				if (!$hasFinalLifecycle) {
-					$availableStatuses = $einvoicing->getEinvoiceStatusOptions(1, 1, 1);
-					$url_button = array();
-					foreach ($availableStatuses as $code => $label) {
-						$url_button[] = array(
-							'lang' => 'einvoicing',
-							'enabled' => true,
-							'perm' => ($forcedisabling ? -1 : ((bool) $user->hasRight("fournisseur", "facture", "creer") && empty($forcedisabling))),
-							'label' => (string) $label,
-							'url' => '/fourn/facture/card.php?id=' . $object->id . '&action=sendStatusMessage&pdpstatuscode=' . $code . '&token=' . newToken()
-						);
-					}
+				$url_button = array();
+				foreach ($availableStatuses as $code => $label) {
+					$url_button[] = array(
+						'lang' => 'einvoicing',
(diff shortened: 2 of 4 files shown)


daGrumpf-bxp:
Put back to draft: the end-to-end acceptance is not done, and I would rather say so than let the
earlier note read as a pass.

The sandbox has stopped processing submitted flows this morning. Everything sent today is stuck, while
what was sent yesterday went through in a second:

| flow | submitted | acknowledgement | flowProfile |
|---|---|---|---|
| `i_218995` | 12 Aug 09:19 | `Ok` (+1.4 s) | `CIUS` |
| `i_322069` | 13 Aug 09:51 | `Pending` | `Undefined` |
| `i_322518` | 13 Aug 09:56 | `Pending` | `Undefined` |
| `i_323547` | 13 Aug 10:08 | `Pending` | `Undefined` |

`updatedAt` still equals `submittedAt` on all three, so the platform never even classified them. It is
not the document: a full diff of the document that went through against the one that is stuck shows
only the reference, the dates and the generation timestamp - same guideline `urn:cen.eu:en16931:2017`,
same business process `S1`, same routing address `0225:315143296_2342`. A control cycle run on `main`
behaves identically, so it is not this branch either.

Everything that does not depend on the platform is done and reported above. I will run a complete cycle
and mark this ready as soon as the sandbox routes again.

daGrumpf-bxp:
Follow-up on why the cycle could not be completed, now settled from the platform's own interface rather
than inferred: **the sandbox stopped analysing submitted documents this morning.** Nothing on this
branch, nor in the documents it produces, is involved.

The platform's event log tells it plainly:

| flow | events |
|---|---|
| `i_218995` (12 Aug, this exact document) | `Téléversée` 11:19:50.103 → **`Déposée (validée)` 200** 11:19:51.471 → `Émise` 201 → `Approuvée` 205 |
| `i_211026` (8 Aug, a document it did not like) | `Téléversée` 11:54:26.630 → **`Rejetée` 213** 11:54:27.437 |
| `i_322069`, `i_322518`, `i_323547`, `i_324266` (13 Aug) | `Téléversée` — **and nothing after** |

The analyser normally answers in under a second, whether it accepts (200) or rejects (213). Today it
does not answer at all: the four flows carry no invoice number and no totals in the invoice list, their
detail page is empty of seller, buyer and lines, and the download button returns
`{"http_status_code":400,"message":"invoice has not been analyzed yet"}`. That is also what the API
metadata was saying with `flowProfile: Undefined` and an `updatedAt` still equal to `submittedAt`.

Two controls rule out this side:

- **The document that the platform accepted yesterday, resubmitted today**, byte for byte apart from
  its invoice number — which had to change, the platform keying duplicates on BT-1 and not on the
  trackingId — comes back `Undefined` / `Pending` like the rest (`i_324266`). Same file, same declared
  `flowProfile: CIUS`, opposite outcome, one day apart.
- **SuperPDP's own validator** (`api.superpdp.tech/v1.beta/validation_reports`) reports the four
  documents valid with **zero failures**, at conformance level `urn:cen.eu:en16931:2017`. Negative
  control on the same document, with `ram:CalculatedAmount` moved from 40.00 to 40.06 and nothing else:
  invalid, `[BR-CO-14]`. The tool bites, so the four passes mean something.

This branch stays in draft until a complete cycle runs. Everything that does not depend on the platform
is done and reported above.

daGrumpf-bxp:
The sandbox came back around 11:00 UTC and the cycle now completes end to end on this branch, in a
single pass:

| step | result |
|---|---|
| emission (dolidev, Dolibarr 23.0.3) | `TRI-FA-2608-0146`, flow `i_325843`, `code=15` |
| platform | `Déposée (validée)` 200 → `Émise par la plateforme` 201 |
| reception on the other instance | supplier invoice 1154, `ref_supplier=TRI-FA-2608-0146`, 200.00 net / 240.00 gross, both line periods preserved, on the very first synchronisation |
| statuses back to the issuer | 200 and 201 |

Taking it out of draft. Full state of the acceptance for this change:

- the gate measured on real records of three instances (Dolibarr 18, 23 and 24): invoices sitting at
  "approved" get their 211 back, invoices already approved and paid keep no button;
- the 211 thus made reachable really sent by hand and accepted by the platform (flow `ie_915345`,
  `validation=Ok`);
- `ReceivedInvoiceStatusesTest`, 5 cases, green on the eight supported versions;
- the customer side checked rather than assumed: recording a payment still sends the "Cashed in" (212)
  through the `PAYMENT_CUSTOMER_CREATE` trigger (flow `ie_916003`, `validation=Ok`);
- `phpcs` and `phan` clean.

PR #587, opened 2026-08-13
fix #586 : Use authorization Header

### LINKED ISSUE #586: [Einvoicing] Prerfer Authorization header to retrieve Token
### Feature Request

Hi,

During connection tests through the PDPLibre proxy, we observed a difference in the way the token is retrieved. Using the **Authorization** header appears to be more compliant than passing the token as a request argument.

It would be preferable for getToken() to use the Authorization header first, rather than a request parameter. 
In this case behavior should be identical between all PA

### Module

--Please select a module--

### Use case

_No response_

### Suggested implementation

_No response_

camlafit:
* Esalink provide two mode to retrieve token
* Authorization header looks more compliant and safer between all PA usecase

### DIFF
--- a/einvoicing/class/providers/EsalinkPDPProvider.class.php
+++ b/einvoicing/class/providers/EsalinkPDPProvider.class.php
@@ -283,15 +283,25 @@ class EsalinkPDPProvider extends AbstractPDPProvider
 	{
 		global $langs;
 
+		$extraHeaders = [];
+
 		// OAuth2 client_credentials — RFC 6749, application/x-www-form-urlencoded
 		// Replace old POST /v1/token (JSON username/password) disabled since v1.2.6 (2026-07).
-		// Prefers Authorization to be more compliant between PA usage
-		$param = http_build_query(array(
-			'client_secret' => $this->config['password'],
-		));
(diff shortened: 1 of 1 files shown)


einvoicing/class/providers/EsalinkPDPProvider.class.php
```diff
@@ -285,13 +285,13 @@ public function getAccessToken()
 
 		// OAuth2 client_credentials — RFC 6749, application/x-www-form-urlencoded
 		// Replace old POST /v1/token (JSON username/password) disabled since v1.2.6 (2026-07).
+		// Prefers Authorization to be more compliant between PA usage
```
eldy: Because 2 modes are possible and this section is only for Esalink, to not lose the other method that may be needed in future, can you introduce an option
```
if getDolGlobalString('ESALINK_AUTHENT_USING_CLIENT_CREDENTIAL') {
// old code using grant_type = client_credentials
} else {
// method using Basic
}
```

camlafit: it's done, thanks

Issue #132, opened 2026-04-17
[EInvoicing] Improve customer invoice sending


atm-maxime:
Add an option to generate and send automatically an invoice when validated
Add a mass action to generate and send several invoices at once

eldy:
> Add an option to generate and send automatically an invoice when validated
There is already option to generate e-invoice automatically when validation is done.
Still need: 
[] Option to "Send automatically"
[] Mass action to generate and send e-invoice    (for the case of the einvoice is already generated, we just send)

lelex86:
Hello,

While investigating how to implement both features, I identified a blocking issue
related to **FactureX generation**: there is a known incompatibility between TCPDF
(used by Dolibarr for base PDF generation) and horstoeko/zugferd library used for
FactureX embedding. This is tracked in issue #218 .

So, here is what I think we can do.

---

### Feature 1 — Automatic sending on invoice validation

The problem with triggering generation + sending at validation time is timing:

- If we trigger **`BILL_VALIDATE`**: TCPDF is not yet loaded, but the final PDF
  does not exist yet and FactureX requires it
- If we hook into **`afterPDFCreation`**: the final PDF exists, but TCPDF is still
  in memory, causing the horstoeko/zugferd conflict

**Proposal:** implement the automatic sending option for **CII protocol only**.
When the user enables "automatic sending on validation", the protocol is forced to CII
with a clear warning explaining that FactureX is not compatible with this option.

This avoids the TCPDF-horstoeko/zugferd conflict entirely, and CII (pure XML, no PDF dependency)
works cleanly from the `BILL_VALIDATE` trigger.

---

### Feature 2 — Mass actions on customer invoice list

The same constraint applies: generating FactureX on-the-fly during a mass action
would trigger the same incompatibility.

**Proposal:** 
- Limit the sending mass action to invoices that **already have a generated
document** (i.e. `STATUS_GENERATED`) and an eInvoicing file is really present on disk. These invoices
have already passed the generation step and only need to be submitted to the PDP —
no PDF or FactureX generation required at action time.

Invoices in `STATUS_NOT_GENERATED` or with a missing eInvoicing would be skipped with a
clear notice in the results report.

- Limit the mass action to generate Invoices documents to CII protocol only

---

It's clearly not perfect, but that's all that comes to my mind. What do you think? Any other idea?

daGrumpf-bxp:
[...]

PR #590, opened 2026-08-14
FIX Make the Factur-X files valid PDF/A-3

daGrumpf-bxp:
## The problem

A Factur-X file is a PDF/A-3 file carrying the XML. **veraPDF 1.30.2 rejects every file this module produces**, on Dolibarr 17 to 24 alike, so none of them is a conformant Factur-X — however clean its XML is, and the XML is clean (EN 16931, Factur-X EXTENDED and the CTC-FR schematron all pass, 0 failure).

That gap is easy to miss because the validators usually reached for — the SuperPDP one, the FNFE schematron — only look at the XML. They say "valid" on a file no PDF/A validator accepts.

### Cause 1 — an unparsable XMP packet, below v24 only (ours)

`buildFacturxXmpRdf()` copied the extension schema of the horstoeko/zugferd template as a `rdf:Description` of its own, next to the one TCPDF had already written for the pdf, xmpMM and pdfaid schemas: two descriptions repeating `pdfaExtension:schemas` on the same `rdf:about=""` subject, which the XMP specification forbids. The packet stops parsing there, and a reader that gives up sees no `pdfaid` either — so the file is not even *identified* as PDF/A (clauses 6.6.2.1 and 6.6.4). The v24 path, where horstoeko/zugferd writes the whole packet, never had this one.

### Cause 2 — no embedded fonts (ISO 19005-3, 6.2.11.4.1), both merge paths

At the default `PDF_USE_A = 0` the core draws the invoice with the Standard 14 fonts, which are not embedded. Merging can only carry those pages over, it cannot repair them. This one is a setting of Dolibarr, not of the module.

## The fix

1. **The Factur-X schema entry is grafted into the bag TCPDF writes**, from `_out()`, instead of being a description of its own, with the `/Length` of the metadata stream brought back in line. That is cause 1, entirely inside the module.
2. **Cause 2 is reported, not decided.** Following @eldy's review: the module never writes `PDF_USE_A`, in memory or otherwise — a module that rewrites the PDF before or after us finds the request exactly as it would have been without us. When the format is Factur-X and the PDF format of Dolibarr is not PDF/A-3b, the user is told, on the setup page of the module (with a link to *Home - Setup - PDF*) and on the invoice at generation, that the recommended format is CII and that keeping Factur-X means setting *PDF documents format* to PDF/A-3b for the whole Dolibarr. It is a warning and not a refusal to generate, so that an upgrade does not stop the installations that generate Factur-X today; say so and it becomes an error.
3. **The specimen** of the setup page obeys the same rule, where it used to force the setting around its own `generateDocument()`: what a user opens to check the module is what the setup of the instance actually produces.

## Verified, on the eight cores 17 → 24

| | PDF format left at its default | PDF format set to PDF/A-3b |
|---|---|---|
| veraPDF `-f 3b` on a generated invoice, before this PR | 0/8 conformant | 0/8 conformant |
| veraPDF `-f 3b` on a generated invoice, after | 8/8 fail on clause 6.2.11.4.1 **alone** (the fonts of the carrier — nothing the module writes) | **8/8 conformant** |
| the 24 specimens (8 cores × 3 invoice types) | — | **24/24 conformant** |
| the warning | shown, on the setup page and on the invoice | not shown |

The defect was reproduced on the unpatched code on all eight instances first, then the patch applied and the whole batch replayed:

- generated **XML identical to the byte** before and after, on all eight (timestamp comment filtered) - this patch touches the container and nothing else;
- **PHPUnit** green on the eight cores, file by file (13 files x 8);
- the 24 specimens of the setup page conformant, with the PDF format set;
- **a full sandbox cycle emitting Factur-X** (not CII), replayed on this exact code: invoice `TRI-FA-2608-0152` routed by the Access Point (flow `i_332851`, `http=200`, acknowledgement `Ok`), received on the other instance with its two line periods intact, statuses 200 and 201 back at the sender - and the document that actually went over the wire passes veraPDF (`PASS 3b`) and the SuperPDP validator (216 checks, 0 failure, `EXTENDED`);
- the warning shown and not shown where it should be, on both sides of the setting, on every core.
- **eight more cycles in CII**, since the check lives in `checkRequiredinformations()`, which both protocols share - each one emitted, routed (`acknowledgement: Ok`), received with its header period (BT-73/BT-74) and its two line periods, and answered 200 + 201 back at the sender, with no transaction left open and no warning anywhere:

| | emitter | receiver | invoice | flow |
|---|---|---|---|---|
| 23 → 23 | dolidev 23.0.3 | dd23 23.0.3 | `TRI-FA-2608-0153` | `i_332897` |
| 23 → 23, other way round | dd23 23.0.3 | dolidev 23.0.3 | `BQ-FA-2608-0116` | `i_332902` |
| **17 → 23** | dd17 **17.0.4** | dolidev 23.0.3 | `D17-FA-2608-0008` | `i_332905` |
| **23 → 17** | dolidev 23.0.3 | dd17 **17.0.4** | `TRI-FA-2608-0154` | `i_332907` |
| **18 → 23** | dd18 **18.0.10** | dolidev 23.0.3 | `FA2608-0053` | `i_332911` |
| **23 → 18** | dolidev 23.0.3 | dd18 **18.0.10** | `TRI-FA-2608-0155` | `i_332913` |
| **24 → 23** | dd24 **24.0.0-beta** | dolidev 23.0.3 | `D24-FA-2608-0001` | `i_332924` |
| **23 → 24** | dolidev 23.0.3 | dd24 **24.0.0-beta** | `TRI-FA-2608-0156` | `i_332929` |

  The emitted XML of the first one is valid on the SuperPDP validator (203 checks, 0 failure). The cross-version pairs are the point: 17 is the minimum the module claims and 24 the newest, and in those rows the version that emits and the version that receives are not the same one.

The tooling is worth keeping around: `veraPDF -f 3b` is what catches this class of defect, and no XML validator will. #592 puts it under CI.


daGrumpf-bxp:
Follow-up filed as #591: nothing in CI checks the container half of the format, so the two defects this PR fixes could come back unnoticed. The issue carries a workflow sketch that runs both mergers from the repository alone — no Dolibarr instance — and, run against the code before this PR, fails on exactly the three XMP clauses fixed here.

einvoicing/class/actions_einvoicing.class.php
```diff
@@ -70,6 +70,62 @@ public function messageOfTheDay($parameters, $object, &$action, $hookmanager)
...
+			return 0;
+		}
+
+		$einvoicing = new EInvoicing($db);
+		if ($einvoicing->checkModulePrerequisites() < 0 || !$einvoicing->needEInvoiceManagement($invoiceObject)) {
+			return 0;
+		}
+
+		$conf->global->PDF_USE_A = 3;
+		dol_syslog(__METHOD__ . ' PDF_USE_A raised to 3 for ' . $invoiceObject->ref . ': a Factur-X carrier has to be a PDF/A', LOG_DEBUG, 0, '_einvoicing');
```
eldy: This may generate troubles with other modules that will modify on the fly the pdf before or after.

I recommand instead to just:
 Check that PDF_USE_A has been set the 3. And if not, report an error to say to the end user that:
* Recommanded format for generated einvoice is CII, not Factur-X
* But if user still really want to use FacturX, he must modify globally manually its Dolibarr setup in "setup - pdf" to force the use PDF/A-3b globally so all its Dolibarr environent will use this format.

daGrumpf-bxp:
@eldy Done, in the push above — the module no longer touches the setting of the core, it reads it and reports.

`beforePDFCreation` is gone, and so is `EINVOICING_DISABLE_PDFA_CARRIER`, which only existed to opt out of it. Nothing raises `PDF_USE_A`, in memory or anywhere else, so a module that rewrites the PDF before or after us finds the request exactly as it would have been without us. The specimen of the setup page follows the same rule now (it used to force the setting around its own `generateDocument()`): what a user opens to check the module is what the setup of the instance actually produces, which is the point of a specimen.

What replaces it is the check you describe. When the format is Factur-X and `PDF_USE_A` is not 3, the user is told — on the setup page of the module, with a link to *Home - Setup - PDF*, and on the invoice at generation, through `checkRequiredinformations()` where the other configuration checks already live — that the recommended format for a generated e-invoice is CII, not Factur-X, and that keeping Factur-X means setting *PDF documents format* to PDF/A-3b for the whole Dolibarr. One reservation on which I followed the softer reading, tell me if you want the other: it is reported as a warning, not as a refusal to generate. Blocking would stop, on upgrade, every installation that generates Factur-X today, on a setting nobody has ever been asked to set. Say the word and it becomes an error.

What is left of the PR is the half that is ours: below v24 the merger wrote the Factur-X extension schema as a second `rdf:Description` repeating `pdfaExtension:schemas` on the same `rdf:about=""` subject, which the XMP specification forbids. The packet stops being parsed there and the file is not even *identified* as PDF/A any more (veraPDF: 6.6.2.1 t4/t5 and 6.6.4). The entry is grafted into the bag TCPDF already writes instead.

Replayed on the eight cores, 17 to 24, after the change:

| | PDF format left at its default | PDF format set to PDF/A-3b |
|---|---|---|
| veraPDF on a generated invoice | 8/8 fail, on clause 6.2.11.4.1 **alone** — the fonts of the carrier, nothing the module writes | **8/8 conformant** |
| the warning | shown, on the setup page and on the invoice | not shown |

Plus the 24 specimens (8 cores x 3 invoice types) conformant, 13 PHPUnit files green on the eight cores, and generation still returning 1 everywhere in both states.

---

One word on the format itself, since it is the recommendation you make in the message and in #132. We did not go and make the whole outgoing Factur-X path work, on eight versions, for the pleasure of it: **we already have users who ask for Factur-X**, in emission and in reception, and a PDF their customers and their accountants can read with the XML inside is what those users expect. That CII is the safer format is not in dispute and we say so to them. But as long as the module offers Factur-X, it has to work for the people who have to use it — which is why the effort went there.

Issue #591, opened 2026-08-14
[einvoicing] Nothing in CI checks that the Factur-X files are valid PDF/A-3


daGrumpf-bxp:
## Why

A Factur-X file is a PDF/A-3 file carrying the XML. Everything the project currently validates — the FNFE schematron, the SuperPDP validator, the EN 16931 XSLT — looks **only at the XML**. All of them say "valid" on a file that no PDF/A validator accepts, so the whole container half of the format has no guard at all.

That is not hypothetical: #590 fixes two defects that had been shipping for a while and that every XML validator called clean, on Dolibarr 17 to 24 alike. Nothing in CI would have flagged them, and nothing in CI stops them coming back.

`veraPDF` is the reference implementation for PDF/A and exits non-zero on failure, so it plugs into a workflow as-is.

## What can actually run in CI

I built and ran the job locally before opening this, so the recipe below is not a suggestion — it produces the files and the verdicts quoted.

Both mergers run **standalone, with no Dolibarr instance**:

- `CtcFrPdfMerger` (the v24 path) needs nothing but the committed `einvoicing/vendor/`.
- `FacturxTcpdfMerger` (below v24) needs a TCPDF and nothing else of the core — a plain `tecnickcom/tcpdf` release works.

Both take the carrier already shipped in the repo, `einvoicing/doc/00_ZugferdDocumentPdfBuilder_PrintLayout.pdf`, and the three fixtures in `einvoicing/test/phpunit/fixtures/einvoicing_samples/`. Six documents, no database, no install.

Two small things the harness needs: a three-line `dol_syslog()` stub, since both mergers log through it, and the TCPDF download for the second one.

Results here, veraPDF 1.30.2, `-f 3b`:

| | verdict |
|---|---|
| the 3 fixtures × `CtcFrPdfMerger` | 3/3 **PASS** |
| the 3 fixtures × `FacturxTcpdfMerger` (with #590) | 3/3 **PASS** |
| the same job on `FacturxTcpdfMerger` **before** #590 | **FAIL** — 6.6.2.1 t4, 6.6.2.1 t5, 6.6.4 t1 |

The last line is the point: the job reproduces the defect #590 fixes, from the repository alone. It is a real regression guard, not a smoke test.

## Sketch

```yaml
name: pdfa
on:
  pull_request:
    branches: [ main ]
  push:
    branches: [ main ]
jobs:
  pdfa:
    name: Factur-X container is PDF/A-3
[...]

PR #592, opened 2026-08-14
NEW PHPUnit test that the Factur-X container is a valid PDF/A-3

daGrumpf-bxp:
Related to #591, which #595 closes.
**Stacked on #590** — merge that one first. Until then this PR carries its commit too; the diff of its own is the single file below.

**The CI job has been taken out of this PR**, as @eldy asked: it is #595, which is stacked on this one. This PR is only the test.

## What it does

A Factur-X file is a PDF/A-3 file carrying the XML, and everything else this project validates looks at the XML only: the container had no guard at all, which is how two defects that made every generated file invalid PDF/A shipped unnoticed on Dolibarr 17 to 24 alike.

`einvoicing/test/phpunit/FacturxPdfaContainerTest.php` builds one Factur-X document per XML fixture, with each of the two mergers, and reads the container back:

- a single, parsable XMP packet identifying the file as PDF/A-3B;
- exactly one `pdfaExtension:schemas` bag, with the Factur-X extension schema declared inside it;
- a metadata stream whose declared `/Length` lands on the end of the packet;
- the XML embedded as an associated file, under the name `factur-x.xml`, with the `/Data` relationship and a document level `/AF`;
- the PDF/A output intent, and the `fx:` properties the extension schema is what makes legal.

6 tests, 168 assertions, one file.

## It needs no Dolibarr, and it uses one when there is one

Both mergers want a logger and nothing else of the core, and the carrier PDF (`einvoicing/doc/00_ZugferdDocumentPdfBuilder_PrintLayout.pdf`) and the three XML fixtures are already committed. So the test takes the instance through `DOLIBARR_HTDOCS` or the `htdocs/custom` symlink the way the other tests of that directory do, and stands in for `dol_syslog()` when there is none — which is what lets a CI with no instance run the very same file.

TCPDF — which the core ships and this repository does not — is found through the `TCPDF_PATH` the core defines, and through `TCPDF_DIR` otherwise. `FACTURX_PDFA_OUTDIR` keeps the generated documents in a directory of your choice instead of a temporary one, so a PDF/A validator can be pointed at them afterwards.

## It is a regression guard, not a smoke test

Run against `FacturxTcpdfMerger` **as it stands on `main` today**, it fails on exactly the defect #590 fixes, while the other merger still passes:

```
✘ Facturx tcpdf merger builds a facturx container with data set "cii_standard"
  tcpdf_cii_standard.pdf carries 2 PDF/A extension schema declarations instead of one
Tests: 6, Assertions: 114, Failures: 3.
```

and veraPDF, on the documents that run left behind, agrees for the same reason:

```
PASS ctcfr_cii_creditnote.pdf 3b
PASS ctcfr_cii_deposit.pdf 3b
PASS ctcfr_cii_standard.pdf 3b
FAIL tcpdf_cii_creditnote.pdf 3b     Duplicate property or field node 'pdfaExtension:schemas'
FAIL tcpdf_cii_deposit.pdf 3b
FAIL tcpdf_cii_standard.pdf 3b
```

With #590 applied: `OK (6 tests, 168 assertions)`, and 6/6 `PASS` on veraPDF.

Measured **on the eight cores 17 to 24**, one instance per major version: `OK (6 tests, 168 assertions)` on each, with no skip — TCPDF is found on all of them. And on the repository alone, no Dolibarr, with a TCPDF release: same result.

## One thing worth knowing, because I got it wrong first

The first form of this PR ran the two mergers in two separate processes, on the grounds that they cannot coexist. They can. What cannot coexist with the FPDF the horstoeko/zugferd writer descends from is the `class FPDF extends TCPDF` shim of `htdocs/includes/tcpdi/tcpdi.php` — not TCPDF itself, which `FacturxTcpdfMerger` reaches through the namespaced FPDI flavour and which needs no `FPDF` at all. The test therefore exercises the zugferd merger first and loads TCPDF only afterwards, and it skips the first one if it ever finds the shim already in place, which is exactly the case the second merger exists for.

## What it does not cover

Said plainly so the check is not read as more than it is. The **font** half of #590 comes from the invoice PDF the *core* produces (`PDF_USE_A`), and there is no Dolibarr here to produce one; that half is a setting of the instance, which #590 reports and does not decide. What is covered is everything the module itself writes into the container — the XMP packet, the extension schema, the `/AF` array, the output intent, the attachment and its relationship — which is where both of the module's own defects lived.


eldy:
Because this test is specific to einvoice module, can you first add it into a phpunit script inside einvoice/test/phpunit dir ?

daGrumpf-bxp:
@eldy Done — it is a PHPUnit test now: `einvoicing/test/phpunit/FacturxPdfaContainerTest.php`. The script under `dev/tools/` is gone; the workflow runs the test.

The test builds one Factur-X document per XML fixture, **with each of the two mergers**, and reads the container back: a single parsable XMP packet identifying the file as PDF/A-3B, exactly one `pdfaExtension:schemas` bag with Factur-X declared inside it, a metadata stream whose declared `/Length` lands on the end of the packet, the XML embedded as an associated file with the `/Data` relationship, and the PDF/A output intent. 6 tests, 168 assertions.

Two things I had to settle to get it into `test/phpunit`, both worth knowing:

- **It runs on an instance and without one.** It takes the Dolibarr instance the way the other tests of that directory do (`DOLIBARR_HTDOCS`, or the `htdocs/custom` symlink), and when there is none it stands in for the only thing of the core the mergers need — `dol_syslog()`. That is what lets the same file be the CI check, which has no Dolibarr, and a test a developer runs on his instance like the twelve others.
- **Both mergers in one process, which I claimed was impossible.** It is the `class FPDF extends TCPDF` shim of `htdocs/includes/tcpdi/tcpdi.php` that cannot coexist with the FPDF the horstoeko/zugferd writer descends from — not TCPDF itself, which `FacturxTcpdfMerger` reaches through the namespaced FPDI flavour and which needs no `FPDF` at all. So the test exercises the zugferd merger first, then loads TCPDF and exercises the other one, and it skips the first if it ever finds the shim in place — which is exactly the case the second merger exists for.

TCPDF is found through `TCPDF_PATH` when there is an instance, and through a `TCPDF_DIR` the CI sets to a plain release download otherwise. `FACTURX_PDFA_OUTDIR` keeps the documents where a validator can be pointed at them; that is how the workflow hands the very same six files to veraPDF afterwards, for the clauses PHP alone cannot settle. The job fails on a skipped test on purpose: a merger the test could not build must not pass for a green run.

Measured:

- **on the eight cores 17 to 24**, one instance per major version: `OK (6 tests, 168 assertions)` on each, no skip — TCPDF is found on all of them;
- **on the repository alone**, no Dolibarr, with a TCPDF release: same result, and veraPDF then says `PASS 3b` on the six documents;
- **negative control**, the merger as it stands on `main` today: 3 failures, `carries 2 PDF/A extension schema declarations instead of one`, and veraPDF `FAIL` on the same three files with `Duplicate property or field node 'pdfaExtension:schemas'`. The other merger still passes, before and after.

Still happy to pin the veraPDF version, drop the artifact upload, or fold the job into an existing workflow.

.github/workflows/pdfa.yml
```diff
@@ -0,0 +1,99 @@
+---
```
eldy: Can you remove this workflow ci from the PR.  CI is related to tools outside of code (other ci could be used for same code) so should be into a separate PR.

daGrumpf-bxp:
@eldy Done — the workflow is out of this PR. It is now **#595**, stacked on this one, and this PR is only `einvoicing/test/phpunit/FacturxPdfaContainerTest.php`.

Split that way, #591 is closed by #595 (the issue is about the CI) and this PR just references it. Nothing else changed: the test is the same file, 6 tests / 168 assertions, still green on the eight cores 17 to 24 with no skip, and still failing on exactly the defect #590 fixes when run against the merger as it stands on `main`.

Merge order is #590 → #592 → #595.

.github/workflows/pdfa.yml
daGrumpf-bxp: Removed — the workflow now lives in its own PR, #595, stacked on this one. This PR is only the PHPUnit test.

PR #595, opened 2026-08-15
NEW CI job running the Factur-X container test and veraPDF

### LINKED ISSUE #591: [einvoicing] Nothing in CI checks that the Factur-X files are valid PDF/A-3

daGrumpf-bxp:
Closes #591.

Split out of #592 at @eldy's request: the CI belongs in its own PR. **This PR is one file and nothing else**, `.github/workflows/pdfa.yml`, on top of `main`.

Now that #590 and #592 are merged, the job is green on this PR — no stacking needed anymore.

## What it does

Runs `einvoicing/test/phpunit/FacturxPdfaContainerTest.php` (#592), then hands **the very documents that test produced** to veraPDF, the reference implementation for PDF/A. The test settles everything PHP can see in the container; veraPDF settles the rest.

Without it, nothing in this repository looks at the container half of a Factur-X file — which is how the two defects #590 fixes shipped unnoticed on Dolibarr 17 to 24 alike, while every XML validator kept saying "valid".

## It needs no Dolibarr and no database

The mergers only want a logger, the carrier PDF and the three XML fixtures are already committed, and TCPDF — which the core ships and this repository does not — is a plain release download. A few seconds of PHP on top of two downloads.

Two choices worth explaining:

- **`--fail-on-skipped`.** The test skips a merger it could not build (no TCPDF, or a global class `FPDF` already claimed). A silent skip would leave half the coverage out of a green run, so a skip is a failure here while it stays a skip for a developer running the file on his instance.
- **`FACTURX_PDFA_OUTDIR`.** The test writes to a temporary directory by default; the job points it at `out/` so the validation step and the artifact see the same six files the assertions were made on — not files rebuilt separately, which could differ.

## Cost

Measured on the runs this workflow has already produced on this branch, against the jobs of `main` over the same period (`ubuntu-latest`, job duration as GitHub reports it, `started_at` → `completed_at`):

| job | runs | min | median | max |
|---|---|---|---|---|
| **pdfa (this PR)** | 8 | 23 s | **30 s** | 42 s |
| phan | 20 | 22 s | 28 s | 48 s |
| phpcs | 20 | 30 s | 38 s | 56 s |
| assign-and-label | 20 | 5 s | 8 s | 13 s |

**The new job is shorter than the longest job already running.** The four jobs are independent and start together, so the time to "all checks green" on a PR is still set by `phpcs`, and does not move. What is added is one more parallel runner for half a minute — 1 billable minute per run at GitHub's per-job rounding, and public repositories are not billed for `ubuntu-latest` at all.

Where those 30 seconds go, median over 5 runs:

| step | time | |
|---|---|---|
| checkout | 2 s | |
| setup PHP 8.3 + phpunit | 5 s | |
| fetch TCPDF 6.7.5 | 1 s | 16 MB |
| fetch veraPDF | 7 s | 31 MB |
| **build the 6 documents and read them back** | **1 s** | the test itself |
| validate them with veraPDF | 3 s | |
| report + artifact | 3 s | |

The work is ~4 seconds; the rest is fetching the two tools. No Dolibarr checkout, no database service, no composer install, no cache.

## What it does not cover

The **font** half of #590 comes from the invoice PDF the *core* produces (`PDF_USE_A`), and this job has no Dolibarr to produce one; that half is a setting of the instance, which #590 reports and does not decide. What is covered is everything the module itself writes into the container.

Happy to pin the veraPDF version, drop the artifact upload, or fold the job into an existing workflow rather than add a fourth one — say which and I will change it.


daGrumpf-bxp:
@eldy Bonjour Laurent,

J'ai ajouté au corps de la PR la mesure du coût CI, dans le même esprit que sur #636, pour que la décision se prenne sur des chiffres plutôt que sur une impression :

- le job tourne en **30 s de médiane** (8 exécutions), contre 28 s pour `phan` et 38 s pour `phpcs` ;
- les jobs étant indépendants et démarrant ensemble, le temps pour avoir toutes les checks vertes ne bouge pas : c'est toujours `phpcs` qui le fixe ;
- pas de Dolibarr, pas de base de données, pas de composer : ~4 s de travail réel, le reste ce sont les deux téléchargements (TCPDF 16 Mo, veraPDF 31 Mo).

Autre chose : maintenant que #590 et #592 sont mergées, la PR est verte. L'avertissement du début (« ce job ne peut pas être vert ici ») n'a plus lieu d'être, je l'ai retiré — c'est un seul fichier sur `main`, rien à empiler.

Ce que je voulais surtout te demander, c'est ton avis. Est-ce que ce 4e workflow te convient tel quel ? Si tu préfères que je le replie dans un workflow existant, que je fige la version de veraPDF ou que je retire l'artefact, je le fais tout de suite. Et si tu juges que ça n'a pas sa place ici, dis-le simplement et je ferme la PR, sans souci — le test de #592 reste utile à lancer en local de toute façon.

Merci !

eldy:
> @eldy Bonjour Laurent,
> 
> J'ai ajouté au corps de la PR la mesure du coût CI, dans le même esprit que sur #636, pour que la décision se prenne sur des chiffres plutôt que sur une impression :
> 
> - le job tourne en **30 s de médiane** (8 exécutions), contre 28 s pour `phan` et 38 s pour `phpcs` ;
> - les jobs étant indépendants et démarrant ensemble, le temps pour avoir toutes les checks vertes ne bouge pas : c'est toujours `phpcs` qui le fixe ;
> - pas de Dolibarr, pas de base de données, pas de composer : ~4 s de travail réel, le reste ce sont les deux téléchargements (TCPDF 16 Mo, veraPDF 31 Mo).
> 
> Autre chose : maintenant que #590 et #592 sont mergées, la PR est verte. L'avertissement du début (« ce job ne peut pas être vert ici ») n'a plus lieu d'être, je l'ai retiré — c'est un seul fichier sur `main`, rien à empiler.
> 
> Ce que je voulais surtout te demander, c'est ton avis. Est-ce que ce 4e workflow te convient tel quel ? Si tu préfères que je le replie dans un workflow existant, que je fige la version de veraPDF ou que je retire l'artefact, je le fais tout de suite. Et si tu juges que ça n'a pas sa place ici, dis-le simplement et je ferme la PR, sans souci — le test de #592 reste utile à lancer en local de toute façon.
> 
> Merci !
> 

Is there a way to have this ci run only if we change a file from einvoicing directory only ?

daGrumpf-bxp:
Yes — done in the last commit: a `paths` filter on both triggers, `einvoicing/**` plus this workflow file. A PR touching another module never starts the job; it still ran here, and is green, because this file is in the list.

Only caveat: if you ever make it a required check, a PR outside `einvoicing/` would wait forever for a check that never comes — say the word and I add the `paths-ignore` twin that reports success.

daGrumpf-bxp:
A data point on what this job costs and what it catches, from two PRs opened since: #654 and #655 each add a reference fixture for an invoice type that had none. Running this workflow against both branches on my fork — [#654](https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/33056471364/job/98464402700) 25 s, [#655](https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/33056474234/job/98464412571) 36 s — it validated eight documents instead of six in both cases, and ten on the two merged, with no line of this file to touch: the test reads the whole fixture directory.

So the coverage follows the fixtures, and the cost does not: still under the phpcs job, still not the one that sets the time to green.

PR #598, opened 2026-08-17
NEW(einvoicing): add a sortable/filterable Tiers column to the flow list

Pichinov-Jose:
Follow-up usability improvement for the e-invoicing **flow list** (`einvoicing/document_list.php`).

## Problem
The flow list has no way to see, filter or sort by the related **thirdparty**. Unlike native supplier/customer document lists (which own an `fk_soc` column), `llx_einvoicing_document` has **no `fk_soc`**: a flow only references its invoice through `fk_element_id` / `fk_element_type`. So the counterparty is invisible in the list and users must open each flow to know who it concerns.

## Change
Add a **"Tiers"** column resolved on the fly from the linked invoice (`facture` / `facture_fourn` → `societe`), matching native invoice lists:

- **Display** — clickable thirdparty `getNomUrl()` (customer invoice → `Facture`, supplier invoice → `FactureFournisseur`).
- **Filter** — text input, matched with correlated `EXISTS` subqueries on both invoice types (there is no `fk_soc` on the flow table to join directly).
- **Sort** — a computed `thirdparty_name` SELECT alias, so the header behaves like any other sortable column.

## Notes
- The column is **virtual**: excluded from `getFieldList()`, from the generic search loop, and from the count query (the alias lives in `$sqlfields`).
- The `GROUP BY` block of the list is disabled, so ordering on the alias is safe.
- No schema change, no new dependency.

Tested on a Dolibarr 23 instance (column renders, filter input present, header sorts with no SQL error). Draft for maintainer review.

### DIFF
--- a/einvoicing/document_list.php
+++ b/einvoicing/document_list.php
@@ -1193,6 +1193,9 @@ $i = 0;
 $savnbfield = $totalarray['nbfield'];
 $totalarray = array();
 $totalarray['nbfield'] = 0;
+require_once DOL_DOCUMENT_ROOT.'/societe/class/societe.class.php';
+$companystatic = new Societe($db);
+
 $imaxinloop = ($limit ? min($num, $limit) : $num);
 while ($i < $imaxinloop) {
 	$obj = $db->fetch_object($resql);
@@ -1354,13 +1357,9 @@ while ($i < $imaxinloop) {
 					// Thirdparty resolved in the SELECT (thirdparty_id/thirdparty_name); render through a
 					// reusable light company object to avoid re-fetching the whole invoice for every row.
 					$out = '';
-					if (!empty($object->thirdparty_id) && !empty($object->thirdparty_name)) {
-						if (!isset($companystatic) || !is_object($companystatic)) {
-							require_once DOL_DOCUMENT_ROOT.'/societe/class/societe.class.php';
-							$companystatic = new Societe($db);
-						}
-						$companystatic->id = (int) $object->thirdparty_id;
-						$companystatic->name = $object->thirdparty_name;
+					if (!empty($obj->thirdparty_id) && !empty($obj->thirdparty_name)) {
+						$companystatic->id = (int) $obj->thirdparty_id;
+						$companystatic->name = $obj->thirdparty_name;
 						$out = $companystatic->getNomUrl(1, '', 24);
 					}
 					print $out;
--- a/einvoicing/document_list.php
+++ b/einvoicing/document_list.php
@@ -166,16 +166,21 @@ $object->fields['recap'] = array(
 	'notsearchable' => 1
 );
 
-// Add a virtual "thirdparty" field into $object->fields for the list (resolved on the fly from the linked invoice, like native invoice lists)
-$object->fields['thirdparty'] = array(
-	'label' => $langs->trans("ThirdParty"),
-	'type' => 'text',
-	'visible' => 1,
-	'enabled' => '1',
-	'position' => 52,
-	'checked' => 1,
-	'notsearchable' => 1
-);
+// Add a virtual "thirdparty" field into $object->fields for the list (resolved on the fly from the linked invoice, like native invoice lists).
+// Opt-in: resolving the thirdparty relies on correlated subqueries (llx_einvoicing_document has no fk_soc), which can be
+// heavy on very large bases, so the whole column is gated behind a hidden constant and disabled by default.
+$showthirdpartycol = getDolGlobalInt('EINVOICING_SHOW_THIRDPARTY_COLUMN');
+if ($showthirdpartycol) {
+	$object->fields['thirdparty'] = array(
+		'label' => $langs->trans("ThirdParty"),
+		'type' => 'text',
+		'visible' => 1,
+		'enabled' => '1',
+		'position' => 52,
+		'checked' => 1,
+		'notsearchable' => 1
+	);
+}
 
 
 // Initialize array of search criteria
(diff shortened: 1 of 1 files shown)


daGrumpf-bxp:
Useful column, and the SQL side is done the right way (alias added before `$sqlfields`, so the `COUNT(*)` query stays clean, and the `GROUP BY` block of this list is commented out, so sorting on the alias is safe).

One thing I would change before merge: the **display path re-loads the whole invoice for every row**, while the thirdparty name is already in the result set.

```php
$linkedobj = new Facture($db);            // or FactureFournisseur
$linkedobj->fetch((int) $object->fk_element_id);
$linkedobj->fetch_thirdparty();
$out = $linkedobj->thirdparty->getNomUrl(1, '', 24);
```

`Facture::fetch()` is not a plain SELECT: at the end of the method it calls `fetch_optionals()` and then **`fetch_lines()` unconditionally** (`compta/facture/class/facture.class.php`, same in 18 → 24). So per displayed row this costs roughly:

1. the main `SELECT` on `llx_facture` (with its joins),
2. a `SELECT` for the invoice extrafields,
3. a `SELECT` for **all the lines of the invoice**, plus one `FactureLigne` object per line,
4. `Societe::fetch()` through `fetch_thirdparty()` (`core/class/commonobject.class.php`),
5. a `SELECT` for the thirdparty extrafields.

That is ~5 queries and a fully hydrated invoice with its lines, thrown away right after, to print one link. On a default page of 50 flows: ~250 queries and several hundred short-lived objects, only for the Tiers column.

The native lists do the opposite (`compta/facture/list.php`): the thirdparty columns come from the list query itself, and a single reusable `$companystatic` is filled by hand before `getNomUrl()` — zero query per row.

The fix is small here, because this PR already computes `thirdparty_name` in the SELECT for the sort. Add the id to the same `CASE` (or a second one):

```php
// alongside thirdparty_name
$sql .= ", (CASE WHEN t.fk_element_type = 'facture' THEN (SELECT f.fk_soc FROM ".$db->prefix()."facture as f WHERE f.rowid = t.fk_element_id)";
$sql .= " WHEN t.fk_element_type = 'invoice_supplier' THEN (SELECT ff.fk_soc FROM ".$db->prefix()."facture_fourn as ff WHERE ff.rowid = t.fk_element_id)";
$sql .= " ELSE 0 END) as thirdparty_id";
```

and in the rendering loop:

```php
$companystatic->id = $obj->thirdparty_id;
$companystatic->name = $obj->thirdparty_name;
print $companystatic->id > 0 ? $companystatic->getNomUrl(1, '', 24) : '';
```

Same output, same sorting, same filtering, no per-row fetch — and it matches the pattern of the native invoice lists, which is the stated goal of the PR.

Minor, and probably acceptable as is: the correlated `CASE (SELECT ...)` is only evaluated for the returned rows, *except* when the user sorts on `thirdparty_name` — then the engine has to compute it over the whole filtered set before the `ORDER BY`. Those are primary-key lookups on `facture.rowid` / `facture_fourn.rowid`, so it should stay reasonable; it is the price of having no `fk_soc` on `llx_einvoicing_document`.

einvoicing/document_list.php
```diff
@@ -372,6 +386,12 @@
 $sql .= $hookmanager->resPrint;
 $sql = preg_replace('/,\s*$/', '', $sql);
 
+// Virtual "thirdparty_name" column resolved from the linked invoice, used to sort the Tiers column
+$sql .= ", (CASE";
```
eldy: There is a risk i see: Using this king of SQL will make SQL hang on very large database (not index use become possible) and number of request will be n x n instead of n (n number of record)
Use of CASE and 2 subrequest in the WHERE with a OR can lead in several dozen of seconds waiting with simply few thousand of record. The OR on 2 subrequest is a killer sql.

So to take no risk, can you add a hidden constant to enable all the feature that show the column "thirdparty" only when option is on. This will let us timeto find a better solution (for exemple by using a denormalized column in document_list when we insert it).

Pichinov-Jose: @eldy Done. I will prepare another PR to improve the thirdparty issue.

Pichinov-Jose: Until then this PR already works

Pichinov-Jose:
Thanks for the reviews. The three points are addressed:

- @daGrumpf-bxp (per-row `Facture::fetch()` → `fetch_lines()`): the render no longer loads the invoice. The thirdparty id is now exposed in the SELECT next to `thirdparty_name`, and a single reusable `Societe` is filled from the row (the native `compta/facture/list.php` pattern) — 0 query per row.
- @eldy (correlated subqueries + the OR-of-EXISTS filter could hang on very large bases): the whole column — virtual field, SELECT aliases and filter — is now behind a hidden constant `EINVOICING_SHOW_THIRDPARTY_COLUMN`, off by default, so the list behaves exactly as before unless an admin opts in. This leaves room for the denormalised `fk_soc` column you suggested, which I'll propose as a separate PR.
- phan CI: fixed (read the aliases from the raw row `$obj`, and hoist `$companystatic` before the record loop).

CI is green.

daGrumpf-bxp:
I ran this branch on eight instances sharing the same checkout of the module — Dolibarr 17.0.4, 18.0.10,
19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and 24.0.0-beta — with real flow tables (270 to 880 rows), plus a
throwaway 120 000-row base to answer @eldy's point about large databases. Numbers below, then the one thing
I would still change.

## The per-row fetch is gone, measured

Page of 50 flows on 23.0.3 (34 of them with a third party), SQL queries counted in the MariaDB general log:

| version | SQL queries | time |
|---|---|---|
| column off (= `main` today) | 100 | 0.19 s |
| **this branch, column on** | **100** | 0.15 s |
| first commit of this PR (`Facture::fetch()` per row) | **214** | 0.31 s |

Sorted on the column: 149 here against 328 with the first commit. So **zero query per row** for the Tiers
column — the point is settled, and the rendered values match the database row by row (50 rows × 6 pages
checked against a control query, the only difference being the expected truncation at 24 characters).

Unrelated to this PR, for whoever picks it up later: the ~2 queries per row that remain in the baseline are
the `tracking_idref` column, which does a `Facture::fetch()` — so `fetch_lines()` — per row. A page whose
rows all point to a deleted invoice costs 17 queries instead of 100.

## @eldy is right on the SQL, and the hidden constant is the right answer

Throwaway base, same query shapes as the page builds, 120 000 flows / 120 000 invoices / 2 000 third parties:

| | sort on Tiers | filter |
|---|---|---|
| default page (neither sort nor filter on Tiers) | — | 2.8 ms → **5.9 ms** |
| this PR (`CASE` + correlated subqueries, `OR` of two `EXISTS`) | **1.85 s** | COUNT 2.15 s + list 1.89 s ≈ **4 s per page** |
| same, rewritten with `LEFT JOIN` | 0.92 s | 0.77 s |
| denormalised `fk_soc` + index (your suggestion) | 0.54 s | 0.81 s |

It is linear (30 000 rows: 0.45 s / 0.61 s), so ~15-18 s per query at a million flows — "several dozen seconds"
is the right order of magnitude. `EXPLAIN` shows a full scan plus `filesort` on `llx_einvoicing_document`
(which has no index on `fk_element_id` / `fk_element_type`) and `DEPENDENT SUBQUERY` in `eq_ref` per row, so
it is n, not n×n, with a floor of ~15 µs per row. The default page is untouched — only 51 rows evaluate the
`CASE` there — which is exactly what the constant buys.

For the follow-up PR: the denormalised column pays most on the sort; the filter stays capped by the
`LIKE '%…%'` on `societe.nom`, which no index can serve.

## Functional

Rendering, sorting (both ways), filtering, pagination and the constant off/on, on 18 → 24: no PHP error or
warning, no SQL error, count of records correct with and without the filter (643 of 880 on my 23). Filter
escaping is fine (`natural_search`); `"` and `<script>` are stopped by the core with a 403 before reaching the
page. Flows whose invoice no longer exists show an empty cell, and sorting ascending puts them first.

**The 500 on Dolibarr 17 is not yours.** I checked with the file of `upstream/main` alone: same 500, same
message — `getFieldList()` only takes what to exclude since Dolibarr 18, so `t.recap` has always gone into
the SELECT there. Fixed in #619, along with two other things that page hits on 17. With both branches, the
Tiers column works on 17 too (38 third-party links out of 50 rows, sort included).

Note that #619 replaces the very line this PR changes: whichever lands first, the other resolves by putting
`'thirdparty'` into `$fieldstoexcludefromselect`.

## One thing I would still change

Turn the constant back off while a `sortfield=thirdparty_name` is still around — a bookmark, a shared URL, or
the search values Dolibarr keeps in session for the page — and the `ORDER BY` names an alias the SELECT no
longer carries: technical error page. Reproduced. One line guards it:

```php
if (!$showthirdpartycol && $sortfield == 'thirdparty_name') {
	$sortfield = 't.rowid';
}
```

PR #605, opened 2026-08-18
add findSupplierIdByRef more tolerate

frederic34:


einvoicing/class/protocols/CIIProtocol.class.php
```diff
@@ -662,6 +662,57 @@ public function generateSampleInvoiceOld($einvoicing, $thirdpartySeller = null,
...
+		global $db;
+
+		if ($ref === null || $ref === '') {
+			return 0;
+		}
+
+		$refNoSpaces = str_replace(' ', '', (string) $ref);
+
+		$sql = "SELECT rowid, ref_supplier FROM " . MAIN_DB_PREFIX . "facture_fourn";
+		$sql .= " WHERE REPLACE(ref_supplier, ' ', '') LIKE '%" . $db->escape($db->escapeforlike($refNoSpaces)) . "%'";
```
eldy, line 690: Hum, it means the sender does not really follow the einvoice rules ?
We need to be very paranoiac at begining because swith to einvoice willsurely generate a lot of problem in data (at leat at begin)so i suggest to introduce a hidden option to fix this situation but keeping the default on exact match. So we will be able to switch quickly if we discouver that not matching supplier ref is too frequent without taking the risk of false positive matching at begining

daGrumpf-bxp:
Strong +1 on @eldy's request: the default must stay on exact match. The refactoring of the five copy-pasted `SELECT ... WHERE ref_supplier = ...` blocks into a single helper is a good idea and worth keeping, but the matching rule as written can silently drop an invoice.

### Why this is not just a "false positive" risk

`findSupplierInvoiceIdByRef()` returns as soon as the `LIKE '%ref%'` query yields **one** row, without ever checking that the row is an exact match:

```php
if (count($candidates) === 1) {
    return (int) $candidates[0]->rowid;
}
```

Applied to the duplicate check in `doCreateSupplierInvoiceFromSource()`, that gives:

- already in base for supplier S: `ref_supplier = "FA2026100"`
- an e-invoice arrives with `documentno = "FA202610"`
- `REPLACE(ref_supplier,' ','') LIKE '%FA202610%'` matches `FA2026100`, one single candidate
- the function returns its id, and the caller answers `res > 0`, `"Supplier Invoice with reference FA202610 already exists"`, calls `cleanUpTemporaryFiles()` and marks the flow as processed

The invoice `FA202610` is never created, and nothing in the UI or the logs says so. Any sequential numbering (`INV-1` ... `INV-10` ... `INV-100`) triggers this, and a purely numeric `documentno` such as `24` matches `FA2024-118`. There is no minimum length, no delimiter constraint and no cross-check against the issue date, even though `$dateDoc` and `$typeDoc` are extracted two lines above and left unused.

On the referenced-document lookups the same false positive does not lose data but sets `fk_facture_source` (and the line links) to the wrong invoice.

### Other points

- `-1` now means "DB error" **and** "ambiguous", and the three call sites testing `<= 0` all report `'not found in Dolibarr'`. A SQL failure is reported to the user as a missing document, and the previous `'Database error while checking existing supplier invoice: ' . $db->lasterror()` is lost.
- The loose query manufactures the ambiguity it then rejects: with two candidates and no exact match, the import is aborted with an error, where the strict query would simply have found the right row.
- SQL `LIKE` is collation-insensitive but the exact-match filter uses `===`, so `"fac-001"` + `"xx FAC-001 yy"` is reported ambiguous although it is trivially decidable.
- `str_replace(' ', '', ...)` does not cover tabs or non-breaking spaces, which is exactly what a copy-paste from accounting produces.
- `global $db` while the class already has `$this->db` from `AbstractProtocol`.
- `FacturXProtocol::doCreateSupplierInvoiceFromSource()` is an override carrying a full copy of the same logic (lines 1474, 1499, 1526, 1613) and is not touched, so Factur-X imports keep the strict rule. And `SuperPDPProvider::findSupplierInvoiceByVendorReference()` already solves the very same problem, with a safer strategy: it disambiguates on the legal identifier carried by the CDAR and returns 0 rather than the wrong invoice. Three divergent rules for one question.
- The escaping order `$db->escape($db->escapeforlike(...))` is correct, and the `entity` / `fk_soc` filters are properly kept.

### Suggested shape

Keeping @eldy's paranoid default, and making the tolerant path narrow enough that it is still safe once switched on:

1. **Exact match first, always**, and sufficient on its own. Behaviour is then byte-for-byte the current one when the option is off.
2. **Hidden option** `EINVOICING_TOLERANT_SUPPLIER_REF_MATCH`, off by default, enabling a fallback tried *only* when the exact match found nothing.
3. **Bounded fallback**: refuse to substring-search a reference shorter than `EINVOICING_TOLERANT_SUPPLIER_REF_MIN_LENGTH` (8) or purely numeric, and require the match to be delimited by a non-alphanumeric character or an edge of the string, so `FA202610` matches `PAY123 - FA202610 - dinner` but not `FA2026100`.
4. **Report ambiguity instead of guessing**: distinct return codes `0` not found / `-1` DB error / `-2` several matches, and call sites that say which one happened.

The delimiter rule and the whitespace tolerance fight each other, and that is the part worth getting right: stripping the spaces destroys the boundary they form, so `acompte FA 2026 10 solde` stops matching `FA202610`. Recording where a gap was removed and treating it as a delimiter keeps both properties:

```php
	/**
	 * Tell whether a ref_supplier embeds a reference as a delimited substring.
	 *
	 * The reference must be bounded on both sides by a non alphanumeric character or by an edge of
	 * the string, so that "FA202610" matches "PAY123 - FA202610 - dinner" but not "FA2026100".
	 *
	 * @param  string|null 	$refSupplier 	ref_supplier value read from database
	 * @param  string 		$refNoSpaces 	Reference searched for, already stripped of whitespace
	 * @return bool 						True when the reference is embedded as a delimited substring
	 */
	private function supplierRefEmbedsReference($refSupplier, $refNoSpaces)
	{
		// Whitespace is removed so that "FA 2026 10" matches "FA202610", but removing it must not
		// destroy the boundary it forms: remember where a gap was, it counts as a delimiter.
		$haystack = '';
		$gapBefore = array();
		$pendingGap = false;
		foreach (preg_split('//u', (string) $refSupplier, -1, PREG_SPLIT_NO_EMPTY) as $char) {
			if (preg_match('/^(\s|\xc2\xa0)$/u', $char)) {
				$pendingGap = true;
				continue;
			}
			$gapBefore[strlen($haystack)] = $pendingGap;
			$pendingGap = false;
			$haystack .= $char;
		}
		$gapBefore[strlen($haystack)] = $pendingGap;

		$length = strlen($refNoSpaces);
		$offset = 0;
		while (($pos = stripos($haystack, $refNoSpaces, $offset)) !== false) {
			$end = $pos + $length;
			$startIsBounded = ($pos == 0 || !empty($gapBefore[$pos]) || !ctype_alnum($haystack[$pos - 1]));
			$endIsBounded = ($end >= strlen($haystack) || !empty($gapBefore[$end]) || !ctype_alnum($haystack[$end]));
			if ($startIsBounded && $endIsBounded) {
				return true;
			}
			$offset = $pos + 1;
		}
		return false;
	}
```

Checked against the cases that matter (the SQL `LIKE` stays as a prefilter, this rule decides):

| stored `ref_supplier` | searched ref | result |
|---|---|---|
| `PLTHDDD5OAAABBB - TEST-GROUP-2026-07-06-19407 - EVENEMENT 30/06/2026 A PARIS` | `TEST-GROUP-2026-07-06-19407` | match |
| `PAY123 - FA202610 - dinner` | `FA202610` | match |
| `acompte FA 2026 10 solde` | `FA202610` | match |
| `FA2026100 et FA202610` | `FA202610` | match (second occurrence) |
| `fa202610` | `FA202610` | match |
| `FA2026100` | `FA202610` | **no match** |
| `INV-2026-1000` | `INV-2026-100` | **no match** |
| `XFA202610` | `FA202610` | **no match** |

I have the full version (helper + the four call sites + the error messages) ready as a patch if that helps — happy to push it here or open it as a separate PR, whichever you prefer. It has not been run against a live instance yet, only against the matching cases above.

frederic34:
@daGrumpf-bxp ok make a better PR I will clode this one

daGrumpf-bxp:
@frederic34 done: #622.

It keeps your starting idea - one helper instead of the five copy-pasted `SELECT ... WHERE ref_supplier = ...` blocks - and @eldy's request, the default stays an exact match. The tolerant matching becomes a hidden option, `EINVOICING_TOLERANT_SUPPLIER_REF_MATCH`, off by default, whose fallback is tried only when the exact match found nothing and is bounded so it cannot answer for the wrong invoice: no substring search on a reference shorter than 8 characters or purely numeric, and the substring must be delimited, so `FA202610` matches `PAY123 - FA202610 - dinner` but not `FA2026100`. Stray whitespace (`FA 2026 10`, tabs, non-breaking spaces) is still tolerated, and the gap it leaves counts as a delimiter. Several candidates are reported as an ambiguity (`-2`) instead of guessed, and a database error (`-1`) is no longer reported to the user as a missing document.

Two additions to what was discussed here: `FacturXProtocol::doCreateSupplierInvoiceFromSource()` carries a copy of the same logic, so its four lookups go through the same helper - the two import paths cannot drift apart again; and the helper lives in `SupplierInvoiceHelper`, where it is covered by PHPUnit.

Ran on eight instances, Dolibarr 17.0.4 to 24.0.0-beta: the module's whole PHPUnit suite on 18.0.10 and 23.0.3, and a real import bench (`createSupplierInvoiceFromSource()`, CII and Factur-X) checking on the live database that the shorter document is created rather than swallowed as a duplicate, with a read back from a second process. Details in the PR.

Your call on closing this one.

PR #615, opened 2026-08-21
FIX CDAR lifecycle statuses for received supplier invoices

lelex86:
- Read InvoiceIssuerGlobalID from stored received invoice XML (SellerTradeParty.GlobalID) instead of Dolibarr third-party SIREN, so the PDP can match the original invoice regardless of test/synthetic SIRENs
- Add schemeID="0002" to IssuerTradeParty.GlobalID for statuses 205/210/211 (was already present for status 212)
- Fallback to Dolibarr SIREN when no stored XML is available

### DIFF
--- a/einvoicing/class/utils/CdarHandler.class.php
+++ b/einvoicing/class/utils/CdarHandler.class.php
@@ -252,10 +252,30 @@ class CdarHandler
 		// SIREN (0002)
 		$mysocGlobalID = idprof($mysoc);
 
-		// Issuer SIREN (0002) of the invoice the status is about: us when we sell, the vendor otherwise
-		$InvoiceIssuerGlobalID = $isOurOwnInvoice
-			? $mysocGlobalID
-			: thirdpartyidprof($object);
+		// Issuer SIREN (0002) of the invoice the status is about: us when we sell, the vendor otherwise.
+		// For a received supplier invoice, the platform indexed the incoming flow under the identifier
+		// the vendor's PDP placed in SellerTradeParty.GlobalID — which may differ from the Dolibarr
+		// third-party SIREN (e.g. test environments where the PDP assigns synthetic identifiers, or a
+		// vendor whose card has not been kept up to date). Reading it from the stored XML is the only
+		// reliable way to reference the same invoice the platform knows. The same parse also yields the
+		// vendor's electronic address (BT-34), used further down for MDT-73 when no routing is recorded.
+		$vendorIdentity = array('globalid' => '', 'uriid' => '');
+		if ($isOurOwnInvoice) {
+			$InvoiceIssuerGlobalID = $mysocGlobalID;
+		} else {
+			// fetch_thirdparty() populates $object->thirdparty, which the MDT-73 routing ladder below
+			// requires. thirdpartyidprof() used to provide this side effect unconditionally; now that it
+			// only runs as a fallback, the call must be explicit so the ladder is never skipped on a
+			// null thirdparty (e.g. automatic 211 from PaiementFourn::create(), where fetch() nulls it
+			// and setPaid() triggers the status before anyone fetches it back).
+			$object->fetch_thirdparty();
+			$vendorIdentity = $this->getVendorIdentityFromReceivedInvoice($object);
+			$InvoiceIssuerGlobalID = $vendorIdentity['globalid'];
+			if ($InvoiceIssuerGlobalID === '') {
+				$InvoiceIssuerGlobalID = thirdpartyidprof($object);
+				dol_syslog(__METHOD__ . ' no GlobalID found in stored XML for supplier invoice id=' . $object->id . ', falling back on Dolibarr third-party SIREN: ' . $InvoiceIssuerGlobalID, LOG_DEBUG);
+			}
+		}
 
 		// Invoice reference
 		$IssuerAssignedID = $isOurOwnInvoice
@@ -300,7 +320,7 @@ class CdarHandler
 			$vendorURIID = ($vendorRouting > 0) ? $einvoicing->removeSpaces((string) $vendorRouting) : '';	// 0 when none is recorded, -1 on error
 
 			if ($vendorURIID === '') {
-				$vendorURIID = $einvoicing->removeSpaces($this->getVendorAddressFromReceivedInvoice($object));
+				$vendorURIID = $einvoicing->removeSpaces($vendorIdentity['uriid']);
 				if ($vendorURIID !== '') {
 					dol_syslog(__METHOD__ . ' no routing ID recorded for vendor SIREN ' . $InvoiceIssuerGlobalID . ', replying to the electronic address of the invoice it sent us: ' . $vendorURIID, LOG_NOTICE);
 				}
@@ -488,20 +508,32 @@ class CdarHandler
 	}
 
 	/**
-	 * Electronic address (BT-34) the vendor put on the e-invoice we received from it.
+	 * What the e-invoice we received says about its vendor: the legal identifier it issued under
+	 * (BT-29, or BT-30 when the former is absent) and the electronic address it exchanges from (BT-34).
 	 *
-	 * That address is the vendor saying where it exchanges from, so it is the natural place to send a
-	 * status back to when no routing was recorded for it in Dolibarr. Read from the e-invoice stored with
-	 * the supplier invoice, never by calling the platform back: addressing a status is no reason for a
-	 * network round trip, and an invoice keyed by hand simply has none.
+	 * The globalid is used as InvoiceIssuerGlobalID in the CDAR so the status references the invoice
+	 * under the same identifier the platform indexed it — which may differ from the Dolibarr third-party
+	 * SIREN in legitimate cases: test environments where the PDP assigns synthetic identifiers that
+	 * Dolibarr's own validation would reject as input, or a vendor that sends via one PDP and receives
+	 * via another (its sending routing address differs from its receiving one, and cannot be derived
+	 * from the third-party card alone). Fallback to the card SIREN when the XML carries nothing.
+	 *
+	 * The uriid is the natural reply-to address when no routing is recorded for that vendor in Dolibarr.
+	 *
+	 * Read from the e-invoice stored with the supplier invoice, never by calling the platform back:
+	 * addressing a status is no reason for a network round trip, and an invoice keyed by hand has none.
+	 * A single parse serves both values, since a CDAR always needs the two together.
 	 *
 	 * @param  FactureFournisseur $object  Supplier invoice the status is sent on
-	 * @return string                      The address, '' when there is no stored e-invoice to read it from
+	 * @return array{globalid:string,uriid:string}  Empty strings when there is no stored e-invoice,
+	 *                                              when it is unparsable, or when it carries no such field
 	 */
-	private function getVendorAddressFromReceivedInvoice($object)
+	private function getVendorIdentityFromReceivedInvoice($object)
 	{
+		$identity = array('globalid' => '', 'uriid' => '');
+
 		if (empty($object->id) || $object->element !== 'invoice_supplier') {
-			return '';
+			return $identity;
 		}
 
 		dol_include_once('/einvoicing/class/utils/SupplierInvoiceHelper.class.php');
@@ -513,26 +545,41 @@ class CdarHandler
 			$xmlData = (string) SupplierInvoiceHelper::getXmlData((int) $object->id, false);
 		} catch (Exception $e) {
 			dol_syslog(__METHOD__ . ' no e-invoice stored for supplier invoice id ' . $object->id . ': ' . $e->getMessage(), LOG_DEBUG);
-			return '';
+			return $identity;
 		}
 		if ($xmlData === '') {
-			return '';
+			return $identity;
 		}
 
 		$xml = @simplexml_load_string($xmlData);
 		if ($xml === false) {
 			dol_syslog(__METHOD__ . ' the e-invoice stored for supplier invoice id ' . $object->id . ' is not parsable XML', LOG_WARNING);
-			return '';
+			return $identity;
 		}
 
-		// Only ram: is needed, so the same read works on a CII and on the XML extracted from a Factur-X
+		// Only ram: is needed — the same read works on a CII and on the XML extracted from a Factur-X
 		$xml->registerXPathNamespace('ram', 'urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100');
-		$found = $xml->xpath('//ram:SellerTradeParty/ram:URIUniversalCommunication/ram:URIID');
-		if (empty($found)) {
-			return '';
+
+		// BT-29: GlobalID[@schemeID="0002"] is the SIREN in the French mandate
+		$found = $xml->xpath('//ram:SellerTradeParty/ram:GlobalID[@schemeID="0002"]');
+		if (!empty($found)) {
+			$identity['globalid'] = removeAllSpaces((string) $found[0]);
+		} else {
+			// BT-30: SpecifiedLegalOrganization/ID[@schemeID="0002"] — legal registration number.
+			// The schemeID guard is required: without it a SIRET (0009), a GLN (0088), or a foreign
+			// registration number would be shipped to the platform declared as a SIREN.
+			$found = $xml->xpath('//ram:SellerTradeParty/ram:SpecifiedLegalOrganization/ram:ID[@schemeID="0002"]');
+			if (!empty($found)) {
+				$identity['globalid'] = removeAllSpaces((string) $found[0]);
+			}
 		}
 
-		return trim((string) $found[0]);
+		$found = $xml->xpath('//ram:SellerTradeParty/ram:URIUniversalCommunication/ram:URIID');
+		if (!empty($found)) {
+			$identity['uriid'] = trim((string) $found[0]);
+		}
+
+		return $identity;
 	}
 
 	/**


daGrumpf-bxp:
Thanks for this one — and sorry for the rewrite, my first read of it was too quick on the main point.

The missing `schemeID` on the buyer-side `IssuerTradeParty` is a real gap and it is our own: 56a7063c added the attribute to the seller branch (212) and forgot the other one. It has been shipping on every 205 / 210 / 211 since, and the platforms have accepted them anyway, which is why nobody saw it. Worth fixing regardless.

The other half needs a different conclusion, though.

### The identifier divergence is a third-party data problem, not a lookup problem

The premise is right: when the SIREN of the Dolibarr third-party card differs from the one in `SellerTradeParty` of the flow we received, the status will not match the invoice. But that divergence means the third-party card is wrong — and it is wrong everywhere else at the same time. The same value is used to issue invoices *to* that vendor, to look it up in the directory, and to derive its routing when none is recorded. Reading around it in `CdarHandler` fixes the CDAR and leaves the module quietly issuing non-conformant invoices to that same vendor, with nothing anywhere to say so.

The module already has the place to record "this vendor exchanges under X, whatever else we think": `fetchDefaultRouting()`, per third party, which is rung 1 of the MDT-73 ladder thirty lines below. Note that ladder's shape — the value deliberately recorded in Dolibarr first, the received flow only as a fallback. That order was not arbitrary; it is what a long series of "L'adresse electronique (MDT-73) est invalide" refusals taught us. The identifier deserves the same treatment: the card first, the flow as a fallback when the card holds nothing, and the divergence made loud rather than absorbed.

For a test environment where the PDP mints synthetic identifiers, the fix is to put the synthetic identifier on the third-party card. That is one field, no code, and it makes every other flow to that vendor consistent at the same time.

### Two more things on the implementation

**The BT-30 fallback can ship a non-SIREN labelled as a SIREN.**

```php
$found = $xml->xpath('//ram:SellerTradeParty/ram:SpecifiedLegalOrganization/ram:ID');
```

`SpecifiedLegalOrganization/ID` is read whatever its `@schemeID`, and the value then goes out with `'SchemeID' => CdarHandler::SCHEME_SIREN_0002` hardcoded a few lines below. A SIRET (0009), a GLN (0088), or the registration number of a vendor that is not French would be declared to the platform as a SIREN. Both XPaths should require `[@schemeID="0002"]`. That also keeps the read consistent with `idprof()`, which is country-aware.

**It duplicates `getVendorAddressFromReceivedInvoice()` almost line for line** — same guard, same `try`/`catch`, same `simplexml_load_string`, same `registerXPathNamespace`, only the XPath differs. Both run on every supplier-invoice CDAR, so each status now costs two DB reads and two full parses of the same document. A CDAR always needs the two values together, so one reader returning both is shorter and cheaper.

### Also: the branch predates the `helpers/` → `utils/` move, and will fatal on merge

The base is from 2026-08-17, two days before 6a30f601 moved `class/helpers/` to `class/utils/`. `dol_include_once('/einvoicing/class/helpers/SupplierInvoiceHelper.class.php')` no longer resolves on `main`; it fails silently, and `SupplierInvoiceHelper::getXmlData()` then raises an `Error`, which `catch (Exception $e)` does not catch. That is a fatal on every 205 / 210 / 211 send — the three statuses the PR is about, and which currently work in production. A rebase sorts it.

### Suggested version

On top of current `main`. It keeps your `schemeID` fix as is, folds the two readers into one parse, constrains both identifier XPaths to `schemeID="0002"`, and keeps the third-party card as the reference — using the flow only when the card is empty, and logging a warning when the two disagree so the card gets corrected instead of bypassed.

```diff
diff --git a/einvoicing/class/utils/CdarHandler.class.php b/einvoicing/class/utils/CdarHandler.class.php
index 0a6cefef..58077b63 100644
--- a/einvoicing/class/utils/CdarHandler.class.php
+++ b/einvoicing/class/utils/CdarHandler.class.php
@@ -252,10 +252,27 @@ class CdarHandler
 		// SIREN (0002)
 		$mysocGlobalID = idprof($mysoc);
 
-		// Issuer SIREN (0002) of the invoice the status is about: us when we sell, the vendor otherwise
-		$InvoiceIssuerGlobalID = $isOurOwnInvoice
-			? $mysocGlobalID
-			: thirdpartyidprof($object);
+		// Issuer SIREN (0002) of the invoice the status is about: us when we sell, the vendor otherwise.
+		// The vendor's legal identifier is read from its Dolibarr third-party card, which is the one place
+		// meant to hold it. The e-invoice it sent us carries it too, so the two are compared: when they
+		// disagree the third-party card is wrong, and it is wrong everywhere else as well (on the invoices
+		// we issue to that vendor, on the directory lookups, on the routing). Substituting the value read
+		// from the flow would paper the divergence over and leave the card broken, so it is logged instead
+		// and only used when the card holds nothing at all.
+		$vendorIdentity = array('globalid' => '', 'uriid' => '');
+		if ($isOurOwnInvoice) {
+			$InvoiceIssuerGlobalID = $mysocGlobalID;
+		} else {
+			$vendorIdentity = $this->getVendorIdentityFromReceivedInvoice($object);
+			$InvoiceIssuerGlobalID = thirdpartyidprof($object);
+
+			if ($InvoiceIssuerGlobalID === '') {
+				$InvoiceIssuerGlobalID = $vendorIdentity['globalid'];
+				dol_syslog(__METHOD__ . ' no legal identifier on the third-party card of supplier invoice id ' . $object->id . ', falling back on the one carried by the e-invoice it sent us: ' . $InvoiceIssuerGlobalID, LOG_NOTICE);
+			} elseif ($vendorIdentity['globalid'] !== '' && $vendorIdentity['globalid'] !== $InvoiceIssuerGlobalID) {
+				dol_syslog(__METHOD__ . ' the vendor of supplier invoice id ' . $object->id . ' issued it under ' . $vendorIdentity['globalid'] . ' while its Dolibarr third-party card says ' . $InvoiceIssuerGlobalID . ': the status is referenced with the third-party card, which should be corrected if the platform does not match the invoice', LOG_WARNING);
+			}
+		}
 
 		// Invoice reference
 		$IssuerAssignedID = $isOurOwnInvoice
@@ -300,7 +317,7 @@ class CdarHandler
 			$vendorURIID = ($vendorRouting > 0) ? $einvoicing->removeSpaces((string) $vendorRouting) : '';	// 0 when none is recorded, -1 on error
 
 			if ($vendorURIID === '') {
-				$vendorURIID = $einvoicing->removeSpaces($this->getVendorAddressFromReceivedInvoice($object));
+				$vendorURIID = $einvoicing->removeSpaces($vendorIdentity['uriid']);
 				if ($vendorURIID !== '') {
 					dol_syslog(__METHOD__ . ' no routing ID recorded for vendor SIREN ' . $InvoiceIssuerGlobalID . ', replying to the electronic address of the invoice it sent us: ' . $vendorURIID, LOG_NOTICE);
 				}
@@ -380,6 +397,7 @@ class CdarHandler
 			// We issue the status as the BUYER of a supplier invoice, and it goes back to the vendor
 			$CdarIssuerTradeParty = [
 				'GlobalID' => $mysocGlobalID, // GlobalID of CDAR SENDER
+				'SchemeID' => CdarHandler::SCHEME_SIREN_0002,
 				'RoleCode' => CdarHandler::ROLE_BY
 			];
 
@@ -488,20 +506,26 @@ class CdarHandler
 	}
 
 	/**
-	 * Electronic address (BT-34) the vendor put on the e-invoice we received from it.
+	 * What the e-invoice we received says about its vendor: the legal identifier it issued under
+	 * (BT-29, or BT-30 when the former is absent) and the electronic address it exchanges from (BT-34).
 	 *
-	 * That address is the vendor saying where it exchanges from, so it is the natural place to send a
-	 * status back to when no routing was recorded for it in Dolibarr. Read from the e-invoice stored with
-	 * the supplier invoice, never by calling the platform back: addressing a status is no reason for a
-	 * network round trip, and an invoice keyed by hand simply has none.
+	 * Both are the vendor telling us how it exchanges. The address is used when Dolibarr records no
+	 * routing for that vendor, since it is the natural place to reply to. The identifier is only used to
+	 * check the third-party card against it, and to stand in when the card carries none. Read from the
+	 * e-invoice stored with the supplier invoice, never by calling the platform back: addressing a status
+	 * is no reason for a network round trip, and an invoice keyed by hand simply has none. A single parse
+	 * serves both, since a CDAR always needs the two together.
 	 *
 	 * @param  FactureFournisseur $object  Supplier invoice the status is sent on
-	 * @return string                      The address, '' when there is no stored e-invoice to read it from
+	 * @return array{globalid:string,uriid:string}  Empty strings when there is no stored e-invoice, when it
+	 *                                              is unparsable, or when it carries no such information
 	 */
-	private function getVendorAddressFromReceivedInvoice($object)
+	private function getVendorIdentityFromReceivedInvoice($object)
 	{
+		$identity = array('globalid' => '', 'uriid' => '');
+
 		if (empty($object->id) || $object->element !== 'invoice_supplier') {
-			return '';
+			return $identity;
 		}
 
 		dol_include_once('/einvoicing/class/utils/SupplierInvoiceHelper.class.php');
@@ -513,26 +537,41 @@ class CdarHandler
 			$xmlData = (string) SupplierInvoiceHelper::getXmlData((int) $object->id, false);
 		} catch (Exception $e) {
 			dol_syslog(__METHOD__ . ' no e-invoice stored for supplier invoice id ' . $object->id . ': ' . $e->getMessage(), LOG_DEBUG);
-			return '';
+			return $identity;
 		}
 		if ($xmlData === '') {
-			return '';
+			return $identity;
 		}
 
 		$xml = @simplexml_load_string($xmlData);
 		if ($xml === false) {
 			dol_syslog(__METHOD__ . ' the e-invoice stored for supplier invoice id ' . $object->id . ' is not parsable XML', LOG_WARNING);
-			return '';
+			return $identity;
 		}
 
 		// Only ram: is needed, so the same read works on a CII and on the XML extracted from a Factur-X
 		$xml->registerXPathNamespace('ram', 'urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100');
+
+		// BT-29 first, then BT-30: both carry the SIREN, and both are only usable here when they say so
+		// themselves with schemeID="0002". Taking an identifier of another scheme (0009 SIRET, 0088 GLN,
+		// the registration number of a vendor that is not French...) and shipping it back under
+		// SCHEME_SIREN_0002 would declare it to be something it is not.
+		$sirenScheme = CdarHandler::SCHEME_SIREN_0002;
+		$identifierPaths = array('ram:GlobalID', 'ram:SpecifiedLegalOrganization/ram:ID');
+		foreach ($identifierPaths as $identifierPath) {
+			$found = $xml->xpath('//ram:SellerTradeParty/' . $identifierPath . '[@schemeID="' . $sirenScheme . '"]');
+			if (!empty($found)) {
+				$identity['globalid'] = trim((string) $found[0]);
+				break;
+			}
+		}
+
 		$found = $xml->xpath('//ram:SellerTradeParty/ram:URIUniversalCommunication/ram:URIID');
-		if (empty($found)) {
-			return '';
+		if (!empty($found)) {
+			$identity['uriid'] = trim((string) $found[0]);
 		}
 
-		return trim((string) $found[0]);
+		return $identity;
 	}
 
 	/**
```

`php -l` and the Dolibarr sniffer are clean on it. It has not been exercised against a platform yet — the one open question there is whether adding `schemeID` to the 205 / 210 / 211 `IssuerTradeParty` changes anything for a PDP that accepts them without it today. We can run that on our side.

Last thing: the `schemeID` line is independent of all of the above. If you prefer, split it into its own one-line PR and it can go in straight away, leaving this one to settle the identifier question.

daGrumpf-bxp:
One question that would settle the identifier part quickly: where does the divergence actually come from on your side?

Concretely, when the SIREN of the third-party card and the one in `SellerTradeParty` of the received flow disagree — which of the two is the odd one?

- Is the identifier in the flow one the PDP minted itself, that exists in no registry (a genuine sandbox artifact)?
- Or is it a real SIREN, and the third-party card simply carries a different or mis-keyed one?
- Does it happen in production too, or only on the test platform?
- And do the invoices you *issue* to that same vendor go out with the card's SIREN and get accepted, or do they hit the same mismatch?

It matters for what the fix should be. If the flow carries something no registry knows, the module has to accommodate it and the question is only where — I would argue on the third-party card rather than in `CdarHandler`, so every other flow to that vendor lines up too. If the card is simply wrong, then correcting it fixes the CDAR, the invoices you issue to that vendor, the directory lookups and the routing in one go, and this PR narrows down to the `schemeID` line — which is worth merging on its own either way.

A sample would help if you can share one: the `SellerTradeParty` block of a received flow next to what the third-party card holds, SIRENs redacted as long as it is visible whether they differ in shape (9 digits vs something else) or just in value.

lelex86:
Thanks for the thorough review—the technical points are valid, and this version addresses them.

What changed:

getVendorAddressFromReceivedInvoice() and getVendorGlobalIDFromReceivedInvoice() are merged into a single getVendorIdentityFromReceivedInvoice(), returning both values from one parse.

Both identifier XPaths (BT-29 and BT-30) now require [@schemeID="0002"]. You're right that without it, a SIRET, GLN, or foreign registration number would be shipped as a SIREN.
The branch is rebased on the current main; the helpers/ → utils/ path is resolved.

On the identifier priority:
The XML-first approach is kept, for two reasons that go beyond test environments.

- First, on the synthetic identifier: SuperPDP's sandbox assigns its own identifiers to test companies precisely because real SIRENs are not valid in their isolated test registry. 000000002 is not a miskeyed card—it is the identifier the platform itself issued and indexed the invoice under, and Dolibarr's SIREN validation rejects it as input anyway, so the card cannot hold it.

- Second, a production case: a vendor that sends via PDP X and receives via PDP Y has two different routing addresses (we have clients that choose 2 different PA). The one on its Dolibarr card is typically the receiving address (PDP Y). Sending the CDAR to that address routes it to PDP Y, which has no record of the invoice.

On the "just send the SIREN, the PDP will find it" fallback: that may hold for PDPs that resolve a status by SIREN alone, but the spec does not require it and we have no guarantee it holds across all platforms. Cross-platform testing between two PDPs (Esalink and SuperPDP, for instance) is not technically feasible in a test environment, so we cannot verify the behaviour empirically. Reading the routing from the received invoice guarantees the CDAR carries exactly the address the vendor's PDP used when it delivered the invoice — which is by definition the address that PDP will accept for the reply.

The fallback to the Dolibarr SIREN when the XML carries nothing, or there is no XML at all, is kept, so invoices entered by hand are unaffected.

daGrumpf-bxp:
Thanks — the rebase, the single reader, both `schemeID="0002"` guards and the `IssuerTradeParty` attribute are exactly right, and `php -l` is clean on the branch. Two things below: field data on the cross-platform question, and one regression that needs sorting before this can go in.

## The round trip does work cross-PDP, on current `main`, in production

One of our invoices, issued by our entity on SuperPDP, was delivered to a customer running Dolibarr + this module on Esalink/Hubtimize. Its lifecycle on our side, straight from the SuperPDP event log:

| Timestamp | Event | FR code |
|---|---|---|
| 07/08/2026 14:42:21 | Téléversée | |
| 07/08/2026 14:42:23 | Déposée (validée) | 200 |
| 07/08/2026 14:42:23 | Émise par la plateforme | 201 |
| 07/08/2026 14:42:24 | Déposée envoyé au PPF | |
| 07/08/2026 14:42:24 | Flux 1 envoyé au PPF | |
| 07/08/2026 14:43:05 | Reçue par la plateforme | 202 |
| 07/08/2026 14:43:10 | Mise à disposition | 203 |
| 07/08/2026 14:45:27 | Flux 1 accepté par PPF | |
| 07/08/2026 14:50:21 | Flux 1 reçu par PPF | |
| 07/08/2026 14:51:23 | Déposée reçu par PPF | |
| **24/08/2026 18:29:31** | **Approuvée** | **205** |

And yes, seventeen days between the 203 and the 205. Nothing wrong with the platforms: the customer was on holiday. August in France is part of the lifecycle now — someone should write the schematron rule.

The last line is the one that matters. That 205 is a CDAR built by this module on the customer's Dolibarr, on the Esalink side, and it crossed Esalink → PPF → SuperPDP and matched our invoice. Buyer side and seller side on two different PDPs, with the current code and `thirdpartyidprof()`.

So "cross-platform testing between two PDPs is not technically feasible in a test environment" is true of the *sandbox*, but the exchange itself is verifiable — it just has to be done in production, and it works.

## And the detail that answers the identifier question: we were not in that customer's suppliers

The vendor third party did not exist on their Dolibarr at all. Their synchronization created it from our flow — which is what `_syncOrCreateThirdpartyFromEInvoiceSeller()` does: it merges BT-30 into `sellerGlobalIds`, maps `0002` / `0225` to `idprof1` via `_mapGlobalIdSchemeToIdprof()`, and writes that value onto the card, both on the creation branch and on the update one.

So on the ordinary path the card's SIREN **is** `SellerTradeParty`'s identifier, by construction, and normalized through `removeAllSpaces()` on the way in. The two cannot diverge unless the card was edited afterwards — which is precisely the case where the card is what wants correcting. That is why the current code produced a 205 the other platform matched, for a vendor it had never seen before.

On **"the card cannot hold a synthetic identifier"**: it can. `Societe::id_prof_check()` opens with `if (getDolGlobalString('MAIN_DISABLEPROFIDRULES')) return 1;` — checked on 18, 22, 23 and 24 (18 spells it `$conf->global->`). It is a core option meant for exactly this, and it only ever makes sense on a test instance. One constant, no code, and every other flow to that vendor lines up at the same time.

On the **"sends via PDP X, receives via PDP Y"** case: that one is about the *routing address*, and this PR does not touch its precedence — `fetchDefaultRouting()` (the Dolibarr card) still wins over BT-34, exactly as before. If that is the problem you are hitting, this diff does not fix it: the change it calls for is on rung 1 of the MDT-73 ladder, not on the identifier. It is a real question and worth its own issue.

## The regression, on one of the three statuses this PR is about

`thirdpartyidprof()` calls `$object->fetch_thirdparty()`. It used to run unconditionally for every status ≠ 212, and that side effect is what populates `$object->thirdparty` before the guard thirty lines below:

```php
if ($statusCode != 212 && $object->thirdparty instanceof Societe) {   // the whole MDT-73 ladder
```

The PR only calls it in the fallback. So as soon as the XML carries a GlobalID — the nominal case the PR targets — `fetch_thirdparty()` never happens.

And there is a path where nobody else does it either. In `PaiementFourn::create()`, same shape on 18, 20, 22, 23 and 24 (line numbers from 23):

- l.282 `$invoice = new FactureFournisseur($this->db); $invoice->fetch($facid);` — and `fetch()` explicitly does `$this->thirdparty = null;`
- l.371 `$invoice->setPaid($user, '', '')` → trigger `BILL_SUPPLIER_PAYED` → our 211
- l.398 `$invoice->fetch_thirdparty()` — *after*

`setPaid()` does not fetch it either. So the automatic 211 goes out with `thirdparty === null`, the entire ladder (recorded routing → BT-34 → directory → `getBuyerCommunicationURI()`) is skipped, and `$RecipientURIID` stays the bare identifier shipped under `URISchemeID = 0225`. That is the "L'adresse electronique (MDT-73) est invalide" the ladder was written to stop. The 205 is safe (`validate()` fetches the third party) and so is the card button (the card fetched it); the automatic 211 is not.

An explicit `$object->fetch_thirdparty();` before the `else` fixes it, but it deserves a line of comment — nobody expects that side effect to be load-bearing.

## Two smaller ones

- **The blast radius of `$InvoiceIssuerGlobalID` is wider than the description says.** Besides the invoice reference it feeds `CdarRecipientTradeParty['GlobalID']` (the identity of the party the CDAR is addressed to, declared `schemeID` 0002), the `checkRecipientDirectory()` lookup, and the last-resort URIID. Changing its source is not only "reference the invoice differently", it is also "address the CDAR to whoever the flow names".
- **No `removeAllSpaces()` and no country awareness on the value read from the XML.** `idprof()` / `thirdpartyidprof()` always go through `removeAllSpaces()` and branch on `country_code`; the sync writes the card that way too. The new path does a bare `trim()`, so `<GlobalID schemeID="0002">123 456 789</GlobalID>` goes out with its spaces.
- Cosmetic: the fallback `dol_syslog(..., LOG_NOTICE)` now fires on every hand-keyed supplier invoice, not only on an anomaly.

## Standing offer

The `schemeID` on the buyer-side `IssuerTradeParty` is a genuine gap of ours and is independent of everything above. Split it into a one-liner and it can go in straight away, leaving the identifier question to settle on its own.

daGrumpf-bxp:
Split the `schemeID` line out myself so it stops waiting on the rest: #640, credited to you. It is your finding — say the word if you would rather carry it on your own branch and I will close mine.

That leaves this PR to the identifier question alone, plus the `fetch_thirdparty()` regression above.

daGrumpf-bxp:
Ran this version end to end on the sandbox today, in the exact scenario the PR is written for: a receiving Dolibarr that has never seen the vendor, so the third party is created by `_syncOrCreateThirdpartyFromEInvoiceSeller()` from the incoming flow. Setup and results below, with a negative control on `main` for every claim.

**The four review points are addressed.** The explicit `fetch_thirdparty()`, `removeAllSpaces()` on both identifier branches, `LOG_DEBUG` on the fallback, and the `schemeID` line split out to #640 — all confirmed in the deployed tree, `php -l` clean.

## Setup

dolidev (23.0.3) emits, dd23 (23.0.3) receives, SuperPDP sandbox. Deployed tree = this PR's head plus our private broker commits (the receiving instance authenticates through a partner and cannot get a token on upstream code alone); base is the PR's own base, so nothing else of ours is in the measurement.

A real `DELETE` of the vendor card was not possible — it carries 245 supplier invoices and 116 customer invoices — so every column the lookup keys on was blanked instead: `siren`, `siret`, `tva_intra`, `email`, `nom`, `name_alias`, `ref_ext`. All seven are needed, because after the idprof and VAT steps `findNearest()` still matches on name, alias, ref_ext and email. That puts the synchronization on the same creation branch, and it is reversible.

Invoice `TRI-FA-2608-0238` emitted, routed (`acknowledgement: Ok`), received on dd23, 200 and 201 back on the issuer.

## 1. The synchronization creates the card with the flow's own identifier

It created `soc#38 "Brasserie X Poitou"` with `idprof1 = 000000001` — exactly the `SellerTradeParty` identifier of the flow, and the `0225` entry of `sellerGlobalIds` does not overwrite it.

So on this path `thirdpartyidprof()` and the XML read return the same value, and **the CDAR is byte-identical with and without this PR**. That is the case the PR narrows down to: the card cannot diverge from the flow unless someone edited it afterwards.

## 2. The CII this module emits carries no BT-29 at all

```
stored XML BT-29 GlobalID[0002]      : (absent)
stored XML BT-30 LegalOrg ID[0002]   : '000000001'
stored XML BT-34 URIID               : '000000001'
```

`SellerTradeParty/GlobalID[@schemeID="0002"]` is simply not written. The SIREN lives only in `SpecifiedLegalOrganization/ID`. So the BT-30 branch of this PR is not a corner case, it is the only source — and the `schemeID="0002"` guard you added on it is what stops a SIRET or a foreign registration number being shipped as a SIREN. Good that it is there.

## 3. The gap is the routing address, and it is not this PR's

Look at BT-34 above: the flow's electronic address is `000000001`, the bare SIREN. Not the vendor's actual routing (`315143296_2341`). With a freshly created card there is no recorded routing, so the MDT-73 ladder takes rung 2 — the address the vendor put on the invoice it sent us — and the platform answers:

```
sendStatusMessage(205) -> {"res":-1,"message":"Failed to send CDAR file to PDP.
  Status code: 400. Message: L’adresse électronique (MDT-73) est invalide"}
```

**Negative control on `main`, same invoice, same instant: identical 400.** So this is pre-existing and *not* a regression of this PR — I want to be clear about that. But it does bear on the reasoning in the PR description:

> Reading the routing from the received invoice guarantees the CDAR carries exactly the address the vendor's PDP used when it delivered the invoice — which is by definition the address that PDP will accept for the reply.

Measured here, that does not hold: BT-34 is what the *sender* declared, not what the platform routed under, and the platform refuses it. The ladder walked:

| rung | source | value |
|---|---|---|
| 1 | `fetchDefaultRouting()` | `0` (nothing recorded — the card is new) |
| 2 | BT-34 of the stored flow | `000000001` — **taken, refused by the platform** |
| 3 | `checkRecipientDirectory()` | `status: absent` (the sandbox directory is empty) |
| 4 | `getBuyerCommunicationURI()` | `000000001` |

Rung 2 being non-empty pre-empts rungs 3 and 4, so a bad address stops the ladder. Only rung 1 works. Once the routing is recorded on the created card, the same 205 goes out `validation=Ok` and lands back on the issuer, next to its 200 and 201:

```
lifecycle_msg 200 () flow=ie_1007864
lifecycle_msg 201 () flow=ie_1007865
lifecycle_msg 205 () flow=ie_1007920
```

That is a separate issue and worth its own — the "sends via PDP X, receives via PDP Y" case you raised belongs there too, since it is about rung 1's precedence, which this PR does not touch.

## 4. The `fetch_thirdparty()` fix is load-bearing, and it works

Built the CDAR for the automatic-211 shape (invoice fetched, nobody calls `fetch_thirdparty()` beforehand) on the same invoice, on three trees:

| tree | `$object->thirdparty` after | MDT-73 emitted |
|---|---|---|
| `main` | loaded | `315143296_2341` ✅ |
| this PR with the `fetch_thirdparty()` line removed | **NULL** | `000000001` ❌ |
| this PR as it stands | loaded | `315143296_2341` ✅ |

And `000000001` under scheme 0225 is exactly the value the platform rejected in §3. So without that line the automatic 211 would have gone out with an address the platform refuses. With it, a real payment through `PaiementFourn::create()` + `setPaid()` fires the 211 and it is accepted (`validation=Ok`) and reaches the issuer.

## Where that leaves it

The code is correct and does no harm — same document as `main` on the nominal path, and the regression is closed. My reservation is unchanged and is about scope rather than correctness: on the path measured here the two identifier sources agree by construction, so the change buys nothing, while the failure it is meant to prevent turns out to be a routing-address problem one rung up. If you have a captured flow where `SellerTradeParty`'s identifier genuinely differs from what the card holds, that would settle it — I could not produce one.

Incidental, since it showed up in the generated file: the buyer-side `IssuerTradeParty` still goes out as `<ram:GlobalID>000000002</ram:GlobalID>` with no `schemeID` — the platform accepted it, but that is the gap #640 closes.

lelex86:
One point of clarification on the framing: this PR is not specifically about the case where the vendor is unknown and created by _syncOrCreateThirdpartyFromEInvoiceSeller(). That scenario is one data point, not the premise.

The premise is simpler: a CDAR must reference the invoice under the same identifier the platform indexed it with. The platform indexed it from SellerTradeParty in the received flow. So we read InvoiceIssuerGlobalID from there — not because we expect the card to be wrong, but because the received invoice is by definition the authoritative source for that value. If they agree, the result is identical to thirdpartyidprof() and nothing changes. If they ever differ — composite identifier, sync that could not map the value to idprof1, card edited after the fact, ... — the CDAR still works.

It is a safety net, not a targeted fix for a specific path. The concrete confirmed benefit is the fetch_thirdparty() fix, which your own measurements show prevents a real failure on automatic 211s. The rest is a guarantee that costs nothing.

PR #616, opened 2026-08-21
fix(einvoicing): received supplier-invoice status dropdown shows "Array"

Pichinov-Jose:
### Bug

On a **received supplier invoice**, the **"E-invoice"** action button opens a dropdown whose items all read **`Array`** instead of the sendable PDP statuses:

`addMoreActionsButtons()` (context `invoice_supplier`) builds the dropdown from `getSendableStatusesForReceivedInvoice()` → `getEinvoiceStatusOptions()`. That method was changed to return **array-shaped values** (`['label' => ..., 'data-html' => ...]`, as its docblock now documents), but this consumer still does:

```php
'label' => (string) $label,
```

Casting the array to a string yields the literal `"Array"`, one per sendable status.

### Fix

Extract the `label` field, with a fallback so the code also works if a plain string is returned:

```php
'label' => (string) (is_array($label) ? ($label['label'] ?? '') : $label),
```

### Notes

- Reproduced on Dolibarr 23.0.3 with a supplier invoice in a "received" PDP state offering several sendable statuses.
- Only this consumer was affected; the select-list consumers already handle the array shape.

Draft for maintainer review.

### DIFF
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -3,6 +3,7 @@
  * Copyright (C) 2025		Laurent Destailleur		<eldy@users.sourceforge.net>
  * Copyright (C) 2026		Charlene Benke			<charlene@patas-monkey.com>
  * Copyright (C) 2026       Frédéric France         <frederic.france@free.fr>
+ * Copyright (C) 2026		Jose Martinez				<jose.martinez@pichinov.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
@@ -473,7 +474,7 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 						'lang' => 'einvoicing',
 						'enabled' => true,
 						'perm' => ($forcedisabling ? -1 : ((bool) $user->hasRight("fournisseur", "facture", "creer") && empty($forcedisabling))),
-						'label' => (string) $label,
+						'label' => (string) (is_array($label) ? ($label['label'] ?? '') : $label),
 						'url' => '/fourn/facture/card.php?id=' . $object->id . '&action=sendStatusMessage&pdpstatuscode=' . $code . '&token=' . newToken()
 					);
 				}


daGrumpf-bxp:
Reviewed and tested. The diagnosis is right, and the fix is the correct one-liner.

### Root cause

`getEinvoiceStatusOptions()` normalises **every** value to `array('label' => …, 'data-html' => …)`
since 20c665a7 ("Better label and position for einvoice status combo list"). The other consumers cope:
`Form::selectarray()` normalises non-array options itself (identically from 17.0 to 24.0), and the two
providers only read the keys. The button group of the received supplier invoice card was the only place
left casting the value with `(string)`, so each dropdown item printed `Array`, with an
`Array to string conversion` warning behind it.

No fatal is possible here: `Translate::trans()` is untyped on every core from 17.0 to 24.0, so the
array reaches it and comes back as the literal `Array`.

### Tested on 8 cores, before and after

One received supplier invoice per instance (a real one, with its `einvoicing_extlinks` row), the hook
replayed in the `invoicesuppliercard` context and the produced HTML read back:

| Dolibarr | before (`main` @ ab35dbc0) | after (`main` + this PR) |
|---|---|---|
| 17.0.4 · 18.0.10 · 19.0.4 · 20.0.4 · 21.0.4 · 22.0.5 · 23.0.3 · 24.0.0-beta | `Array \| Array \| Array` + PHP warning | `(210) Refused by customer \| (205) Approved by customer \| (211) Payment transmitted` |

8/8 broken before, 8/8 correct after. Control on the three other consumers (both list filters and the
create/edit combo): 24/15/3 options, no `Array`, on the same 8 cores.

### End to end, against a sandbox Access Point

Deployed tree = `main` + this patch. An invoice emitted on 23.0.3 was routed by the AP and received on
both 23.0.3 and 18.0.10; the dropdown items were then actually used:

- `sendStatusMessage(205)` from the 23.0.3 card → accepted (`validation=Ok`),
- `sendStatusMessage(211)` from the 18.0.10 card → accepted,
- back on the issuer: 200, 201, 205, 211.

The group also narrows correctly afterwards (only 211 left on one side, 210/205 on the other), including
the single-item case where the core renders a plain button instead of a dropdown.

PHPUnit `ReceivedInvoiceStatusesTest` (5 tests) and `SupplierInvoiceHelperTest` (29 tests): green on the
8 cores.

LGTM.

### Unrelated finding, same commit

While testing the combos on 17.0.4 I hit a side effect of 20c665a7 that this PR does not cover: the
`data-html` decoration it adds contains a `<span class="small opacitymedium">`, and `selectarray()` of
17.0 prints `data-*` values **raw** (`html.form.class.php`), where 18.0 and later wrap them in
`dol_escape_htmltag()`. On 17.0 the attribute therefore closes on its first quote:

```html
<option value="200" data-html="Deposited  <span class="small opacitymedium">(200 - Mandatory)</span>">Deposited</option>
```

The module declares `need_dolibarr_version = array(17, -3)`, so 17.0 is in the announced range. a separate PR for it opened as #617.

daGrumpf-bxp:
Opened as #617.

Pichinov-Jose:
Thanks a lot for the thorough review and the cross-version testing — much appreciated.

Agreed on the root cause: getEinvoiceStatusOptions() normalises every value to `array('label' => …, 'data-html' => …)`, and this received-invoice button group was the last consumer still `(string)`-casting it, hence the "Array" per item.

Good catch on the separate 17.0 `data-html` escaping issue — keeping it out of this PR is the right call; #617 is the proper place for it. This one stays a focused one-liner for the "Array" regression.

PR #619, opened 2026-08-22
FIX [einvoicing] Make the flow list page work on Dolibarr 17

daGrumpf-bxp:
`einvoicing/document_list.php` has never been able to render on **Dolibarr 17**, the version
`modEInvoicing.class.php` declares as the minimum the module supports (`need_dolibarr_version = array(17, -3)`).
Every other page of the module renders there — index, document card, call list, the four admin pages — this one
answers a technical error. Three things stand in the way, and each one only shows once the previous is out.

## 1. `getFieldList()` takes what to exclude only since Dolibarr 18

```php
$sql .= " ".$object->getFieldList('t', array('recap', 'xml_data', 'thirdparty'));
```

On 17 the signature is `getFieldList($alias = '')`: the second argument is ignored without a word, so `recap` —
a **virtual column of this list, with no column in the table** — goes into the SELECT and the page reports:

```
DB_ERROR_NOSUCHFIELD — Unknown column 't.recap' in 'SELECT'
```

Below 18 the page now builds the same list the core builds since 18; on 18 and above `getFieldList()` is still
the one doing it, so **nothing changes there** (checked field by field on the 8 versions: the string the page
builds and the string the core returns are identical from 18 on).

## 2. `GETPOSTDATE()` does not exist on 17 either

Once the query passes, the page dies on `Call to undefined function GETPOSTDATE()` (the lookback of the sync
form). That helper arrived in the core in 18, and it is backported here in **`compat/functions.lib.php`**, next
to the core helpers the module already ships for 17 (`compat/files.lib.php`, `compat/profid.lib.php`). The page
also loads the module library, for `dolPrintHTMLForAttribute()` — a function of Dolibarr 19, already polyfilled
there, that the same page calls to render its text columns and that would fatal on 17 and 18 for the same reason.

**The 18 implementation is the one to copy, and that is not a detail.** The core rewrote this function twice:
18 reads the `hour`, `min` and `sec` fields; **19, 20 and 21 read `minute` and `second`** — names their own
`Form::selectDate()` does not post, so on those three versions the core's own `GETPOSTDATE()` silently ignores
the minutes and seconds typed by the user; 22 came back to `min`/`sec` and added a fourth argument this module
does not use. On 17 the selector posts `hour`/`min`/`sec`, which is what the backport reads.

Checked, not assumed: the backport was run side by side with the core implementation on 18, 20, 23 and 24 over
seven inputs (empty selector, `getpost`, a forced `HH:MM:SS`, out-of-range values, a leap day, a date before
1970's-era timestamps, three timezones). **Identical on 18, 23 and 24**; the two differences on 20 are the
`minute`/`second` quirk above, on the core's side.

## 3. `$conf->main_checkbox_left_column` only exists since Dolibarr 21

Six places read it directly, so every load on 17, 18, 19 and 20 wrote `Undefined property:
Conf::$main_checkbox_left_column` in the log, for a value that is empty there anyway. They read it through
`empty()` now — which is what a seventh place already did.

## Tested

Live on eight instances, Dolibarr **17.0.4, 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and 24.0.0-beta**,
sharing the same checkout of the module, each page loaded over HTTP with a real session.

- **17.0.4**: **HTTP 500 → 200**. The list renders its rows, paginates (page 2), sorts, filters
  (`search_flow_direction=Out` → 229 and `In` → 574, out of 803), honours `limit=50`, the synchronization form
  and its date selector are back, and the page emits no PHP warning of its own.
- **18 → 24**: same page, same rows, and the **same number of SQL queries as on `main`** (57, 16, 74, 20, 20,
  35, 44), no new warning. The Tiers column of #598 still renders on the instance that has it.
- The code path that calls `GETPOSTDATE()` is behind `EINVOICING_SYNC_MARGIN_TIME_HOURS`; the constant was set
  on 17, 18 and 20 so the page really goes through it, with and without the date fields posted.
- Module unit tests, file by file, on the **eight** cores: 112 files, no failure.
- Five end-to-end sandbox cycles on this branch, invoice emitted → routed by the Access Point → imported on the
  other side, statuses 200/201 back to the sender: **17 → 23**, **23 → 17**, 18 → 23, 24 → 23 and 23 → 23. Line
  billing periods intact each time, no transaction left open.
- `phpcs` (Dolibarr ruleset) clean on the three files.

## Note for the maintainer

Rebased on `main` after #598 landed; `'thirdparty'` is in the excluded array, which is the only thing that had
to carry it.


eldy:
I don't think it is a good idea to try to keep the module compatible with too old version of dolibarr. 
V18 remains a good choice, it is a very old version but still maintained by a dedicated LTS team where v17 is older and with very low (none) maintenance contributions.
So I think we should just forgot the compatibility with v17.
Also all past communication done on this module always mention that compatibility will be supported until v18 and not lower.

Will ask confirmation to the board team.

daGrumpf-bxp:
Closing: this is already in `main`.

The branch was stacked, so merging #623 brought three commits in at once - a832aace (this PR), 9357a2fc (#620) and d316e658 (#623). The branch was simply behind, which is all the conflict was about: `main` merged into it resolves on `ChangeLog.md` alone, and even there the entry of this PR was already present, so the resolution is `main`'s own file, with no duplicate. After that merge `git diff origin/main HEAD` is empty - the flow list page fix, the `getFieldList()` handling and the `GETPOSTDATE()` backport are all in place.

Nothing left to review or to merge here.

PR #620, opened 2026-08-23
CHANGE [einvoicing] Move the dolPrintHTMLForAttribute backport next to the other core helpers, in compat/

daGrumpf-bxp:
`dolPrintHTMLForAttribute()` entered the core in **Dolibarr 19**. The module backports it for the versions
below, and that backport sat in `lib/einvoicing.lib.php`, among the functions of the module itself.

The module already has a home for this kind of code: `compat/`, where `files.lib.php` backports `dolChmod()`
and `profid.lib.php` the professional-id helpers. The backport moves there, into `compat/functions.lib.php`,
named after the core file it comes from.

**Pure move.** `lib/einvoicing.lib.php` requires that file, so the six call sites — `document_card`,
`call_card`, the third party routing widget, the supplier invoice list column, `call_list.php` and
`document_list.php` — keep finding the function exactly where they used to. `document_list.php`, the only
page that loaded the library for nothing else, now loads the compat file alone.

## Worth knowing for the versions below 19

The backport calls `dol_escape_htmltag()` with six arguments, and **that function only takes five on
Dolibarr 17**. PHP passes the extra argument to a userland function without complaining, so the sixth one
(`$cleanalsojavascript`) is simply not read there. It changes nothing for this use: the fifth argument is
`0`, which already escapes the whole string.

Checked rather than assumed — same input through `dolPrintHTMLForAttribute()` on four cores:

| input | 17 (backport) | 18 (backport) | 19 (core) |
|---|---|---|---|
| `<script>alert(1)</script>` | escaped | escaped | escaped |
| `<a href="javascript:alert(1)">x</a>` | `x` | `x` | `x` |
| `` | empty | empty | empty |
| `<b>bold</b> and <i>italic</i>` | `&lt;b&gt;bold&lt;/b&gt; and italic` | same | same |
| `quote " and apostrophe '` | `quote &quot; and apostrophe '` | same | same |

Identical on the three, plain text and accents included. The one difference is a newline kept after the
`<br>` on 17 and 18 and dropped on 19 — that comes from `dol_htmlentitiesbr()` in the core, not from this
code.

## Tested

Live on eight instances, Dolibarr **17.0.4, 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and
24.0.0-beta**, sharing the same checkout of the module.

- Every page that calls the function, loaded over HTTP on 17, 18, 19 and 23: the flow list, the call list,
  the flow card, the module index, the vendor reference list, the third party card, the supplier invoice
  list and card. HTTP 200, no PHP error, no warning of the module's own.
- Module unit tests, file by file, on the **eight** cores: 112 files, no failure.
- `phpcs` (Dolibarr ruleset) clean on the three files.

## Note for the maintainer

**This moves code you placed yourself, so the question deserves to be asked out loud.** On 2026-05-20,
two commits titled "Introduce compatibility code for older versions" landed together: `bdaacae7` created
`compat/` with `profid.lib.php` and `commonhookactions.class.php`, and `e02346ca` added `GETPOSTFLOAT()`
and `dolPrintHTMLForAttribute()` to the library of the module. So `compat/` already existed when these
two were written, and they were not put there.

A reading of that split is that `compat/` takes what can be substituted **wholesale** - a core file or a
core class the module loads instead of the core one - while a lone helper of `functions.lib.php` cannot
be substituted (the core always loads that file) and only needs a `function_exists()` guard, which works
from anywhere.

The reason to move it anyway: `compat/` reads as *the* place where this module keeps what it backports
from the core, and `dolPrintHTMLForAttribute()` is a verbatim copy of the core function - checked
character for character against Dolibarr 23 and 24, the only difference being the phan annotation. Kept
in the library, it sits among the functions the module itself owns, and a reader has no way to tell one
from the other. The guard makes the file it lives in a free choice, so it may as well be the explicit one.

Happy to close this and keep the backports in the library if you prefer the original split - the two
files it touches are the only thing at stake. The same question applies to `GETPOSTFLOAT()`,
`getDolGlobalFloat()`, `getMultidirTemp()`, `getMultidirVersion()` and `dolPrintHTML()`, still in the
library; whichever way this goes, they should follow, one PR each.

Stacked on #619, which creates `compat/functions.lib.php` for `GETPOSTDATE()`. If #619 lands first this
applies as is; taken alone, the file is created here instead.


einvoicing/document_list.php
```diff
@@ -378,7 +382,25 @@
 // Build and execute select
 // --------------------------------------------------------------------
 $sql = "SELECT";
-$sql .= " ".$object->getFieldList('t', array('recap', 'xml_data', 'thirdparty'));
+// Fields of the table for the SELECT. 'recap' and 'thirdparty' are virtual columns of this list - they
+// have no column in the table - and 'xml_data' is never displayed, so the three must stay out of the query.
+$columns = array('recap', 'xml_data', 'thirdparty');
+if ((float) DOL_VERSION < 18) {
```
eldy, line 388: We can abandon maintenability for version < 18, it is an endless race.
Support of v17 does not worth the cost.

daGrumpf-bxp: Agreed, and no argument from me - I said the same on #621. Two things on this block in particular.

It does not come from this PR: it entered `main` with #619 (commit a832aace), which your merge of #623 pulled in along the way, this branch being simply behind. So the line you are pointing at is already in `main`, and removing it is its own change, not a fix to this one.

The removal is ready whenever the board confirms. There are five places built for 17 alone - `compat/files.lib.php` (dolChmod), `GETPOSTDATE()` in `compat/functions.lib.php`, this `getFieldList()` block, the `data-html` escaping in `class/einvoicing.class.php:817-828`, and the dropdown polyfill in `lib/einvoicing.lib.php:394-427` with its two call sites - roughly 120 lines and 4 branch points. The full list is in my comment on #621. `compat/` does not empty out: what stays there is for 19, 20 and 21, not for 17.

I would rather wait for the board answer than land it now, so it is one clean PR against a decision, not a piecemeal removal. Say the word and I open it.

daGrumpf-bxp:
Closing: this is already in `main`.

The branch was stacked, so merging #623 brought three commits in at once - a832aace (#619), 9357a2fc (this PR) and d316e658 (#623). Rebased on today's `main`, this branch replays to nothing: `git diff origin/main HEAD` is empty, the move of `dolPrintHTMLForAttribute()` to `compat/functions.lib.php` is in place, and the only thing left to merge was the ChangeLog, whose entry `main` already carries too.

Nothing to review here any more. The v17 point you raised is answered in the thread above and on #621.

PR #621, opened 2026-08-23
FIX [einvoicing] Two pages die on Dolibarr 17 on the supplier filter of their third party selector

daGrumpf-bxp:
Two pages of the module answer a technical error on **Dolibarr 17**, the version
`modEInvoicing.class.php` declares as the minimum it supports (`need_dolibarr_version = array(17, -3)`):
`einvoicing/vendorref_list.php` and `einvoicing/product_mapping.php`.

Both narrow their third party selector to suppliers, and both write that criteria in the Universal Search
syntax:

```php
print $form->select_company($search_socid, 'search_socid', '(s.fournisseur:=:1)', ...);
```

`Form::select_company()` hands the criteria to `select_thirdparty_list()`, and **that method only converts
the Universal Search syntax from Dolibarr 18 on**. On 17 there is no conversion at all — the criteria is
appended to the query exactly as given:

```php
if ($filter) {
    $sql .= " AND (".$filter.")";
}
```

so the page runs

```sql
... WHERE s.entity IN (1) AND ((s.fournisseur:=:1)) ORDER BY nom ASC
```

and stops on `DB_ERROR_SYNTAX — check the manual ... near ':=:1)) ORDER BY nom ASC'`. The error page is
served in place of the list.

## Why the criteria has to depend on the version

The two forms are not interchangeable across the range the module supports:

| Dolibarr | what `select_thirdparty_list()` does with the criteria | plain SQL | Universal Search |
|---|---|---|---|
| 17 | appended as it stands | works | **breaks the query** |
| 18 → 23 | converted when it holds parentheses, appended as it stands otherwise | works (with a `dol_syslog` warning) | works |
| 24 | **always** converted, the parentheses test is gone | **mangled** | works |

So neither form alone covers 17 to 24, and the criteria is picked from the version:

```php
$vendorfilter = ((float) DOL_VERSION < 18) ? 's.fournisseur = 1' : '(s.fournisseur:=:1)';
```

## Tested

Live on eight instances, Dolibarr **17.0.4, 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and
24.0.0-beta**, sharing the same checkout of the module, both pages loaded over HTTP with a real session.

- **17.0.4**: both pages go from the error page to **HTTP 200**, with no PHP error and no warning of their
  own. On `main` the same two pages report `DB_ERROR_SYNTAX` there.
- **18 → 24**: both pages unchanged, HTTP 200, no error, no new warning.
- **The filter really filters**, which an error-free page alone would not prove: the number of entries
  offered by the selector was compared to the number of active suppliers in the database on 17, 18, 20, 21
  and 24 — 2/2, 3/3, 3/3, 2/2 and 8/8, against 3, 9, 10, 3 and 11 active third parties in total.
- Module unit tests, file by file, on the eight cores: 112 files, no failure.
- `phpcs` (Dolibarr ruleset) clean on both files.

### DIFF
--- a/einvoicing/product_mapping.php
+++ b/einvoicing/product_mapping.php
@@ -277,7 +277,14 @@ print '<input type="text" class="width200" name="flowid" value="'.dol_escape_htm
 print '</div>';
 print ' &nbsp; ';
 print '<div class="inline-block valignmiddle paddingright">'.$langs->trans("Supplier").' ';
-print $form->select_company($socid, 'socid', '(s.fournisseur:=:1)', 'SelectThirdParty', 0, 0, array(), 0, 'minwidth200');
+// Form::select_company() does not read this filter the same way on every version. Until Dolibarr 18
+// the criteria is appended to the query as it stands, so it has to be plain SQL; from 18 a criteria
+// holding parentheses is converted by forgeSQLFromUniversalSearchCriteria(), and on 24 that conversion
+// is no longer conditional - every criteria goes through it. Passing the Universal Search syntax to 17
+// therefore ends the page on "DB_ERROR_SYNTAX ... near ':=:1))'", and passing plain SQL to 24 would be
+// mangled the other way round.
+$vendorfilter = ((float) DOL_VERSION < 18) ? 's.fournisseur = 1' : '(s.fournisseur:=:1)';
+print $form->select_company($socid, 'socid', $vendorfilter, 'SelectThirdParty', 0, 0, array(), 0, 'minwidth200');
 print '</div>';
 print '<input type="submit" class="button small" value="'.$langs->trans("Refresh").'">';
 print '</form>';
--- a/einvoicing/vendorref_list.php
+++ b/einvoicing/vendorref_list.php
@@ -345,7 +345,14 @@ if ($leftcolumn) {
 	print '</td>';
 }
 print '<td class="liste_titre">';
-print $form->select_company($search_socid, 'search_socid', '(s.fournisseur:=:1)', '1', 0, 0, array(), 0, 'maxwidth200');
+// Form::select_company() does not read this filter the same way on every version. Until Dolibarr 18
+// the criteria is appended to the query as it stands, so it has to be plain SQL; from 18 a criteria
+// holding parentheses is converted by forgeSQLFromUniversalSearchCriteria(), and on 24 that conversion
+// is no longer conditional - every criteria goes through it. Passing the Universal Search syntax to 17
+// therefore ends the page on "DB_ERROR_SYNTAX ... near ':=:1))'", and passing plain SQL to 24 would be
+// mangled the other way round.
+$vendorfilter = ((float) DOL_VERSION < 18) ? 's.fournisseur = 1' : '(s.fournisseur:=:1)';
+print $form->select_company($search_socid, 'search_socid', $vendorfilter, '1', 0, 0, array(), 0, 'maxwidth200');
 print '</td>';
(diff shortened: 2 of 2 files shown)


eldy:
I don't think it is a good idea to try to keep the module compatible with too old version of dolibarr.
V18 remains a good choice, it is a very old version but still maintained by a dedicated LTS team where v17 is older and with very low (none) maintenance contributions.
So I think we should just forgot the compatibility with v17.
Also all past communication done on this module always mention that compatibility - will be supported until v18 but not lower.

Will ask confirmation to the board team.

daGrumpf-bxp:
No objection from me, and the call is yours and the board's, not mine. Just one factual point so the decision is taken on what the tree actually contains today.

The descriptor is back to `array(18, -3)` since 6fb1149e this morning, but **the code shipped in `main` still branches on 17 in five places**. So right now the descriptor says 18 while the module is, in fact, still built to run on 17:

| what | where in `main` | added to the core in |
|---|---|---|
| `dolChmod()` backport | `einvoicing/compat/files.lib.php` (50 lines), included by `class/protocols/CIIProtocol.class.php:38-41` under `DOL_VERSION < 18` | 18 |
| `GETPOSTDATE()` backport | `einvoicing/compat/functions.lib.php:28-69`, included by `document_list.php:85` | 18 |
| `getFieldList()` exclusion list rebuilt by hand | `einvoicing/document_list.php:388-402` (`DOL_VERSION < 18`) | 18 |
| `data-html` of a combo option escaped by the module | `einvoicing/class/einvoicing.class.php:817-828` (`DOL_VERSION < 18`) | 18 escapes it itself |
| `einvoicingDolGetButtonActionDropdown()` polyfill | `einvoicing/lib/einvoicing.lib.php:394-427`, called at `class/actions_einvoicing.class.php:428-430` and `:492-494` | 18 (`dolGetButtonAction()` array mode) |

That is roughly 120 lines plus 4 branch points that exist for 17 alone. Everything else in `compat/` is **not** 17: `commonhookactions.class.php` and `dolPrintHTMLForAttribute()` are for 19, `GETPOSTFLOAT()` and `profid.lib.php` for 20, `getDolGlobalFloat()` for 21 — dropping 17 leaves all of those in place.

**This PR is one of those five branches** — the last v17 breakage I had found. If v18 becomes the floor, then it is simply not needed: the criteria already on `main`, `'(s.fournisseur:=:1)'`, is the correct one from 18 on, and I close this PR with no code change at all. Same for the other 17-only work.

So, concretely, on board confirmation: I open a single PR "drop Dolibarr 17 support" that removes the five items above, and the module then matches its own descriptor. Tell me the verdict and I do it — happy either way, it is less code to carry.

daGrumpf-bxp:
Rebased: the `ChangeLog.md` entry is dropped, since `.agents/AGENTS.md` asks not to touch that file
(it is generated from the commit titles at release time). The diff is now the two pages only.

Worth raising while this is open: the premise of the PR has moved. `modEInvoicing` now declares
`need_dolibarr_version = array(18, -3)`, so the Dolibarr 17 branch this adds (`s.fournisseur = 1`
below 18) guards a version the module no longer claims to support, and on 18 and above the value
passed is exactly the one already in `main`. If 17 is definitively out, this PR has nothing left to
do and I will close it - tell me which way you want it.

eldy:
Board meeting is in 2 week. I keep the PR open until this time.

daGrumpf-bxp:
Reopened — closing this was my mistake: @eldy asked to keep it open until the board meeting, and I closed it three days ago while cleaning up my branches. Sorry for the noise. Branch restored at the same commit (ae93747), still mergeable on main, and it stays open until the board decides on v17.

PR #628, opened 2026-08-23
NEW Add new community module 'DoliSecure'

LaurentBouquet:
This PR proposes the integration of DoliSecure, a module that automatically checks if the installed Dolibarr version is affected by known vulnerabilities (CVEs) by querying the public NVD database.

Security monitoring for an ERP is critical but usually relies on admins manually checking external advisories. DoliSecure automates this. Including it in the community repository gives administrators a direct way to be alerted about vulnerabilities, reducing the risk of running unpatched instances.

Features : 
- Queries the NVD database based on the current Dolibarr version.
- Displays a danger-level indicator on the home page.
- Sends an email to administrators as soon as a new vulnerability is detected.
- Provides a dedicated report page listing each CVE.

The module strictly follows Dolibarr development standards. It does not modify any core files, uses standard module mechanisms, and is licensed under GPLv3+.

To test the module, you can install it on a Dolibarr version 23 instance, and the module will report the vulnerabilities in that version.

Thanks


dev/tools/phan/baseline.txt
```diff
@@ -45,6 +45,7 @@ return [
         'einvoicing/class/providers/SuperPDPProvider.class.php' => ['PhanTypeArraySuspiciousNullable', 'PhanTypeMismatchArgumentProbablyReal', 'PhanUndeclaredProperty'],
         'einvoicing/compat/commonhookactions.class.php' => ['PhanRedefineClass'],
         'einvoicing/document_card.php' => ['PhanUndeclaredProperty'],
+        'einvoicing/einvoicingindex.php' => ['PhanNonClassMethodCall'],
```
eldy: Is adding this line done on purpose ?

LaurentBouquet:
I added this line so that the PR could be approved. It might be a mistake on my part, but I thought that if the tests remained stuck, the module I'm proposing would never be reviewed.

daGrumpf-bxp:
Hi @LaurentBouquet, no worries — but the phan failure you were fighting does not come from `einvoicing`, it comes from your own module, so the two changes made to `einvoicing` (the `@phan-var` line and the `baseline.txt` entry) can both be dropped.

### Root cause

The CI runs phan on the **whole repository** (`--analyze-twice`), so phan keeps a single inferred type for each global variable across all files. In `dolisecure/class/actions_dolisecure.class.php:77`:

```php
global $user;
...
if (!isModEnabled('dolisecure') || empty($user) || empty($user->id)) {
```

the bare `empty($user)` teaches phan that the global `$user` can be falsy, and that union then leaks into every other file that relies on the standard header docblock. Hence the message on our home page:

```
einvoicing/einvoicingindex.php: PhanNonClassMethodCall: Call to method hasRight on non-class type
  ?''|?'0'|?0|?0.0|?array{}|?false|?non-empty-mixed
```

That type is literally the falsy complement of an `empty()` test — there is no `empty($user)` anywhere in `einvoicing`.

### Reproduced locally (phan 5.5.2, repo config + baseline, your branch merged with current `main`)

| state | result on `einvoicingindex.php` |
|---|---|
| `main` alone (control) | clean |
| your branch as submitted | clean, but only because of the `baseline.txt` entry |
| baseline entry removed, `/** @phan-var User $user */` kept | **still fails** |
| both `einvoicing` changes reverted, `empty($user)` removed in `dolisecure` | **clean** |

Two things worth knowing: `/** @var User $user */` and `/** @phan-var User $user */` written as a docblock have **no effect** here (phan does not attach a floating docblock to the next `if`/`global` statement), so the only thing that turned the CI green was the baseline entry — and that entry disables `PhanNonClassMethodCall` for the **entire** `einvoicing/einvoicingindex.php` file, for good. We would rather keep phan watching that file.

### Suggested fix (one line, in your module)

```diff
--- a/dolisecure/class/actions_dolisecure.class.php
+++ b/dolisecure/class/actions_dolisecure.class.php
@@ -74,7 +74,7 @@ class ActionsDolisecure
 
 		$this->resprints = '';
 
-		if (!isModEnabled('dolisecure') || empty($user) || empty($user->id)) {
+		if (!isModEnabled('dolisecure') || empty($user->id)) {
 			return 0;
 		}
```

`$user` is always a `User` object when a hook runs, so `empty($user->id)` alone already covers the "not logged in" case.

If you prefer to keep the defensive `empty($user)`, this also works and is the idiom already used elsewhere in this repo (a string statement, not a docblock):

```php
global $user;
'@phan-var User $user';
```

With either variant, and with `einvoicing/einvoicingindex.php` + `dev/tools/phan/baseline.txt` restored to their `main` content, phan comes back clean on my side (the only remaining findings are the pre-existing `helloasso/class/helloassomemberutils.class.php:519` ones, which are already there on plain `main`).

So: please revert the two `einvoicing`-related commits (`9b822d96`, `e3ea4fed`) and apply the one-liner above instead.

hansemschnokeloch:
@LaurentBouquet  can you remove the `.DS_Store `file, and exclude this kind of files in your local git config ?

LaurentBouquet:
Thanks @hansemschnokeloch I'll add .DS_Store in .gitignore file.

PR #633, opened 2026-08-24
FIX [einvoicing] The API call log silently loses rows: the next call number is read on the wrong connection

daGrumpf-bxp:
## The problem

`AbstractPDPProvider::logCall()` builds its `Call` on the independent `$dbhistory` and inserts there.
That connection is deliberate: it keeps the trace of an API call even when the caller's transaction is
rolled back (issue #291).

`Call::getNextCallId()` read the highest number through the **global `$db`**, and nothing held that
number between the read and the insert:

```php
public function getNextCallId()
{
    global $db;
    ...
    $sql = "SELECT MAX(CAST(SUBSTRING(call_id, ...)) AS maxref) FROM ..."
```

Two ways to hand out the same number, and the first one alone loses rows on a single request:

* **A stale read.** A page working inside a transaction on `$db` — recording a payment, importing a
  received invoice — holds a consistent-read snapshot taken before the previous call was committed by
  `$dbhistory`. The second API call of the request computes the number the first one already used.
* **No reservation.** Even read on the right connection, the number is not locked: another request
  numbering at the same moment takes it too.

Either way the insert dies on `uk_einvoicing_call_callid`. Nothing is logged, and the call it was
tracing becomes invisible.

### Seen in production

An invoice deposited at the Access Point during a payment left its flow id on the invoice and **no row
at all** in `llx_einvoicing_call`, with this in the log instead:

```
ERR DoliDBMysqli::query SQL Error message: DB_ERROR_RECORD_ALREADY_EXISTS
    Duplicate entry 'Call-000197-1' for key 'uk_einvoicing_call_callid'
```

The transmission that most needed auditing is the one the audit trail lost.

## The fix

The number is read **on the connection the record is written on** — `$this->db`, which *is* the global
`$db` everywhere else a `Call` is instantiated — with a **locking read** held until that insert commits:

```php
$sql .= " WHERE call_id LIKE '".$this->db->escape($prefix)."%'";
$sql .= " FOR UPDATE";
```

`FOR UPDATE` returns the last committed number instead of the caller's snapshot, and holds the range so
a concurrent request waits for the insert rather than taking the same number. `logCall()` keeps its
linear form: the transaction it already opened now covers the numbering, and no retry is needed.

The two halves are one fix, not two options. The lock only holds if the `SELECT` and the `INSERT` sit in
the same transaction — left on the global `$db` it would be taken by the caller's transaction (the
payment) while the insert happens on `$dbhistory`, protecting nothing and making the request wait on a
gap it locked itself.

## Tested

Sandbox devbrasserie, same bench on `main` and on this branch:

| bench | on `main` | on this branch |
| --- | --- | --- |
| two API calls in one request, a transaction open on `$db` as a payment does | the 2nd call recomputes the number of the 1st → `ErrorRefAlreadyExists`, **row lost** | distinct numbers, both rows inserted (`Call-000706`, `Call-000707` on 23.0.3 · `Call-000382`, `Call-000383` on 18.0.10) |
| concurrent numbering: process A takes a number and holds it uncommitted, process B calls the real `logCall()` | B collides on the unique index — `Duplicate entry 'Call-000971-1'`, the production error — and the trace is lost | B **waits 4.4 s on the lock**, then inserts the next number: `{"id":733,"call_id":"Call-000705"}`, no error at all |

The failure path was exercised too: a failure that is *not* a collision (`Data too long for column
'call_type'`) returns `null`, creates no row, and leaves exactly one log line carrying the reason.

Also run:

* the module's PHPUnit suite, file by file, on the **eight cores of the bench, Dolibarr 17 to 24**
  (15 files each): all green;
* one complete sandbox E2E cycle on the branch: invoice `TRI-FA-2608-0195` issued on the emitter, flow
  `i_352140`, received on the other instance as supplier invoice 2710, statuses 200 and 201 back — and
  every API call of the cycle logged, with contiguous numbers (`Call-001006` → `Call-001010` on the
  issuer);
* `phpcs` with the repository ruleset: clean.

Follows the same trail as the independent connection introduced for issue #291.


eldy:
Pb occurs because once an id is ready there is no lock before inserting the new one in the transaction.
Usual practice is to add FOR UPDATE after the select that read the next id.
So read of next id and insert are locked together with no way to have concurrential access between the two.

daGrumpf-bxp:
You are right, and it does more than tighten the fix: `FOR UPDATE` is a locking read, so it also
returns the last committed number instead of the caller's snapshot — which is the failure I actually
hit in production. Adopted, and the retry went with it: `logCall()` is back to its linear form, and the
change is now two lines in `getNextCallId()`.

One thing worth stating, because it is why the connection change stays in the PR: the lock only holds
if the `SELECT` and the `INSERT` are in the same transaction. The row is inserted on the independent
`$dbhistory` — deliberately, so the trace survives a rollback of the caller (#291) — so a `FOR UPDATE`
left on the global `$db` would take the lock in the caller's transaction, a payment for instance, while
the insert happens on another connection. It would protect nothing, and worse: `$db` would hold a
next-key lock on `llx_einvoicing_call` until the payment commits, while `$dbhistory` waits to insert in
that very gap — the request would block on itself until `innodb_lock_wait_timeout`. Reading on
`$this->db` is what makes the lock meaningful.

Measured on the sandbox with two processes: A takes a number and holds it uncommitted, B calls the real
`logCall()`. Before, B collided on the unique index (`Duplicate entry 'Call-000971-1'`, the production
error) and had to number again. Now B waits 4.4 s on the lock and returns
`{"id":733,"call_id":"Call-000705"}` — the error never happens. Full details and the rest of the run
(PHPUnit on the 8 cores, one E2E cycle) are in the description, which I rewrote.

One question while I am in there: you added `call_id_num` to `llx_einvoicing_call` in 7fba27e, and
nothing populates it yet — it is not in `Call::$fields`, so `create()` does not write it, and existing
rows are all NULL. If the intent is to number on it, `SELECT MAX(call_id_num) ... FOR UPDATE` would drop
the `CAST(SUBSTRING(...))` and could use an index, but it needs the field wired and the existing rows
backfilled (`UPDATE llx_einvoicing_call SET call_id_num = CAST(SUBSTRING(call_id, 6) AS SIGNED) WHERE
call_id_num IS NULL AND call_id LIKE 'Call-%'` in `update_v1.1.0.sql`), otherwise the numbering restarts
at 1 on every existing install. Do you want that in this PR, or separately?

eldy:
Yes the call_id_num is to prepare to cancel of subtr and cast is the selected max 
However, for the moment, everything works correctly and there is no need tonprogressbon this direction.
This column is added now to je sure that when we will need itnin a long time in the future ,everybody has this field ready. It will need migration, code change but for the moment this is not a good idea to do it as we are in einvoicing enrolment and first use step.
No need to progress pn this for the moment, we will see in several month...

PR #634, opened 2026-08-24
add stancer module

rycks:
Here is the stancer module back as it was into that repository some month ago (time with remote git repository) but now according to dolibarr wiki rules all source code is in that PR but without history :-(


rycks:
please note i can't apply isModEnabled because this module can be used on dolibarr 18 where isModEnabled do not handle  facture,(invoice) commande(order), adherent(member), banque(bank)

PR #639, opened 2026-08-25
FIX [einvoicing] The Request-Id is sent as a query parameter, where the API expects a header

daGrumpf-bxp:
### What

Send `Request-Id` as the HTTP header the flow API declares, instead of a query parameter, and keep the identifier in the call log.

### Why

The API declares `Request-Id` as a **header** on every operation — `createFlow`, `searchFlows`, `geFlowByUid`, `getHealth`:

```yaml
      operationId: createFlow
      parameters:
        - name: Request-Id
          in: header
          description: Header parameter used to correlate logs from several components
```

Both providers built it into the query string instead:

```php
$resource = 'flows';
$urlparams = array('Request-Id' => $uuid);
$resource .= '?' . http_build_query($urlparams);
```

Sent that way the platform simply ignores it, so the correlation between our logs and the ones of the Access Point — the only purpose of that identifier — was never established. On the invoice submission it was worse: the resource was built and then dropped, the call being made on the literal `'flows'`.

Each `generateUuidV4()` also carried a `TODO : Store it somewhere`. It is now stored: `callApi()` hands the identifier to `logCall()`, which records it next to the call, so a support request can be matched with the platform logs afterwards.

`docType`, which the API does declare as a query parameter, is left untouched.

### Changes

| file | change |
|---|---|
| `SuperPDPProvider`, `EsalinkPDPProvider` | `Request-Id` moved to `$extraHeaders`, on the invoice submission and on `flows/search` |
| `AbstractPDPProvider::logCall()` | new `$requestId` parameter, filled by `callApi()` from the headers |
| `call.class.php` | `request_id` field and property |
| `sql/llx_einvoicing_call.sql`, `sql/update_v1.1.1.sql` | `request_id varchar(36)` column, for new and existing installations |
| `langs/en_US/einvoicing.lang` | `RequestId` key |

### Before / after, read in the call log itself

```
before : endpoint=/flows?Request-Id=9bd5b54e-8aab-4308-9e28-2166061d2408   request_id=(empty)
after  : endpoint=/flows                                                    request_id=5a620969-df1c-4e81-9854-cd227f0b3b4d
```

### How it was tested

Full lifecycle on a sandbox, both ways, on the oldest and the newest supported version — a new invoice is issued, routed by the Access Point, imported by the other instance, approved, paid, and cashed in:

| cycle | lifecycle statuses received |
|---|---|
| Dolibarr 18 → Dolibarr 23 | 200, 201, 205, 211, 212 |
| Dolibarr 23 → Dolibarr 18 | 200, 201, 205, 211, 212 |
| Dolibarr 24 → Dolibarr 23 | 200, 201, 205, 211, 212 |
| Dolibarr 23 → Dolibarr 24 | 200, 201, 205, 211, 212 |

Every submission and every `flows/search` of those runs was accepted, and each one recorded its identifier in `request_id`.

`php -l`, `phpcs` with the repository ruleset and phan with the CI configuration are clean on the branch.

Note on coverage: the sandbox in use is a SuperPDP one, so the Esalink half of the change was not exercised against a live platform. The change is the same on both providers and follows the API contract quoted above, which is the Esalink specification itself.


daGrumpf-bxp:
@eldy heads-up on the `sql/update_v1.1.1.sql` that was dropped before the merge: without it, existing installations do not get the `request_id` column, and they lose their whole API call log.

`_load_tables()` only creates the tables it does not find, so `llx_einvoicing_call.sql` covers a fresh installation only. On an installation that already carried the table, `Call::create()` inserts a column that is not there, and the insert fails — silently, because the trace is written on its own connection and its error only reaches `dol_syslog()`.

Measured on such an instance, on current `main`:

```
http=200   call log rows before=805 after=805   => LOST
```

The invoice is still sent; only the trace that it was sent disappears, which is the one thing an operator looks at afterwards.

If you would rather not have a new `update_*.sql` file, the repair works just as well in PHP: #642 puts it in `init()`, next to the repairs already there for old installations. It reads the table first (`DDLDescTable()`), adds the column only if missing, and is a no-op afterwards — measured: column added on the first pass, nothing on the second, and the call log records again right after.

eldy:
> @eldy heads-up on the `sql/update_v1.1.1.sql` that was dropped before the merge: without it, existing installations do not get the `request_id` column, and they lose their whole API call log.
> 
> `_load_tables()` only creates the tables it does not find, so `llx_einvoicing_call.sql` covers a fresh installation only. On an installation that already carried the table, `Call::create()` inserts a column that is not there, and the insert fails — silently, because the trace is written on its own connection and its error only reaches `dol_syslog()`.
> 
> Measured on such an instance, on current `main`:
> 
> ```
> http=200   call log rows before=805 after=805   => LOST
> ```
> 
> The invoice is still sent; only the trace that it was sent disappears, which is the one thing an operator looks at afterwards.
> 
> If you would rather not have a new `update_*.sql` file, the repair works just as well in PHP: #642 puts it in `init()`, next to the repairs already there for old installations. It reads the table first (`DDLDescTable()`), adds the column only if missing, and is a no-op afterwards — measured: column added on the first pass, nothing on the second, and the call log records again right after.
> 

It has been moved into update_v1.1.0 because release 1.1.0 was not yet published.

daGrumpf-bxp:
Confirmed, and my reading of `_load_tables()` was wrong: it has a fifth pass for `update*.sql` (`substr($file, 0, 6) == 'update'`), present in every core from 17 to 24, so the `ALTER` you moved into `update_v1.1.0.sql` runs on each module enable and does reach the installations that predate the column.

Verified live on 18.0.10, 20.0.4, 22.0.5, 23.0.3 and 24.0.0 — table with `request_id` dropped, probe `update_*.sql` replaying that `ALTER`, real disable/enable cycle:

```
  update_*   : EXECUTED (probe table created)
  request_id : ADDED BACK
  2nd enable : ok, no error (column already present)
```

The second pass is a no-op because `run_sql()` accepts `DB_ERROR_COLUMN_ALREADY_EXISTS` by default. Nothing else needed here, and #642 is closed.

Issue #644, opened 2026-08-26
[einvoicing] EINVOICING_THIRDPARTIES_COMPLETE_INFO Option does not function well if activated


Delthair:
### Bug

On any version, if you activate this option the code does not do anything : 

		// if found, update information
		if ($thirdpartyId > 0) {
			// if complete info is disabled, we return directly the thirdpartyId
			if (getDolGlobalInt('EINVOICING_THIRDPARTIES_COMPLETE_INFO')) {
				dol_syslog(get_class($this) . '::_syncOrCreateThirdpartyFromEInvoiceSeller Complete info disabled, returning existing thirdparty: ' . $thirdpartyId);
				return array(
					'res' => $thirdpartyId,
					'message' => 'Existing thirdparty used without update: ' . $thirdpartyId . ($nameMismatchWarning !== '' ? ' - ' . $nameMismatchWarning : '')
				);
			}

the return array exits the code. 

Proposition is to delete this piece of code and to put the if around the block

			// Update thirdparty information based on priority

### Module

einvoicing

### Steps to reproduce the behavior and expected behavior

If you activate the option and that a client is becoming also a supplier, you will receive the invoice from the company but the company card is not updated with supplier field put to 1 because the code exits too early.

PR #645, opened 2026-08-26
fix ISSUE #644

Delthair:
Fix ISSUE #644


eldy:
@daGrumpf-bxp Do you know why on this PR, some phan/phpstan errors are reported on non modified files when nothing is reported on your PR (even if both does not modify the files the phan is reported on non modified files, we got a different result) ?

daGrumpf-bxp:
@eldy It is not a CI difference: this PR does not parse, and everything else is fallout.

```
$ php -l einvoicing/class/protocols/CommonProtocol.class.php
PHP Parse error:  syntax error, unexpected token "if", expecting "function" on line 776
```

The opening `if ($thirdpartyId > 0) {` was removed, but its closing brace further down was kept. The orphan brace closes `_syncOrCreateThirdpartyFromEInvoiceSeller()` early, so the rest of the method ends up directly in the body of the trait, hence the parse error.

And because `CommonProtocol` is a **trait**, once it fails to parse phan cannot build it at all. Every class doing `use CommonProtocol` then loses all of its methods, which is exactly the cascade you are seeing on the files the PR does not touch:

```
CommonProtocol.class.php: PhanSyntaxError                    <- the only real error
  -> CIIProtocol / FacturXProtocol: PhanUndeclaredTrait
     -> PhanUndeclaredMethod x20 (_syncOrCreateThirdpartyFromEInvoiceSeller,
        getIEC6523Code, _tsToDateTime, getBillingProcessID, ...)
        in buildinvoicelines.inc.php, product_mapping.php, ...
```

So: one root cause, about twenty symptoms. Nothing is reported on my PRs simply because my files parse. Same story for the PHPCS run: the ~40 `Line indented incorrectly` are on the block this PR re-indented.

(There *is* a separate effect that can legitimately report phan errors on non-modified files, and it may be what you had in mind: the workflow analyses the whole repository with `--analyze-twice`, and phan keeps a single inferred type per global variable across all modules, so e.g. an `empty($user)` in another contributor's module can surface a `PhanNonClassMethodCall` inside `einvoicing/`. But that is not what is happening here.)

---

On the substance, #644 is right: the option is inverted. The label says *"Complete missing third-party information when receiving an e-invoice"*, but the code returns immediately without updating anything **when the option is enabled** (the log line even says "Complete info disabled" while testing the opposite).

The minimal fix is to keep the `if ($thirdpartyId > 0) {` guard, drop the early return, and wrap only the priority chain:

```php
		// if found, update information
		if ($thirdpartyId > 0) {
			dol_syslog(get_class($this) . '::_syncOrCreateThirdpartyFromEInvoiceSeller Updating existing thirdparty: ' . $thirdpartyId);
			// TODO: MAYBE we should call PDP to retrieve more information

			$thirdparty = new Societe($db);
			$thirdparty->fetch($thirdpartyId);

			// Complete the third party information from the e-invoice only when the option asks for it.
			if (getDolGlobalInt('EINVOICING_THIRDPARTIES_COMPLETE_INFO')) {
				// Update thirdparty information based on priority
				if ($priority === 'pdp') { // Overwrite Dolibarr data with AP data
					...
				} elseif ($priority === 'dolibarr') { // Fill only empty fields from pdp data
					...
				}
			}

			// Flag the thirdparty as a vendor if it is not one yet
			...
```

The `fournisseur = 1` flag and the `update()` stay outside the option, which is what the issue asks for (a customer receiving its first supplier invoice must still be flagged as a vendor).

I rebuilt it that way on top of current `main` and ran the checks locally: `php -l` clean, PHPCS clean on the file, and a full local replay of the phan workflow reports no einvoicing error left (only the two pre-existing `helloasso/class/helloassomemberutils.class.php:519` ones). Happy to push it as a PR if @Delthair prefers.

daGrumpf-bxp:
@Delthair @eldy I have pushed the same fix with the brace restored, in #663, so the file parses again and the phan cascade goes away. It keeps your idea exactly — the early return removed, the option moved around the priority chain — plus the `if ($thirdpartyId > 0)` guard and the vendor flag left outside the option.

It comes with the bench run: the bug reproduced on `main` first, then the fix checked on Dolibarr 18 to 24, the error path, PHPUnit 16/16 on four cores and one full sandbox cycle. Happy to close #663 if you would rather fix this one, the diff is small either way.

Delthair:
Hi @eldy , @daGrumpf-bxp  I have corrected my PR, was a good elarning exercice, i think all checks out now, can you confirm ?

daGrumpf-bxp:
@Delthair @eldy Confirmed, and thanks for going back to it. Your file now parses:

```
$ php -l einvoicing/class/protocols/CommonProtocol.class.php
No syntax errors detected
```

Comparing the two resulting files, this PR and #663 are the same fix: apart from three comment and
blank lines, they are byte for byte identical - same `if ($thirdpartyId > 0)` guard kept, same
priority chain wrapped in `EINVOICING_THIRDPARTIES_COMPLETE_INFO`, and the vendor flag plus the
`update()` left outside the option so a customer receiving its first supplier invoice is flagged as a
vendor either way.

So there is nothing left to arbitrate between them: I am closing #663 in favour of this one, which
came first. Its description keeps the write-up of why the option read backwards and why the vendor
flag went with it, for whoever comes across the issue later.

One thing worth knowing before this lands, whichever PR carries it: `EINVOICING_THIRDPARTIES_COMPLETE_INFO`
is not seeded at module activation - it is not in `$this->const` of `modEInvoicing`, and no SQL file
inserts it - so on every existing installation the constant is simply absent, i.e. 0. Today, with the
test reading backwards, absent means the completion **runs**. Once the condition is put the right way
round, absent means it **stops**. Every installation that has never opened that setup option will
silently change behaviour on upgrade. Adding it to `$this->const` with value `1` would keep the
current de facto default; that is a product call, not a code one, so I am only flagging it.

daGrumpf-bxp:
@Delthair @eldy Confirmed: `php -l` passes now, and the resulting file is identical to the one in
#663 apart from three comment lines. I have closed #663 in favour of this one.

Worth knowing before merge: `EINVOICING_THIRDPARTIES_COMPLETE_INFO` is a GUI option defaulting to 0
and is never seeded, so on any installation that never opened that setup page the completion stops
running once the condition is put the right way round. That is the intended behaviour.

PR #650, opened 2026-08-26
ENH [einvoicing] Emit additional order references in CII (multi-order invoicing)

lelex86:
## Problem

When an invoice covers several purchase orders (Dolibarr multi-order
billing), the module collects all the linked orders' `ref_client` values
into `_customerOrderReferenceList`. The first reference goes into BT-13
(`BuyerOrderReferencedDocument`), but the remaining ones were never
written to the XML — they were silently dropped.

## Fix

After `ContractReferencedDocument` and before `SpecifiedProcuringProject`
(CII schema sequence), emit one `ram:AdditionalReferencedDocument` per
additional order reference, with `ram:TypeCode` = `130`.

Guards:
- skipped in Chorus mode (which does not accept these nodes)
- skipped for `MINIMUM` profile (which does not carry
  `AdditionalReferencedDocument` in the agreement section)
- the reference already emitted as BT-13 is excluded to avoid
  duplication

## Impact

- **Single-order or no-order invoices**: no change — the loop body
  never executes.
- **Multi-order invoices**: additional order references now appear
  in the XML instead of being lost.

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -2088,6 +2088,24 @@ class CIIProtocol extends AbstractProtocol
 			$contractRef->appendChild($doc->createElement('ram:IssuerAssignedID', htmlspecialchars($invoiceData['contractReference'])));
 		}
 
+		// Additional order references: when an invoice covers several purchase orders, the first is
+		// emitted as BT-13 (BuyerOrderReferencedDocument above) and the others are listed here as
+		// AdditionalReferencedDocument/TypeCode=130 — the same approach Factur-X uses. Restricted to
+		// profiles that carry AdditionalReferencedDocument in the agreement section (not MINIMUM), and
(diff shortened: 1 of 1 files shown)


daGrumpf-bxp:
In the default configuration Factur-X does not emit these either: `FacturXProtocol::generateXML()` returns `parent::generateXML()` (the CII builder) unless the deprecated `EINVOICING_USE_EXTERNAL_FACTURX_BUILDER` is set, and the `addDocumentAdditionalReferencedDocument($ref, "130")` call sits inside that dead `else` branch.

So this is not a CII/Factur-X divergence, it is a reference that nothing emits any more — the fix itself is right, only the rationale needs rewording.

A follow-up PR is coming to delete that unreachable horstoeko builder, so the two paths cannot drift apart again.

lelex86:
You are right on the rationale — the Factur-X reference was misleading.
PR description updated to reflect what the fix actually does: the
additional order references in _customerOrderReferenceList were never
written to the XML in any configuration. This addresses that gap
directly in the CII builder.

PR #651, opened 2026-08-26
CHANGE [dolisecure] Add this module in index.yaml

LaurentBouquet:
Registers the DoliSecure module in index.yaml so it appears in the Dolibarr marketplace (/admin/modules.php?mode=marketplace) and can be installed directly from any Dolibarr instance, without manual download/upload of the module zip.

DoliSecure checks the running Dolibarr version against known CVEs and reports vulnerability status to the admin (dashboard widget + scheduled check), helping users keep their installation secure.

This change only adds the marketplace listing entry — no core Dolibarr code is modified.


eldy:
I remove change of index.yaml file. This is not required. 
File is generated during sync phases from the index.yaml into dolisecu dir. You just need to wait.

LaurentBouquet:
OK @eldy, I'll wait. Thanks.

PR #653, opened 2026-08-27
FIX: #354 [einvoicing] Recording the e-invoice status no longer cancels the invoice

### LINKED ISSUE #354: [EInvoicing] erreur à la génération de factures par tache planifié depuis un modèle de facture

daGrumpf-bxp:
Fixes #354 (the trigger half — see the note at the end for the other half).

## History

This revives #465, which proposed the same change for `BILL_CREATE` alone and was closed on 2026-07-28. The review there objected that aborting is the right behaviour, because an invoice created without its row in `llx_einvoicing_extlinks` would be left orphaned and break the workflow. Two things have been established on a bench since, and both bear on that objection:

- **Nothing downstream needs the row.** Generation and transmission are gated by `mustManageEInvoice()`, read from the invoice itself; the last known status is consulted only to skip the invoices explicitly marked as ignored, which a missing row is not. The e-invoice is still produced, still sent, and `generateInvoice()` writes the status on its way. Such an invoice is not orphaned.
- **Aborting protects nothing, it only blocks.** With a schema left behind, the installation cannot create an invoice at all — and, with the creation alone fixed, cannot validate one either, because the invoice reaches `BILL_VALIDATE` without its row and takes the branch that writes it. That is why both paths are covered here and #465 was not enough.

If the maintainers still prefer the operation to be refused, this can be closed again — but then the refusal should at least say what to do, because today it says `Unknown column`.

## Problem

On `BILL_CREATE` and on `BILL_VALIDATE` the module records the e-invoice status of the invoice in `llx_einvoicing_extlinks`, and returned `-1` when that write failed:

```php
$result = $einvoicing->setEInvoiceStatus($object, $statustouse, '');
if ($result < 0) {
    $this->errors = array_merge($this->errors, $einvoicing->errors);
    return -1;
}
```

A trigger returning `-1` rolls back the operation, so failing to record a status destroys the invoice the core had just created, or refuses a validation.

That is what #354 reports: on an installation whose module files were updated to 1.0.3 without the schema being migrated, `ap_precheck_status` does not exist, the `INSERT` fails, and no invoice can be created at all. It is not specific to the scheduled generation of recurring invoices, which is only where the reporter met it — every creation path goes through this trigger, and the only workaround is to disable the module.

## Fix

Both failures are logged and reported as a warning; the invoice is created, and validated.

The two go together. Fixing only the creation moves the blockage one step: an invoice created without its status row arrives at `BILL_VALIDATE` with `STATUS_UNKNOWN`, takes the branch that writes the row, fails there too — so the installation could create invoices but never validate one. Verified on the bench, see below.

Nothing downstream depends on those writes. Generation and transmission are gated by `mustManageEInvoice()`, read from the invoice itself; `fetchLastknownInvoiceStatus()` is consulted only to skip the invoices explicitly marked as ignored, and a missing row is not one of them. So the e-invoice is still produced and still sent, and `generateInvoice()` writes the status on its way. This is the rule the payment statuses in the same file already follow: a bookkeeping write of the module has no business cancelling an accounting document of the core.

## Reproduced and verified on a bench

Dolibarr 21.0.4, `llx_einvoicing_extlinks` put back in its pre-1.0.3 shape (both `ap_precheck_*` columns dropped).

On the exact path of the ticket — a recurring invoice template generated the way the scheduled job does:

| | `main` | this branch |
|---|---|---|
| `createRecurringInvoices()` | 1 error, `Unknown column 'ap_precheck_status' in 'INSERT INTO'` | no error |
| `llx_facture` | unchanged | the invoice is created |

Direct creation and validation, same schema:

| | `main` | creation fixed only | this branch |
|---|---|---|---|
| create | `KO res=-4`, nothing created | `OK` | `OK` |
| validate | not reached | `KO res=-1`, invoice stays draft | `OK`, invoice validated |

With the fix the reason is in `documents/dolibarr_einvoicing.log`:

```
ERR InterfaceEInvoicingTriggers::runTrigger Failed to record the e-invoice status of invoice
id=21. The invoice is created anyway. Error: Unknown column 'ap_precheck_status' in 'INSERT INTO'
```

Schema restored afterwards: the status row is written at creation again and validation is unchanged. Test invoices removed, bench back to its initial state. phpunit green on Dolibarr 18.0.10, 21.0.4, 23.0.3 and 24.0.0-beta (152 tests).

## What this does not fix

- A stale schema still makes `fetchLastknownInvoiceStatus()` fail on its `SELECT`, and the core prints its database error page around the operation. The operation itself now completes, but the screen is ugly until the schema is migrated.
- On PostgreSQL a failed statement poisons the whole transaction, so the creation still fails, one statement later. That is no worse than today.
- The schema gap itself. The cure for that is to deactivate and reactivate the module: `_load_tables()` replays the `update_*.sql` files of a module on every activation, and `sql/update_v1.0.3.sql` carries the `ALTER TABLE` statements that add the two columns. Verified on the same bench. This PR is about what happens when a schema is behind anyway - warn, instead of destroying the invoice.


daGrumpf-bxp:
Closing. The decision taken on #465 stands: the trigger refuses the operation rather than letting it through with a warning, and the schema is what has to be put right.

Leaving the reason here so this is not proposed a third time: an installation whose schema is behind is repaired by deactivating and reactivating the module, which replays `sql/update_v1.0.3.sql` and its `ALTER TABLE` statements. Verified on a bench on Dolibarr 21.0.4.

PR #654, opened 2026-08-27
FIX [einvoicing] A replacement invoice specimen carries no reference to the invoice it replaces

daGrumpf-bxp:
A replacement invoice (BT-3 = 384) references the invoice it corrects in the same slot a credit note uses, and **BR-FR-CO-04 makes that reference mandatory with a \"fatal\" flag**: one sent without it is refused by the access point, and a receiver has nothing to attach it to. `buildinvoicelines.inc.php` fills that slot for both types.

`CommonProtocol::generateSampleInvoice()` fills it for a credit note only. A specimen of a replacement invoice therefore comes out in the one shape the norm does not accept — right document type code, no reference. Measured on Dolibarr 24, same instance, the only difference being this one condition:

| | `<ram:TypeCode>` | `<ram:InvoiceReferencedDocument>` |
|---|---|---|
| before | 384 | **absent** |
| after | 384 | `FA0000-SPECIMEN-STANDARD` |

Nobody had run into it because the page that generates the samples did not offer the type: `admin/setup_devtools.php` listed standard, deposit and credit note. It now offers the replacement too, and shows it the field for the invoice being replaced.

## The fixture that holds it

Three of the six invoice types of Dolibarr had a reference fixture. The fourth is added here, so `EInvoicingSamplesTest` holds the XML of a replacement invoice like the others — and `FacturxPdfaContainerTest` (#592) picks it up for free, since it reads the whole fixture directory: eight documents instead of six, and the pdfa job of #595 validates all eight if it lands.

The two types still without one are the situation invoice, whose `situation_percent` and BT-25/26 chain no test exercises, and the proforma, which is parsed on reception and never emitted. Neither is in this PR.

## Verified

Run on a bench of Dolibarr instances, module symlinked into `htdocs/custom`:

- the fixture is the output of `scripts/regenerate_einvoicing_fixtures.php` on Dolibarr 24; **the three existing fixtures come back byte-identical** on the same instance, before and after this change;
- `EInvoicingSamplesTest`: OK, 4 tests, 8 assertions — on Dolibarr 20, 23 and 24;
- `FacturxPdfaContainerTest`: OK, 8 tests, 224 assertions;
- veraPDF 1.30.2 `-f 3b` over the eight documents those tests produce: **8/8 PASS**.

On Dolibarr 17 the sample test errors on `Undefined property: Conf::$einvoicing`, identically on `main` and on this branch — an instance where the module is not enabled, not something this change touches.

## It merges with #655 either way

#655 adds the situation invoice to the same four lists. The two sets of entries sit apart — the replacement next to the standard invoice it replaces, the situation at the end — so the merge is clean whichever lands first: `git merge-tree` reports no conflict in either order, and both orders produce the same tree. On that merged state, regenerated on Dolibarr 24: the five fixtures come back byte-identical, `EInvoicingSamplesTest` is OK on 5 tests, `FacturxPdfaContainerTest` OK on 10, and veraPDF passes the ten documents. Neither PR needs the other, and neither needs a rebase.


daGrumpf-bxp:
Ran the pdfa job of #595 against this branch, to check the container of the type it adds. It cannot run here — `pdfa.yml` is not on `main` yet — so it was dispatched on my fork, on `this branch + ci/pdfa-workflow`: [run](https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/33056471364/job/98464402700), green in 25 s.

```
OK (8 tests, 224 assertions)
PASS ctcfr_cii_creditnote / cii_deposit / cii_replacement / cii_standard
PASS tcpdf_cii_creditnote / cii_deposit / cii_replacement / cii_standard
control built from ctcfr_cii_creditnote.pdf
FAIL control.pdf 3b   -> veraPDF refuses the control document, as it has to.
```

Eight documents instead of six, with no line of workflow to touch: the job reads the whole fixture directory, so `cii_replacement.xml` is picked up on its own. Neither PR depends on the other.

PR #655, opened 2026-08-27
QUAL [einvoicing] Hold the XML of a situation invoice with a reference fixture

daGrumpf-bxp:
The share of the work a situation invoice bills modulates every amount of the line: `situation_percent` narrows the unit price the discount already narrowed (`buildinvoicelines.inc.php:538`), and it is the progress `calcul_price_total()` is called with (`:545`). Nothing exercises a line of it — on a computation the module has already been wrong about once, #505.

A fifth sample covers it.

## What the specimen is

The **first situation of its cycle**. That is the shape which needs no predecessor to exist: BT-25/26, the reference to the situation before, only appears from the second one on (`buildinvoicelines.inc.php:311`) and would mean reading an invoice out of a database — which a specimen does not do. So this covers the arithmetic, and not the chain. It does not pretend otherwise.

One line, 5 × 100.05, 10 % discount, 50 % progress:

| | |
|---|---|
| `LineTotalAmount` (BT-131) | 225.11 = 450.23 × 50 % |
| `TaxBasisTotalAmount` | 225.11 |
| `GrandTotalAmount` | 270.14 |

## It changes nothing for the other types

The progress is passed to `calcul_price_total()` where it was left implicit. It defaults to 100, which is what the call already meant, and **the three existing fixtures come back byte-identical** — regenerated on the same instance, before and after.

## One thing the fixture now shows plainly

The progress does not reach the line. BT-146 × BT-129 less the allowance is 450.22, where BT-131 says 225.11: a reader cannot recompute the line from what the document gives him. `$line_unit_price_with_discount` is computed with the progress and then never used — the assignment that would have carried it is commented out at `buildinvoicelines.inc.php:615`, in favour of the gross price plus a `SpecifiedTradeAllowanceCharge`, which expresses the discount and has nothing to express the progress.

No EN 16931 rule ties those three together: the schematron shipped in `vendor/` has BR-CO-10 and BR-CO-13 on the sums, and nothing on the line itself. So the document is conformant, and the share it bills is nowhere in it.

**Left exactly as it is in this PR.** Changing it is a decision about how progress should be modelled — reduce BT-129, or narrow BT-146 — not something a fixture should settle. The fixture makes the current answer visible, which is the first step to changing it if you want to.

## Verified

On a bench of Dolibarr instances, module symlinked into `htdocs/custom`:

- the fixture is the output of `scripts/regenerate_einvoicing_fixtures.php` on Dolibarr 24;
- `EInvoicingSamplesTest`: OK, 4 tests, 8 assertions — on Dolibarr 20 and 24;
- `FacturxPdfaContainerTest`: OK, 8 tests, 224 assertions, since it reads the whole fixture directory;
- veraPDF 1.30.2 `-f 3b` over the eight documents: 8/8 PASS.

## It merges with #654 either way

#654 adds the replacement invoice to the same four lists. The two sets of entries were moved apart so the merge is clean whichever lands first: `git merge-tree` reports no conflict in either order, and both orders produce the **same tree**. On that merged state, regenerated on Dolibarr 24: the five fixtures come back byte-identical, `EInvoicingSamplesTest` is OK on 5 tests, `FacturxPdfaContainerTest` OK on 10, and veraPDF passes the ten documents. Neither PR needs the other, and neither needs a rebase.


daGrumpf-bxp:
Ran the pdfa job of #595 against this branch, to check the container of the type it adds. It cannot run here — `pdfa.yml` is not on `main` yet — so it was dispatched on my fork, on `this branch + ci/pdfa-workflow`: [run](https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/33056474234/job/98464412571), green in 36 s.

```
OK (8 tests, 224 assertions)
PASS ctcfr_cii_creditnote / cii_deposit / cii_situation / cii_standard
PASS tcpdf_cii_creditnote / cii_deposit / cii_situation / cii_standard
control built from ctcfr_cii_creditnote.pdf
FAIL control.pdf 3b   -> veraPDF refuses the control document, as it has to.
```

Eight documents instead of six, with no line of workflow to touch: the job reads the whole fixture directory, so `cii_situation.xml` is picked up on its own. On the merged state of this PR and #654 it is ten, all PASS.

PR #659, opened 2026-08-27
FIX: #658 [einvoicing] A Factur-X no longer embeds its own file name instead of the XML

### LINKED ISSUE #658: [EInvoicing] A Factur-X generated while the invoice has no PDF embeds the file path instead of the XML
## Symptom

When the e-invoice of an invoice is generated while the invoice has **no PDF yet**, the Factur-X file that comes out carries, as its embedded XML, **the path of the temporary XML file** instead of the XML itself:

```
FA2608-0044_facturx.pdf -> [/var/www/dd18/documents/facture/temp/FA2608-0044/factur-x.xml]
```

61 bytes of file name where 8 KB of CII should be. The module reports "E-invoice generated successfully", the status goes to *Generated*, and the file is offered for transmission. Nothing downstream can read it: it is not a Factur-X document, only a PDF with a text attachment named `factur-x.xml`.

On **Dolibarr 24** the same scenario fails loudly instead, because the merger used there validates the XML before embedding it:

```
The e-invoice XML cannot be embedded into the PDF <ref>.pdf: The XML does not match
the requirements for an XML in CII-Syntax. No Factur-X file was produced.
```

and the error path then deletes the Factur-X file that had been produced a moment earlier, so the invoice ends up with nothing at all.

## Reproduction

On any instance with `EINVOICING_PROTOCOL = FACTURX`:

1. take a validated invoice and delete the PDF files of its directory (`documents/facture/<ref>/*.pdf`), which is the state of an invoice validated before the module was set up, of one whose document was removed, or of an invoice reached by a mass generation;
2. from the invoice card, click **Generate the e-invoice**;
3. read what the produced `<ref>_facturx.pdf` carries:

```php
require 'einvoicing/vendor/autoload.php';
echo horstoeko\zugferd\ZugferdDocumentPdfReaderExt::getInvoiceDocumentContentFromFile('<ref>_facturx.pdf');
```

Observed on the bench, same invoice, same steps:

| Dolibarr | Result |
|---|---|
| 18.0.10 | "generated successfully", embedded XML = 61 bytes, the path |
| 23.0.3 | "generated successfully", embedded XML = 67 bytes, the path |
| 24.0.0-beta | generation fails, no file produced |

Generating the same invoice **after** its PDF has been rebuilt works everywhere and embeds the real 7.7 KB XML, which is why this has stayed invisible: every ordinary flow (validation, then generation) has the PDF in place.

## Mechanism

`FacturXProtocol::generateInvoice()` writes the XML to `documents/facture/temp/<ref>/factur-x.xml`, **then** looks for the source PDF. When there is none it calls `generateDocument()` to rebuild it — and that fires the `afterPDFCreation` hook, which generates the e-invoice on its own, merges it, and cleans up that very temporary XML when it is done. The outer call then merges against a path that no longer points at anything.

The two mergers take the XML *as content* and fall back to treating the string as content when the path does not exist, so what gets embedded is the path. The guard right before the merge only checks the PDF:

```php
if (!file_exists($orig_pdf)) {
    throw new \Exception("XML and/or PDF does not exist");
}
```

The message already mentions the XML; the test does not.

Log of one run, Dolibarr 24, the nested generation and the cleanup between the two merges:

```
FacturXProtocol::generateInvoice                                   <- outer call, XML written
ActionsEInvoicing::afterPDFCreation Hook afterPDFCreation called   <- the rebuilt PDF fires the hook
FacturXProtocol::generateInvoice                                   <- nested call
FacturXProtocol::generateInvoice copied original PDF to ..._facturx.pdf
FacturXProtocol::generateInvoice cleaned up temporary XML file: .../temp/<ref>/factur-x.xml
FacturXProtocol::generateInvoice copied original PDF to ..._facturx.pdf   <- outer call resumes
FacturXProtocol::generateInvoice cannot embed the XML into <ref>.pdf : The XML does not match ...
```

## Impact

A file that is announced as a valid Factur-X, accepted by the module, and transmittable, while carrying no invoice data. An Access Point that checks the attachment refuses it; one that does not, forwards an empty invoice to the customer.

daGrumpf-bxp:
Fixes #658.

## Problem

`FacturXProtocol::generateInvoice()` writes the XML to `documents/facture/temp/<ref>/factur-x.xml`, **then** looks for the source PDF. When the invoice has none it rebuilds it with `generateDocument()` — and that fires the `afterPDFCreation` hook, which generates the e-invoice on its own, embeds it, and cleans up that very temporary XML when it is done. The outer call then merges against a path that no longer points at anything.

The two mergers take the XML *as content*, and fall back to treating the string as content when it is not the path of an existing file. So what got embedded was the path:

```
FA2608-0044_facturx.pdf -> [/var/www/dd18/documents/facture/temp/FA2608-0044/factur-x.xml]
```

61 bytes of file name where 8 kB of CII should be, in a file the module announces as generated, and offers for transmission. On Dolibarr 24 the merger validates the syntax first, so the generation failed outright instead — and its error path then deleted the correct Factur-X the nested generation had produced a moment earlier.

The guard right before the merge only tested the PDF, although its message already mentioned the XML.

## Fix

Two lines of substance:

- when the PDF rebuild consumed the temporary XML, write it again — the same `generateXML()` call as at the top of the method, so the same CII builder, not a second path;
- test the XML in the guard, so a missing one can never be embedded again.

## Why this path is reached in real life

The module put that PDF rebuild there on purpose — its own comment says *"Source PDF deleted or never generated: regenerate it with the default PDF model before embedding"*. Invoices reach it: those validated before the module was installed, those whose documents were purged, those validated with `MAIN_DISABLE_PDF_AUTOUPDATE` on, and every invoice of a mass generation (#657) that never had its document built.

## Tests

- **Reproduced and fixed on four cores.** On a clean invoice with no PDF, generating the Factur-X from the invoice card:

| Dolibarr | before | after |
|---|---|---|
| 18.0.10 | "generated", embedded XML = 61 bytes, the path | 8358 bytes, document `DD18-FA-2608-0018` |
| 22.0.5 | "generated", the path | 8004 bytes, document `DD22-FA-2608-0010` |
| 23.0.3 | "generated", 67 bytes, the path | 8112 bytes, document `DD23-FA-2608-0025` |
| 24.0.0-beta | generation fails, no file | 8115 bytes, document `DD24-FA-2608-0011` |

- **Nothing else moved.** Full generation matrix on 8 instances (17.0.4 → 24.0.0-beta), every validated invoice of each through the 6 profiles and both protocols — **7902 generations per run**, payload hashed for each artefact (the XML, or the XML extracted from the Factur-X PDF): **7056 payloads byte-identical between `main` and this branch**, identical generation reports.
- **phpunit** on the 9 instances: 152 tests green, except the 3 pre-existing `EInvoicingSamplesTest` errors on Dolibarr 17.
- **End to end, both directions, dolidev (23.0.3) ↔ dd18 (18.0.10)**, through the SuperPDP sandbox, in CII and in Factur-X: send, sync at the receiver, import, validate, return status 205. Four legs, all acknowledged (202 on import, 205 after validation), and the two Factur-X files transmitted carry their real XML (8443 and 8362 bytes, right invoice numbers).

### DIFF
--- a/einvoicing/class/protocols/FacturXProtocol.class.php
+++ b/einvoicing/class/protocols/FacturXProtocol.class.php
@@ -730,6 +730,16 @@ class FacturXProtocol extends CIIProtocol
 				return -1;
 			}
 			$orig_pdf = $filedir . '/' . $filename . '.' . self::INVOICE_FILE_EXTENSION;				// generateDocument writes <ref>.pdf
+
+			// That rebuild fired the afterPDFCreation hook, which generates the e-invoice on its own and
+			// deletes this very temporary XML once it has embedded it. Write it again rather than merge
+			// against a path that points at nothing: the mergers take the XML as content and fall back to
+			// treating the string as content when the file is gone, so what would be embedded is the path
+			// itself - 61 bytes of file name announced as a Factur-X (issue #658).
+			if (!file_exists($xmlfile)) {
+				dol_syslog(get_class($this) . "::generateInvoice temporary XML consumed by the PDF rebuild, generating it again");
+				$xmlfile = $this->generateXML($invoice, $outputlangs);
+			}
 		}
 
 		// Make a copy of the original PDF file
@@ -770,7 +780,10 @@ class FacturXProtocol extends CIIProtocol
 
 		// TODO A third method can be tried using the atgp/factur-x library.
 
-		if (!file_exists($orig_pdf)) {
+		// The mergers below take the XML as content, and treat the string as content when it is not the
+		// path of an existing file. A missing XML therefore does not fail, it gets embedded: check it here
+		// rather than hand over a PDF carrying its own file name (issue #658).
+		if (!file_exists($orig_pdf) || empty($xmlfile) || !file_exists($xmlfile)) {
 			throw new \Exception("XML and/or PDF does not exist");
 		}
 


daGrumpf-bxp:
Closing in favour of #660, which fixes the same issue at its cause.

This PR treated the symptom: the temporary XML disappears mid-call, so write it again and check it before merging. #660 stops it from disappearing — the file was deleted by `afterPDFCreation` re-entering the very generation that was rebuilding the PDF, and muting the hook for that rebuild also removes the second defect this PR left in place: every e-invoice of an invoice without a PDF was produced twice.

The one piece worth keeping travelled to #660: the guard before the merge now tests the XML as well as the PDF, so a missing XML can never be embedded as content again, whatever removes it in the future. The XML regeneration is not there — with the recursion gone it would be defensive code for a case that can no longer happen.

#660 stands alone on `main` and has the full campaign: the four cores repaired, 8082 generations per run compared byte for byte against `main`, phpunit on 9 instances, and both directions end to end through the sandbox in CII and in Factur-X.

PR #669, opened 2026-08-29
FIX [einvoicing] The last flow is the last one of the entity the deletion happens in

daGrumpf-bxp:
Deleting an e-invoicing document linked to a supplier invoice is accepted since 998e8eb5 when the
document is the last record of the table, since the next synchronization brings it back. The record
it is compared against is read without any entity restriction:

```php
$sqlgetlast = "SELECT rowid FROM ".MAIN_DB_PREFIX."einvoicing_document";
$sqlgetlast .= " ORDER BY rowid DESC LIMIT 1";
```

`llx_einvoicing_document` carries an `entity` column and `Document` declares
`$ismultientitymanaged = 1`, so on a multicompany setup the highest rowid of the whole table usually
belongs to another entity. The flow being deleted is then never the last one and the deletion is
refused although it is the last of its own entity — and, the other way round, the guard is decided on
a record the owner of the flow cannot even see.

The row is also not tested before its `rowid` is read: an empty table answers no row at all, and
reading `rowid` on `false` is a warning on PHP 8.

### Reproduced

On the sandbox bench, on `main` at 511901e3, Dolibarr 23.0.3, on rows duplicated from a real
`invoice_supplier` flow so `isEInvoice()` sees what it sees in production:

```
1. copy id=2459 (entity 1), highest rowid of the table
   -> DELETED
2. copy id=2460 (entity 1), with id=2461 (entity 2) above it
   -> REFUSED (The flow 2460 can't be deleted because it is linked to the Dolibarr supplier invoice n° 3336)
```

The second flow is the last one of its entity, and its deletion is refused on the strength of a row
of another entity.

### What changes

The query restricts itself the way `Document::fetchAll()` does, with `getEntity()` on the same
element, and the row is tested before its `rowid` is read.

### Tested

Sandbox bench, on this branch, same scenarios on Dolibarr 23.0.3: both **DELETED**, the flow of
entity 2 left untouched. Whole PHPUnit suite of the module, file by file, on Dolibarr 18 → 24:
16/16 OK on each.


daGrumpf-bxp:

Issue #672, opened 2026-08-29
[E-invoicing] The item net price (BT-146) of a situation line ignores the progress


daGrumpf-bxp:
### Bug

On a situation invoice, the item net price (BT-146) of a line ignores the progress of that line,
while the line net amount (BT-131) applies it. The two no longer agree, and a receiver that rebuilds
the amounts from the price and the quantity — as this very module does when it imports — reads the
full amount instead of the invoiced one.

### Module

einvoicing

### Dolibarr Version

18 to 24 (checked on all of them)

### Steps to reproduce the behavior and expected behavior

First situation invoice, one service line of 2000.00 HT at 10 % of progress. The generated CII says:

```xml
<ram:NetPriceProductTradePrice><ram:ChargeAmount>2000.00000</ram:ChargeAmount></ram:NetPriceProductTradePrice>
<ram:BilledQuantity unitCode="C62">1.0000</ram:BilledQuantity>
<ram:SpecifiedTradeSettlementLineMonetarySummation><ram:LineTotalAmount>200.00</ram:LineTotalAmount>
```

BT-146 x BT-129 = 2000.00, BT-131 = 200.00.

Round trip, really run between two instances through the sandbox Access Point (invoice
`TRI-FA-2608-0256`, flow `i_375122`, acknowledgement `Ok`): the document leaves at 200.00 HT /
240.00 TTC, and the supplier invoice created on the receiving instance is **1166.67 HT / 1400.00
TTC**, its lines rebuilt at 2000.00 and -833.33. The import recomputes the amounts with
`calcul_price_total()` from `subprice` (fed with BT-146) and the quantity, and never reads BT-131.

An ordinary invoice does not have the problem, including with a discount on the line: there the
module already emits the line allowance (BG-27, `ReasonCode` 95) and the arithmetic holds
(2 x 1000.00 - 200.00 = 1800.00).

### What the standards say

Three validators run on the very document above (Saxon, sources below):

| | verdict |
|---|---|
| EN 16931 (CEN schematron, CII) | **0 failure** — no rule ties BT-131 to BT-146 x BT-129 |
| CTC-FR (FNFE `France_RFE`, CII) | **0 failure** — on these two BT it only carries format rules, `BR-FR-DEC-01/BT-131` and `BR-FR-DEC-03/BT-146` |
| Peppol BIS 3.0 (CII) | **`PEPPOL-EN16931-R120` fatal, once per line** |

[...]

PR #677, opened 2026-08-30
FIX #674 [einvoicing] A situation invoice states the instalment it asks for, not the cumulative amount

### LINKED ISSUE #674: [einvoicing] From the second situation invoice on, the e-invoice states the cumulative amount instead of the instalment
### Bug

From the **second** situation invoice of a cycle on, the e-invoice states an amount that is larger
than the one the invoice itself claims. The customer is asked, by the document that has legal value,
for the cumulative amount of the works instead of the instalment.

**Why the two differ, in the core:**

* `llx_facturedet.total_ht` of a situation line holds the **cumulative** amount
  (`qty × unit price × progress`), computed by `calcul_price_total(..., $situation_percent)`.
* `CommonObject::update_price()` then **subtracts the totals of the previous situations** from the
  header (block "Situations totals", `$this->total_ht -= $sit->total_ht` over `get_prev_sits()`),
  unless `INVOICE_USE_SITUATION == 2`. So `llx_facture.total_ttc` is what this instalment asks for -
  it is the receivable, what the payment screen expects and what accounting records.

**What the module does:** `einvoicing/lib/buildinvoicelines.inc.php` sums the **line** amounts (the
cumulative ones) into `taxBasisTotalAmount`, `grandTotalAmount` and `duePayableAmount`. The deduction
the core applies at header level is never carried over.

Measured on a 1000 EUR cycle at 30 %, then 70 %, then 100 %:

| situation | Dolibarr (`llx_facture`) | e-invoice (BT-109/112/115) | gap |
|---|---|---|---|
| 1 (30 %) | 300.00 / 360.00 | 300.00 / 360.00 | - |
| 2 (70 %) | **400.00 / 480.00** | **700.00 / 840.00** | +300.00 / +360.00 |
| 3 (100 %) | **300.00 / 360.00** | **1000.00 / 1200.00** | +700.00 / +840.00 |

* Reproduced **identically on Dolibarr 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and
  24.0.0-beta**, on `main` at 4b70999f.
* **Both protocols** are affected, CII and Factur-X: the amounts come from the shared
  `buildinvoicelines.inc.php`, and the XML extracted back out of the Factur-X container carries the
  same 700.00 / 840.00.
* **The document is valid.** The public SuperPDP validator says `VALID` on the situation 2 document
  (161 checks): it is internally consistent, so no conformance rule can catch this. Negative control
  on the same run: BT-117 moved by six cents on another document gives `INVALID [BR-CO-14]`, so the
  validator does bite.
* **Confirmed end to end through the sandbox.** Situation 2 issued by one instance (400.00 / 480.00
  in its own database) is received by the buying instance as a **supplier invoice of 700.00 HT /
  840.00 TTC**. The buyer would pay 840 for an invoice of 480.

### Not the same in every situation mode

With `INVOICE_USE_SITUATION = 2` (progress stated per invoice rather than cumulatively) the core does
**not** deduct, the line already holds the instalment, and the module is **right** - measured, 400.00
against 400.00. So a fix has to stay conditional on the mode, or it breaks mode 2.

### The open question a fix has to answer

The header totals cannot simply be lowered on their own: the VAT breakdown (BG-23) is built from the
same line amounts, and BR-CO-* require the two to agree. The representation has to be chosen:

1. **A header allowance** (BG-20) per VAT rate deducting the previously invoiced amount - the lines
   keep stating the work done at their cumulative progress, which is what the printed situation
   invoice shows, and BT-109 - BT-107 = BT-112 = the instalment.
2. **Lines at the delta**: each line states only what this instalment invoices. Simplest arithmetic,
   but the line no longer matches the paper invoice, and it moves what #672 has just settled about
   BT-146.
3. **A negative line per previous situation**, the way a deposit deduction is already carried
   (method 1 of XP Z12-014, `(DEPOSIT)` line + BT-25 reference).

I have a working branch for option 1 and will open it as a pull request, validated and tested end to
end, but the choice is a normative one - happy to switch to another representation.

### Module

einvoicing

### Dolibarr Version

18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3, 24.0.0-beta (all reproduce)

1. `INVOICE_USE_SITUATION = 1` (the default cumulative mode).
2. Create a situation invoice with one line, 1 x 1000.00 HT at 20 %, progress 30 %. Validate, generate.
3. From it, create the next situation of the cycle (`situation_counter + 1`, `createFromCurrent()`),
   set the line progress to 70 %, validate, generate.
4. Compare `llx_facture.total_ht` / `total_ttc` of the second invoice (400.00 / 480.00) with
   `ram:TaxBasisTotalAmount` / `ram:GrandTotalAmount` / `ram:DuePayableAmount` of the produced
   document (700.00 / 840.00).

Expected: the document states the amount the invoice claims, 400.00 / 480.00.

daGrumpf-bxp:
Fixes #674.

## Problem

From the **second** situation invoice of a cycle on, the document claimed more than the invoice it
stands for. A cycle of 1000.00 EUR at 30 %, then 70 %, then 100 %:

| situation | Dolibarr (`llx_facture`) | document, before | document, after |
|---|---|---|---|
| 1 (30 %) | 300.00 / 360.00 | 300.00 / 360.00 | 300.00 / 360.00 |
| 2 (70 %) | **400.00 / 480.00** | 700.00 / 840.00 | **400.00 / 480.00** |
| 3 (100 %) | **300.00 / 360.00** | 1000.00 / 1200.00 | **300.00 / 360.00** |

Dolibarr holds the **cumulative** amount of the work done on the line, and
`CommonObject::update_price()` deducts the situations already invoiced from the header of the invoice
(block "Situations totals"). `llx_facture.total_ttc` is therefore the instalment - what the customer
owes, what the payment screen expects, what accounting records. `buildinvoicelines.inc.php` summed
the line amounts into BT-109, BT-112 and BT-115 and never carried that deduction over.

The document stayed **valid** while saying it: it is consistent with itself, so no conformance rule
catches it. And it reached the other side - the buying instance recorded a supplier invoice of
700.00 HT / 840.00 TTC for an invoice of 400.00 / 480.00.

## What this does

What the previous situations already asked for is carried as a **document level allowance**:

```xml
  <ram:ActualAmount>300.00</ram:ActualAmount>
  <ram:ReasonCode>95</ram:ReasonCode>
  <ram:Reason>Deduction of the situations already invoiced</ram:Reason>
  <ram:CategoryTradeTax>
    <ram:TypeCode>VAT</ram:TypeCode><ram:CategoryCode>S</ram:CategoryCode>
    <ram:RateApplicablePercent>20.00</ram:RateApplicablePercent>
  </ram:CategoryTradeTax>
```

so BT-106 = 700.00 keeps saying what has been done - which is what the printed situation invoice
shows and what #672 has just settled about BT-146 - BT-107 = 300.00 says what was already invoiced,
and BT-109 = 400.00, the VAT breakdown (400.00 / 80.00) and BT-115 = 480.00 land on the instalment.
It is the machinery the module already uses for a global discount; only the entry is new.

The amount deducted for a line is the one **its predecessor recorded**, read from the invoice that
carries it (`fk_prev_id`) rather than recomputed here, so the document deducts to the cent what was
invoiced before. A line that appears in this situation and has no predecessor deducts nothing. And
because the result is built line by line where the core builds it invoice by invoice, a gap between
the two is logged rather than transmitted silently.

`INVOICE_USE_SITUATION = 2` states the progress of each invoice separately, the core deducts nothing
and the lines already hold the instalment: **that mode was right** and is left untouched, which the
tests below check rather than assume.

One `en_US` key added, no other language touched.

## Tested

Live on the bench, on the branch actually deployed there, `main` deployed first to measure the
"before" of every table above.

* **Seven cores, one per major version**: 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.3 and
  24.0.0-beta. The same cycle on each: 300.00 / 360.00, then **400.00 / 480.00**, then
  **300.00 / 360.00**, matching `llx_facture` on all seven. On `main` the same cycle gives
  700.00 / 840.00 and 1000.00 / 1200.00 on all seven.
* **Both protocols**: the XML extracted back out of the Factur-X container carries the same
  400.00 / 480.00, so the PDF path is fixed with the plain CII one.
* **Validity**: the three documents of the cycle pass the public SuperPDP validator, `VALID`, and the
  ones carrying the allowance run **175 checks instead of 157** - the allowance rules ran and passed.
  Negative control on the same harness: BT-117 moved by six cents on another document gives
  `INVALID [BR-CO-14]`.
* **End to end through the sandbox**: situation 1 issued and received at 300.00 / 360.00, then
  situation 2 issued at 400.00 / 480.00 and **received as 400.00 / 480.00** - the work line at 700.00
  and the deduction at -300.00 - where the same cycle before this branch landed as 700.00 / 840.00.
  200 and 201 back on both.
* **`INVOICE_USE_SITUATION = 2` unchanged**: 400.00 / 480.00 before and after.
* **Non regression on the other shapes**, same instance, same values before and after: standard,
  deposit, credit note, replacement, multi-VAT with an exempt line, 100 lines, a line discount with a
  negative line, a zero total invoice, a quantity with four decimals.
* Module unit tests file by file on the eight cores, 17 to 24: 16 OK per instance, unchanged, the
  only failure being the pre-existing one on 17 where the module is disabled.
* `php -l` and `phpcs` (Dolibarr ruleset) clean.

## The choice this makes, and what else was on the table

The deduction could also be stated by putting the **delta** on the lines, or by adding one negative
line per previous situation the way a deposit deduction is carried. The allowance was chosen because
it leaves the lines saying what the invoice prints, it needs no new representation in the reader, and
the reading side already turns a header allowance back into a line - the received invoice shows
700.00 and -300.00. Happy to switch if the normative reading points elsewhere; #674 lays out the
three options.


eldy:
I think you are speaking of the case INVOICE_USE_SITUATION = 1
In this mode, data are bugged and corrupted. Sometimes it is a aggregated, sometimes not. You patch seems to manage the old deprecated option  INVOICE_USE_SITUATION = 1 that was never released.
If yest, can you replace test with if getDolGlobalSting('INVOICE_USE_SITUATION') == 1 because only this mode was bugged. Also note that the bug is on standard invoice, i think for credit note, data are saved correctly.

daGrumpf-bxp:
Yes, this is the `INVOICE_USE_SITUATION = 1` case — but I would keep `!= 2` rather than `== 1`, for
one reason: the guard you are naming does not exist on every supported core.

I read the "Situations totals" block of `CommonObject::update_price()` on all of them:

| core | `INVOICE_USE_SITUATION != 2` guard |
|---|---|
| 18, 19 | **absent** — the deduction is unconditional |
| 20 → 24 | present |

So on Dolibarr 18 and 19 — 18 being the module's minimum — the core deducts the previous situations
*even in mode 2*. With `== 1`, the module would stop deducting there while the core still does, and
the e-invoice would announce the cumulative amount again: exactly the bug of #674, just moved to
another pair of versions.

Where the guard does exist, `!= 2` and `== 1` agree on every invoice that can be created, since a
situation invoice requires the option to be > 0 (`invoice.lib.php`, `if (getDolGlobalInt('INVOICE_USE_SITUATION') > 0)`).
They differ only if the option is switched off afterwards — and there the core still deducts (`!= 2`),
so `!= 2` is the one that follows it.

`!= 2` is written to mirror the core condition exactly, which is why the deduction is read from
`fk_prev_id` rather than recomputed: whatever the core subtracted at header level is what the document
states.

On the credit notes: agreed, and the PR already leaves them alone. The condition is
`$object->type == $object::TYPE_SITUATION`, and a situation credit note is `TYPE_CREDIT_NOTE` — the
same type the core excludes one line above its own guard. Nothing to change there.

eldy:
Note that INVOICE_USE_SITUATION = 1 is not supported (was never supported) by dolibarr so has no need to be supported by external modules too. 
If a user used the experimental option INVOICE_USE_SITUATION = 1, he MUST migrate to mode 2 urgently !
Any other feature of dolibarr are not working with this unstable mode enabled (accountancy, reports, ...)

i will fix the test !=2 into == 1 in core to have this cleaner (corrupted data are only in mode 1, mode 3 does not exists yet but may be one day).

daGrumpf-bxp:
Applied, and checked on the bench: `main` carries your `== 1` (4bafb3db) and `develop` carries the
matching core change (*Fix test for old INVOICE_USE_SITUATION = 1*, 2026-09-01). The module now
follows the core condition exactly. I am not asking to reopen the module side — the configurations it
leaves out are configurations that should not exist, and I will treat them as such.

**One request, on the core side: backport that change to 21.0, 22.0, 23.0 and 24.0.**

It is not cosmetic on the released branches. Measured on 23.0.3, a cycle built in mode 2 (1000.00 HT
at 20 %, situation 1 at 30 %, situation 2 at 100 %):

* after validation, `llx_facture` holds **700.00 / 840.00** — correct, the core does not deduct in
  mode 2 and the lines already carry the instalment;
* the setting is then turned off. The yes/no toggle does not write 0, it *deletes* the constant
  (`ajax_constantonoff` calls `dolibarr_del_const`), so `getDolGlobalInt()` reads 0;
* any later `update_price()` on that invoice: **700.00 / 840.00 → 400.00 / 480.00**. With the guard
  written `!= 2`, the value 0 passes it and the core deducts the previous situation from an invoice
  whose lines already hold the instalment. The stored totals of a **mode 2** invoice are corrupted.

`== 1` closes that. It is closed on `develop`; it is open on 21.0 → 24.0.

### Why our reports keep arriving in mode 1

Not to argue the support policy — only so the reports make sense. Mode 1 is what the setup screen
writes: `INVOICE_USE_SITUATION` is a yes/no toggle in `admin/facture_situation.php`, then
`admin/invoice_situation.php`, and `setAsYesNo()` can only write 0 or 1. That is exactly what core PR
#39098 (ATM-NicolasV, 2026-07-10) fixed, and the reason is in the code it added:

> `INVOICE_USE_SITUATION` is a 3-state flag (0=off, 1=cumulative/legacy, 2=progressive) […] A yes/no
> toggle can only write 0 or 1 and deletes the const when turned off (ajax_constantonoff calls
> dolibarr_del_const): mode 2 could not be selected and was silently dropped.

That fix is on branches 21.0 → 24.0, but of the releases the module is tested on — 18.0.10, 19.0.4,
20.0.4, 21.0.4, 22.0.5, 23.0.3, 24.0.0 — only **24.0.0** carries the 3-value select. So a user who
enabled situation invoices from the interface before that is in mode 1 without having chosen it, and
could not have chosen otherwise. #666 was reported that way (Dolibarr 24, "progression cumulative").
Mode 2 also has no implementation at all before v20: no `== 2` anywhere in 18 and 19.

Which is why the backport also gives those users a clean answer, and it is the one we will give:
upgrade to the version carrying it, then select the progressive mode in the setup screen.

### Not covered, and agreed as out of scope for the module

Measured on 18.0.10 and 23.0.3, same cycle, for the record rather than as a request:

| configuration | `llx_facture` | document |
|---|---|---|
| option = 1 | 700.00 / 840.00 | 700.00 / 840.00 (allowance 300.00) |
| option turned off while the cycle exists | 700.00 / 840.00 | **1000.00 / 1200.00** |
| constant forced to 2 on 18/19, where mode 2 does not exist | 700.00 / 840.00 | **1000.00 / 1200.00** |

Both of those are states that should not be reachable in a supported setup — the option is not meant
to be switched off mid-cycle, and mode 2 cannot be selected on a core that does not implement it. No
change requested in the module for them.

PR #688, opened 2026-08-31
FIX #681 [einvoicing] Generating a Factur-X no longer breaks the next PDF of the request

### LINKED ISSUE #681: Fatal error on einvoice generation
### Bug

1. Errors were generated after validating an invoice, page error 500.
2. After the following warnings and errors, the flash message indicated that electronic invoice generation was successful

Anonymized log:
```text
[Mon Aug 31 13:11:20.567311 2026] [php:warn] [pid 683421:tid 683421] [client 192.168.0.254:17114] PHP Warning:  Undefined array key "errorCode" in /home/www/example.com/dolibarr-community-modules/einvoicing/class/providers/SuperPDPProvider.class.php on line 1199, referer: https://example.com/dolibarr/compta/facture/card.php?facid=453&action=valid&token=3398af698c57ad0f621868ca3c864aed
[Mon Aug 31 13:11:20.567340 2026] [php:warn] [pid 683421:tid 683421] [client 192.168.0.254:17114] PHP Warning:  Undefined array key "errorMessage" in /home/www/example.com/dolibarr-community-modules/einvoicing/class/providers/SuperPDPProvider.class.php on line 1200, referer: https://example.com/dolibarr/compta/facture/card.php?facid=453&action=valid&token=3398af698c57ad0f621868ca3c864aed
[Mon Aug 31 13:11:20.842892 2026] [php:error] [pid 683421:tid 683421] [client 192.168.0.254:17114] PHP Fatal error:  Declaration of setasign\\Fpdi\\FpdfTplTrait::setPageFormat($size, $orientation) must be compatible with TCPDF::setPageFormat($format, $orientation = 'P') in /home/www/example.com/dolibarr-community-modules/einvoicing/vendor/setasign/fpdi/src/FpdfTplTrait.php on line 48, referer: https://example.com/dolibarr/compta/facture/card.php?facid=453&action=valid&token=3398af698c57ad0f621868ca3c864aed
```

### Module

einvoicing

### Dolibarr Version

20.0.5

### Environment PHP

8.2.31

### Environment Database

_No response_

Freshly activated einvoicing_1.0.3.
Selected CII, Super PDP, Automatic invoice generation.

Validated invoice.

### Attached files

_No response_

daGrumpf-bxp:
Found while checking #681 against the current `main`. The fatal reported there is fixed, but the FPDF collision it comes from has a second direction, still open, which answers a blank 500.

### The problem

Below Dolibarr 24 the core declares its own shim, `class FPDF extends TCPDF {}` in `htdocs/includes/tcpdi/tcpdi.php`, and the module carries the real `setasign/fpdf` in its vendor folder. Only one of the two can exist in a PHP request, and until now which one it was depended on the order of that request rather than on the version of Dolibarr.

`FacturXProtocol` chooses between the TCPDF merger and the horstoeko/setasign one by asking whether the shim is already declared. It is declared only once the core has loaded its PDF stack - which a generation that finds the source PDF already on disk never does, because it does not render one. The module then takes the setasign branch and autoloads its own FPDF, and the next PDF the core renders in that same request dies on:

```
PHP Fatal error: Cannot redeclare class FPDF
(previously declared in einvoicing/vendor/setasign/fpdf/fpdf.php:10)
in htdocs/includes/tcpdi/tcpdi.php on line 22
```

Reached by the mass generation of the invoice list: the first invoice still has its PDF, so the setasign branch runs and loads the real FPDF; a later invoice of the same batch has lost its PDF, `generateInvoice()` rebuilds it, and the request dies. The page answers a 500 and none of the remaining invoices of the batch gets its e-invoice.

### The fix

The PDF stack of the core is loaded before the choice is made, so the branch is settled by the Dolibarr version alone: below 24 the container is always built with TCPDF, from 24 on - where the core no longer declares the shim - always with the horstoeko writer. The discarded instance of `pdf_getInstance()` is also what defines the `K_*` constants TCPDF reads, which no PDF render had defined yet in that request.

### Tested

On the bench, Dolibarr 18.0.10, 20.0.4, 22.0.5 and 24.0.0, PHP 8.4:

- the two-invoice mass generation above: 500 before, both e-invoices generated after;
- validation with automatic generation, CII and Factur-X, on the four versions: unchanged;
- standalone generation of the e-invoice of an invoice that keeps its PDF: below 24 it now goes through the TCPDF merger, on 24 it still goes through the horstoeko writer;
- the container produced below 24 is unchanged - PDF/A-3 output intent, `factur-x.xml` as an embedded file, `/AFRelationship`, Factur-X XMP.

### DIFF
--- a/einvoicing/class/protocols/FacturXProtocol.class.php
+++ b/einvoicing/class/protocols/FacturXProtocol.class.php
@@ -267,14 +267,29 @@ class FacturXProtocol extends CIIProtocol
 			$creator = (string) pdfExtractMetadata($orig_pdf, 'Creator');
 		}
 
+		// The choice below reads the global class FPDF, and must not depend on whether the request
+		// happened to render a PDF before reaching here. Below Dolibarr 24 the core declares its own
+		// "class FPDF extends TCPDF {}" in htdocs/includes/tcpdi/tcpdi.php, but only once its PDF stack
+		// is loaded: a generation that finds the source PDF already on disk - the mass generation of the
+		// invoice list, typically - never renders one, so the shim is absent and the branch below picks
+		// the horstoeko/setasign writer, which autoloads the real FPDF of the module. Both classes are
+		// named FPDF and only the first one of the request survives, so the next core PDF of that same
+		// request dies on "Cannot redeclare class FPDF ... in includes/tcpdi/tcpdi.php". Load the PDF
+		// stack of the core now, so the question is settled by the Dolibarr version alone. The instance
+		// is discarded: pdf_getInstance() is called for what it loads and for the K_* constants TCPDF
+		// needs, which no PDF render defined yet in this request.
+		if (!class_exists('FPDF', false)) {
+			require_once DOL_DOCUMENT_ROOT . '/core/lib/pdf.lib.php';
+			pdf_getInstance();
+		}
+
 		try {
 			if (class_exists('FPDF', false) && is_subclass_of('FPDF', 'TCPDF')) {
-				// Below Dolibarr 24, htdocs/includes/tcpdi/tcpdi.php declares "class FPDF extends TCPDF {}"
-				// as soon as any PDF is rendered in the request - the invoice PDF produced at validation,
-				// right before this hook runs. The horstoeko/zugferd writer then inherits from TCPDF instead
-				// of the real FPDF and dies on ZugferdPdfWriter::_getpagesize(). Merge with TCPDF itself,
-				// which needs no FPDF at all and supports PDF/A-3 natively. Reaching this branch means
-				// the core has loaded TCPDF, which is what the merger descends from.
+				// Below Dolibarr 24, htdocs/includes/tcpdi/tcpdi.php declares "class FPDF extends TCPDF {}".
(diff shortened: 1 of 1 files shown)


eldy:
Note for other dev and users:  Using factur-x format is seriously discouraged. It will generates a lot of other bug like this one. For this reason it is highly recommanded to use CII.
A stable mode with Factur-x format will be very difficult (to not say impossible).

daGrumpf-bxp:
For what it is worth, from the other side: Factur-X has been solid here for a while. We generate it on
every core from 18 to 24 on a bench, the container is checked as PDF/A-3 by veraPDF on every PR (the
"Factur-X container is PDF/A-3" job), and full round trips through the Access Point — issue, receive,
lifecycle statuses back — run in Factur-X as well as in CII.

The bugs it does produce are of this kind: not the XML, but the PDF container colliding with whatever
else the request already loaded. That is a bounded problem, and each one has stayed fixed.

None of which argues against recommending CII by default — it has strictly fewer moving parts, and an
integrator with the choice should take it. Just that Factur-X does not look impossible from where we
run it, and plenty of French customers will be handed it rather than choosing it.

PR #692, opened 2026-08-31
Add hooks for CII XML preview generation and cleanup

### LINKED ISSUE #687: Add preview of XML file (CII)
### Feature Request

As we're just starting with the e-invoicing, I prefer to check the generated data to make sure everything is in order.

Currently I need to download and then open in a specific application.

A preview link showing the raw xml would already be helpful.

### Module

einvoicing

### Use case

_No response_

### Suggested implementation

_No response_

mdeweerd:
Fixes #687: Adds three hooks to enable preview of CII XML files:
- showDocuments: Generates .txt preview files for _cii.xml files
- downloadDocument: Cleans up stale preview files on file access
- topmenu: Global throttled cleanup on every page load

The .txt extension ensures dolIsAllowedForPreview() returns true, enabling preview links in the document list UI.

### DIFF
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -4,6 +4,7 @@
  * Copyright (C) 2026		Charlene Benke			<charlene@patas-monkey.com>
  * Copyright (C) 2026       Frédéric France         <frederic.france@free.fr>
  * Copyright (C) 2026		Jose Martinez				<jose.martinez@pichinov.com>
+ * Copyright (C) 2026		MDW							<mdeweerd@users.noreply.github.com>
  *
  * This program is free software; you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


einvoicing/core/modules/modEInvoicing.class.php
```diff
@@ -687,4 +687,103 @@ public function remove($options = '')
...
+	 * generated on-demand and will be automatically cleaned up by downloadDocument
+	 * or topmenu hooks after 1 hour to avoid disk clutter.
+	 *
+	 * @param  array          $parameters	Hook parameters (includes filedir, modulepart)
+	 * @param  CommonObject   $object		Current object
+	 * @param  string         $action		Current action
+	 * @param  HookManager    $hookmanager	Hook manager
+	 * @return int							0
+	 */
+	public function showDocuments($parameters, &$object, &$action, $hookmanager)
```
eldy: If this is hook functions, it should be stored into the file actions_einvoicing.class.php not modEInvoicing.class.php

einvoicing/core/modules/modEInvoicing.class.php
```diff
@@ -687,4 +687,103 @@ public function remove($options = '')
...
+		$modulepart = &$parameters['modulepart'];
+
+		if (!in_array($modulepart, array('facture', 'propal', 'commande', 'societe'))) {
+			return 0;
+		}
+
+		foreach (glob($filedir.'/*_cii.xml') as $xml) {
+			$txt = $xml.'.txt';
+			if (!file_exists($txt) || filemtime($txt) < filemtime($xml)) {
+				file_put_contents($txt, file_get_contents($xml));
```
eldy: i don't understand.
The hook showDocuments is designed to run code when we show documents.So content should be a read only action.
Here it seems you create a file ?

einvoicing/core/modules/modEInvoicing.class.php
```diff
@@ -687,4 +687,103 @@ public function remove($options = '')
...
+				touch($txt, time());
```
eldy: time and touch are not allowed function in dolibarr.

einvoicing/core/modules/modEInvoicing.class.php
```diff
@@ -687,4 +687,103 @@ public function remove($options = '')
...
+		global $conf;
+
+		if (!isset($_SESSION['cii_cleanup']) || $_SESSION['cii_cleanup'] < time() - 300) {
+			$_SESSION['cii_cleanup'] = time();
+			$modules = array('facture', 'propal', 'commande', 'societe');
+			foreach ($modules as $module) {
+				$dir = $conf->$module->dir_output;
+				foreach (glob($dir.'/*/*_cii.xml.txt') as $file) {
+					if (filemtime($file) < time() - 3600) {
+						unlink($file);
```
eldy: Removing file during an operation to render the menu ?
This is a suspicious action.

If you want to add a feature to show a preview of data related to a file, you can add a wrapper to download this data and include the content or link returned by this wrapper at the place you want to show it into another hook.

May be you can describe more precisly what you want to do so i can provide a direction.

einvoicing/core/modules/modEInvoicing.class.php
mdeweerd: True, I copied back to the wrong file.

einvoicing/core/modules/modEInvoicing.class.php
mdeweerd: The 'xml' extension is not accepted, so this does create a temporary file for viewing.

Cleanest way I could think of to allow previewing the xml file.

einvoicing/core/modules/modEInvoicing.class.php
mdeweerd: >  you can add a wrapper to download this data

I did not see how, found showDocuments to be a good location to hook into.
What other hook is available ?


> Removing file during an operation to render the menu

What is the proper way to ensure the temporary file is removed some time after the page was rendered?

einvoicing/core/modules/modEInvoicing.class.php
mdeweerd: OK, I could not find the list of forbidden functions.  Here these are local comparisons (no db), but a rule is a rule.

Found usages in dolibarr too:

```
htdocs/core/lib/agenda.lib.php:262:49:           if ($obj->percent == 0 && $datep && $datep < time()) {
htdocs/core/lib/admin.lib.php:1118:42:                       $arrayofSessions[$idsess]["age"] = time() - filectime($fullpath);
htdocs/core/modules/openid_connect/callback.php:64:58:   dolSetCookie('DOL_rollback_url_'.dol_getprefix(''), "", time() + 1);
```

I am adding a Phan rule to report these forbidden usages, please provide a list.

daGrumpf-bxp:
Two consequences of the `.txt` twin that are independent of where the hook code lives, in case this comes back:

- The twin becomes its own row of the document table. `showDocuments` runs at `html.formfile.class.php:482` (`:370` on 18), before the list is built at `:544` with `dol_dir_list($filedir, 'files', 0, '', '(\.meta|_preview.*\.png)$', ...)` — no include filter, and the exclude filter only covers `.meta` and `_preview*.png` — so one document shows as two, the second being a `.txt` offered for download like a real one.
- `$conf->commande->dir_output` is not set on 18/20/23/24 (orders only have `multidir_output`), so for that module the `topmenu` cleanup evaluates `glob('' . '/*/*_cii.xml.txt')`, which walks the subdirectories of `/` on every page load.

I have opened #696 with the other direction from this thread — a read-only viewer page plus `formBuilddocLineOptions`, the hook executed once per file line, so nothing is written and nothing has to be cleaned up. It is an alternative to this PR, not an addition; happy for either to be the one that lands.

eldy:
It is a beautifull hack 🥇 

But I think using the hook formBuilddcLineOptions looks a better direction.
It allows to include a html picto inside a line. This picto can be a reader of xml file, for example by just opening a new tab in source mode. So i would prefer to continue to exchange on #696 for this point.So i close this one.

Why don't you use a simple clic on xml file name ? It can open it for view into another tab natively ?

But the true real solution for long term is to have the native preview of dolibarr working into the dolibarr core for xml files

mdeweerd:
> It is a beautifull hack 🥇

Thanks.

> But I think using the hook formBuilddocLineOptions looks a better direction. It allows to include a html picto inside a line. This picto can be a reader of xml file, for example by just opening a new tab in source mode.

The idea was to delegate the view functionality to document.php to avoid housekeeping of rights, fetching the file, etc.
The goal was to reuse the existing viewer in a backwards compatible way, so mainly considering it as a black box.
`formBuilddocLineOptions` does allow adding html.

It comes down to the effort: this hack looked like (and was) limited effort.

> So i would prefer to continue to exchange on #696 for this point.So i close this one.

Integration in einvoice is great of course.

> Why don't you use a simple clic on xml file name ? It can open it for view into another tab natively ?

We have `Content-Disposition: attachment; filename="FA2605-0375_cii.xml"` in the server header which instructs the browser to proceed with a download.

> But the true real solution for long term is to have the native preview of dolibarr working into the dolibarr core for xml files

Sure, but we can't add new functionnality to old versions, so only for versions from 25 upwards.

PR #693, opened 2026-08-31
NEW: Add DolibarrForbiddenFunctionPlugin to detect forbidden functions

mdeweerd:
# Add DolibarrForbiddenFunctionPlugin to detect forbidden functions

- Added a new plugin to detect forbidden functions in Dolibarr code
- The plugin checks for the use of 'time()' and 'touch()' functions
- The plugin is added to the list of plugins in the config.php file

### DIFF
--- a/dev/tools/phan/config.php
+++ b/dev/tools/phan/config.php
@@ -90,6 +90,7 @@ return [
 	'plugins' => [
 		__DIR__.'/plugins/SqlInjectionPlugin.php',
 		__DIR__.'/plugins/NoVarDumpPlugin.php',
+		__DIR__.'/plugins/DolibarrForbiddenFunctionPlugin.php',
 		// checks if a function, closure or method unconditionally returns.
 		// can also be written as 'vendor/phan/phan/dev/tools/phan/plugins/AlwaysReturnPlugin.php'
 		'DeprecateAliasPlugin',
--- a/dev/tools/phan/plugins/DolibarrForbiddenFunctionPlugin.php
+++ b/dev/tools/phan/plugins/DolibarrForbiddenFunctionPlugin.php
@@ -63,7 +63,10 @@ class DolibarrForbiddenFunctionVisitor extends PluginAwarePostAnalysisVisitor
 		if (!is_string($name)) {
 			return;
 		}
-		if (!in_array($name, ['time', 'touch'])) {
+		if (!in_array($name, [
+			'time', 'touch',
+			'curl_init', 'curl_exec', 'curl_setopt_array', 'curl_close', 'curl_getinfo', 'curl_error',
+		])) {
 			return;
 		}
 		$this->emitPluginIssue(
--- a/dev/tools/phan/plugins/DolibarrForbiddenFunctionPlugin.php
+++ b/dev/tools/phan/plugins/DolibarrForbiddenFunctionPlugin.php
@@ -1,0 +1,81 @@
+<?php
+/* Copyright (C) 2026		MDW							<mdeweerd@users.noreply.github.com>
+ */
+
+declare(strict_types=1);
+
+use ast\Node;
+use Phan\PluginV3;
+use Phan\PluginV3\PluginAwarePostAnalysisVisitor;
+use Phan\PluginV3\PostAnalyzeNodeCapability;
+
+/**
+ * DolibarrForbiddenFunctionPlugin hooks into one event:
+ *
+ * - getPostAnalyzeNodeVisitorClassName
+ *   This method returns a visitor that is called on every AST node from every
+ *   file being analyzed
+ *
+ * A plugin file must
+ *
+ * - Contain a class that inherits from \Phan\PluginV3
+ *
+ * - End by returning an instance of that class.
+ *
+ * It is assumed without being checked that plugins aren't
+ * mangling state within the passed code base or context.
+ *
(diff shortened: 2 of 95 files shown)


mdeweerd:
@eldy This reveals some violations of the forbidden functions rule for time() in stancer.  Do you want to ignore these or fix them while merging this PR to main?

dev/tools/phan/config.php
```diff
@@ -90,6 +90,7 @@ function directory_list()
 	'plugins' => [
 		__DIR__.'/plugins/SqlInjectionPlugin.php',
 		__DIR__.'/plugins/NoVarDumpPlugin.php',
+		__DIR__.'/plugins/DolibarrForbiddenFunctionPlugin.php',
```
eldy, line 93: Is this new check run ONLY on files modified by the PR ?
Because each module will have a different maintener, we need a CI that run only on modified files.
So if someting is wrong in a module, other module won't be affected.

mdeweerd: The plugin is subject to the phan limitations and we see on Dolibarr that any phan issue also affects the individual PR.

If only PR's without Phan issues are merged to main and no phan issues are introduced by direct edits on main, then this is a non issue.

In Dolibarr, the majority of Phan issues on main come from:
- Direct changes in develop or other integration branches;
- Merged changes made in old branches with less checks.

This new phan plugin is about forbidden functions, so nothing tricky or complex.  I have not ported the "curl" limitation yet from Dolibarr.

IMHO the main question is what to do about the following "violations":

```
stancer/admin/test.php:139 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/admin/test.php:179 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/class/actions_stancer.class.php:180 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/class/actions_stancer.class.php:190 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_dispute.lib.php:518 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_dispute.lib.php:610 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_dispute.lib.php:747 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_dispute.lib.php:899 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_payment.lib.php:1110 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_payment.lib.php:1186 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_refresh.lib.php:38 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_refresh.lib.php:103 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_refresh.lib.php:796 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_refresh.lib.php:856 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_refresh.lib.php:1298 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
stancer/lib/stancer_refresh.lib.php:1496 DolibarrForbiddenFunctionPlugin time() is not allowed in Dolibarr code
```

eldy: We are not able to cotrol with PR is merged or not as each PR will be merged by different people managing themself their own module.
So we need a CI that work only on modified files.

A global CI in cron mode could be add but for pull request, CI must process only modified files.

mdeweerd: So you want to disable all CI now because we can not operate on changed files only.

Suppose that CI only operates on changed files.
What will likely happen is that the issue is reported only once upon the merge, and then it only appears in the cron report.
Which will likely be ignored most of the time as it is not a constant reminder of the issue.  Until somebody picks up the slack.

How about enforcing a rule that no PR can be merged if CI is not working, and allow "anybody" to only merge to main, not push to main.
Except for some rare corner cases, anything merged to PR has a clean CI.  For the moment the PRs could even work on the entire codebase as its size is limited and execution time is reasonable.

mdeweerd:
I guess that you are aware about branch rules:

https://github.com/Dolibarr/dolibarr-community-modules/settings/rules/new?target=branch

image

eldy:
> I guess that you are aware about branch rules:
> 
> https://github.com/Dolibarr/dolibarr-community-modules/settings/rules/new?target=branch
> 
> image

Do you mean rule to force blocking merging if CI fails ?
Yes, but in real life this does not work. To be efficient, a maintener need sometimes to merge even if CI is wrong and fix later. So i don't think this rule will be used.

mdeweerd:
Updated as for Dolibarr - just over 600 violations.

eldy:
Due to high priority on dev of some community modules (einvoicing), i will not strenghten CI too fast for the moment to avoid collisions with the high amount of work nowday

PR #694, opened 2026-09-01
FIX #683 [einvoicing] The deliver-to party names the company, and an address keeps its lines

### LINKED ISSUE #683: CII invoicing address inconsistency when using parent entity contact for invoicing
### Bug

Could be related to #463.

I have an invoice to be invoiced to the mother establishement (using the SIREN), delivered at another address (other SIRET).

So I set a contact of the parent entity as the invoicing contact.  The person's name and email appears under "<ram:DefinedTradeContact>" but the physical invoicing address is not set to that contact.  This creates an inconsistency between the name and the physical address for that person.

I have set "FACTURE_PAYMENTS_WITH_PARENT_COMPANY" to 1 but this does not change the invoicing address.

So I keep getting these warnings - maybe they are just warnings and I can ignore them, or they could be a blocker for the process.

Image

Pre-validations succeed:

Image

Finally, I am not sure we need to have a contact name in the invoice.  Generally invoices are made out to the company.  Adding the name of a person in the CII adds personal data we could avoid.  I'll probably define a generic contact if this can not be avoided otherwise.

### Module

einvoicing

### Dolibarr Version

22.0.5

### Environment PHP

_No response_

### Environment Database

_No response_

Parent Company and Child Company, different addresses
Proposal for Child converted in Invoice (may not be a necessary step).

To bill invoice to parent company, add invoice contact.  (Re)generate the CII (Electronic invoice).  Observe that the contact name and email is present as the billing contact, but that the billing address is unchanged.

Ultimate goal: Create invoice to Parent company for "delivery" to child company.

### Attached files

_No response_

daGrumpf-bxp:
Fixes #683

A delivery address entered as a `SHIPPING` contact came out of the CII as the name of the contact
*person*, with an address made of the two or three lines of the field run together inside one
`LineOne`. Reproduced exactly as reported, on a parent/child pair with the invoice issued to the
parent, on Dolibarr 22.0.5 and current `main`:

```xml
<ram:ShipToTradeParty>
  <ram:Name>Jean DELIVRANCE</ram:Name>
  <ram:PostalTradeAddress>
    <ram:PostcodeCode>95520</ram:PostcodeCode>
    <ram:LineOne>5 avenue de la Fille
Batiment C
Zone des Bruyeres</ram:LineOne>
```

While reproducing, a third symptom of the same cause showed up, and it is the more common one: a
shipping contact that carries **no address of its own** — the ordinary case, since the address is
normally on the company — produces a deliver-to party with no address at all.

```xml
<ram:ShipToTradeParty>
  <ram:Name>Jean DELIVRANCE</ram:Name>
  <ram:PostalTradeAddress>
    <ram:CountryID>FR</ram:CountryID>       <!-- and nothing else -->
```

## What was wrong

BG-15 was built from the contact person instead of from the company the contact belongs to, and a
postal address was handed whole to the first of the three terms the norm has for it.

The core answers both questions, and answers them in the shipping frame of the invoice PDF
(`core/lib/pdf.lib.php`, identical from 18 to 24):

- `pdfBuildThirdpartyName()` given a `Contact` returns **the name of its thirdparty**, never the
  person — BT-70 names a party;
- `pdf_build_address()` with `$usecontact = 1` reads the address of the contact when it carries one,
  and otherwise `$targetcontact->thirdparty`.

BG-15 is now built the same way, so the XML and the PDF of one invoice stop naming two different
things. It also answers the reporter's last point: the document no longer carries a personal name it
has no use for.

The address split is not specific to the deliver-to party — a **seller** or a **buyer** whose address
holds two lines had them run together in `LineOne` just the same (`sellerlinetwo`/`sellerlinethree`
and their buyer twins were hardcoded to `""`, and `buildParty()` emitted `LineOne` only). EN 16931
gives an address three terms — BT-35/36/162, BT-50/51/163, BT-75/76/165 — and they are now filled in
the XSD order of `TradeAddressType` (`PostcodeCode, LineOne, LineTwo, LineThree, CityName,
CountryID`). Nothing is lost when an address holds more than three lines: the surplus joins the
third. **A single-line address, which is the ordinary case, produces exactly the document it produced
before.**

## After

```xml
<ram:ShipToTradeParty>
  <ram:Name>Fille 683 SAS</ram:Name>
  <ram:PostalTradeAddress>
    <ram:PostcodeCode>95520</ram:PostcodeCode>
    <ram:LineOne>5 avenue de la Fille</ram:LineOne>
    <ram:LineTwo>Batiment C</ram:LineTwo>
    <ram:LineThree>Zone des Bruyeres</ram:LineThree>
    <ram:CityName>OSNY</ram:CityName>
    <ram:CountryID>FR</ram:CountryID>
```

and the contact with no address of its own now gets the full address of the company it belongs to,
instead of a lone country code.

## Tested

Bench: `scw-exciting-tu`, 8 Dolibarr instances, on the tree actually deployed (this branch at
`07d20da6`), never on a worktree.

- **Reproduced first on the unpatched tree**, all three symptoms, then shown gone.
- **The three shapes**, on Dolibarr **18.0.10, 20.0.4, 22.0.5, 23.0.3 and 24.0.0** — identical output
  on all five: contact with a three-line address, contact with no address, contact with a one-line
  address (byte-identical to before the patch on that last one).
- **PHPUnit, file by file, on 18 → 24**: 16/16 OK on each. `EInvoicingSamplesTest` compares generated
  XML against the fixtures, so the specimen chain is unchanged.
- **Generation matrix on 22.0.5, CII and Factur-X**: standard, deposit, credit note, replacement and
  a three-invoice situation cycle — 14 documents, all generated, all with the buyer address now
  spread over the three lines.
- **Validation**: the 7 CII of that matrix are **VALID** against the FNFE chain
  (XSD `CrossIndustryInvoice_100pD22B` + `FACTUR-X_EN16931.xslt` + `BR-FR-Flux2-Schematron-CII`),
  0 failure. The before/after pair of the reported case validates on both sides, which is the point:
  the old document was conformant and wrong, not invalid.
- `phpcs` (dolibarr ruleset) clean on the four files touched.

### DIFF
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -3700,6 +3700,33 @@ class EInvoicing
 	}
 
 	/**
+	 * Split a Dolibarr postal address into the three lines EN 16931 has for it.
+	 *
+	 * Dolibarr keeps a postal address as one free text field where the user separates the lines with
+	 * newlines, while the norm gives an address three separate terms - BT-35/36/162 for the seller,
+	 * BT-50/51/163 for the buyer, BT-75/76/165 for the deliver-to party. Handing the whole field to
+	 * the first of them puts raw newlines inside a single element, which the receiving side renders
+	 * as one run-on line (issue #683).
+	 *
+	 * Nothing is dropped: the norm stops at three lines, so a fourth and beyond join the third,
+	 * separated by a comma, rather than disappearing from the document.
+	 *
+	 * @param	string	$address	Address as Dolibarr stores it, lines separated by newlines
+	 * @return	string[]			Exactly three lines, empty strings when the address has fewer
+	 */
+	public function splitAddressLines($address)
+	{
+		$lines = preg_split('/\R/', (string) $address);
+		$lines = array_values(array_filter(array_map('trim', $lines), 'strlen'));
+
+		if (count($lines) > 3) {
+			$lines = array_merge(array_slice($lines, 0, 2), array(implode(', ', array_slice($lines, 2))));
+		}
+
+		return $lines + array('', '', '');
+	}
+
+	/**
 	 * Generates the deposit, standard and credit note CII sample-invoice chain and returns the normalized XML of each.
 	 *
 	 * Generates three sample invoices (deposit, standard, and credit note) using fixed specimen
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -3723,7 +3723,7 @@ class EInvoicing
 			$lines = array_merge(array_slice($lines, 0, 2), array(implode(', ', array_slice($lines, 2))));
 		}
 
-		return $lines + array('', '', '');
+		return array_pad($lines, 3, '');
 	}
 
 	/**
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -2675,8 +2675,17 @@ class CIIProtocol extends AbstractProtocol
 		// TradeAddressType is reduced to the country code by the MINIMUM schema
 		if (!$this->isMinimumProfile($profile)) {
 			$addr->appendChild($doc->createElement('ram:PostcodeCode', $data[$prefix . 'postcode']));
+			// The three address lines the norm has: BT-35/36/162 for the seller, BT-50/51/163 for the
+			// buyer. XSD order inside TradeAddressType is PostcodeCode, LineOne, LineTwo, LineThree,
+			// CityName, CountryID - the elements are written in that order and nowhere else.
 			if (!empty($data[$prefix . 'lineone'])) {
 				$addr->appendChild($doc->createElement('ram:LineOne', htmlspecialchars($data[$prefix . 'lineone'])));
+			}
+			if (!empty($data[$prefix . 'linetwo'])) {
+				$addr->appendChild($doc->createElement('ram:LineTwo', htmlspecialchars($data[$prefix . 'linetwo'])));
+			}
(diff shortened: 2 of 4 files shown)


daGrumpf-bxp:
Non-regression check against the shape BG-15 was originally built for (#267), because naming the
company of the contact could take something away rather than add it. It did, and the last commit
fixes it.

Two cases, each with a shipping contact carrying **its own** address, different from its company's:

| | contact belongs to | `ram:Name` before this PR | first version of this PR | now |
|---|---|---|---|---|
| **A** | the invoiced company itself, contact named `Livraison Site Nord` | `Livraison Site Nord` | `Acheteur REG683 SA` ❌ | `Livraison Site Nord` |
| **B** | another company (issue #683) | `DUPONT` (the person) | `Filiale REG683 SAS` | `Filiale REG683 SAS` |

In case A the company name is already the `BuyerTradeParty` name, so BT-70 was saying nothing, while
the label the user typed was the only thing naming the delivery point.

The core draws the line in the same place, and that is now the condition: `pdf_build_address()`
switches to `$targetcontact->thirdparty` only when `$targetcontact->socid != $targetcompany->id`.

**The address rule was never affected and is unchanged**: in both cases the contact's own address
still wins over its company's — `Adresse CONTACT ligne 1/2`, `99999 VILLE-CONTACT` — it is only
split over `LineOne`/`LineTwo` now instead of being run together. The company address is used only
when the contact carries none, which before this PR produced a deliver-to party with no address at
all.

Verified on Dolibarr 18, 20, 22, 23 and 24: same four outcomes on all five (cases A and B above,
plus the reported case and the contact with no address). PHPUnit file by file on 18 → 24, 16/16 on
each.

PR #696, opened 2026-09-01
NEW: #687 [einvoicing] The e-invoice XML can be read without downloading it

### LINKED ISSUE #687: Add preview of XML file (CII)

daGrumpf-bxp:
Fixes #687.

The core does not preview XML: `dolIsAllowedForPreview()` (`core/lib/functions.lib.php`) whitelists a fixed list of mime subtypes — `bmp, jpeg, png, gif, tiff, pdf, plain, css, webp` (+ `webm/mp4` since 23, `avif` since 24, `svg+xml` behind `MAIN_ALLOW_SVG_FILES_AS_IMAGES`). A `.xml` resolves to `text/xml`, which is not in it, so `FormFile::showPreview()` renders no picto, and `document.php` serves anything outside that list as `application/octet-stream`. Neither `dol_mimetype()` nor `dolIsAllowedForPreview()` is hookable, so the list cannot be extended from a module.

That exclusion is not an oversight: an inline `text/xml` response served from the Dolibarr origin can carry an `xml-stylesheet` processing instruction, and the XSLT it points at renders HTML in that origin. So this does not serve the XML inline, does not relax the whitelist, and does not touch `document.php`. It reads the file on the server and answers it escaped inside a `<pre>`.

Only the XML is concerned. Under Factur-X the e-invoice is a PDF, and a PDF Dolibarr previews on its own, so nothing is added on that line.

### What it adds

- `einvoicing/xmlpreview.php?id=<invoice id>` — a read-only page. It checks the right to read the invoice through `restrictedArea()`, resolves the file, and prints it. Nothing is written to disk, so there is nothing to clean up afterwards. Asked with `&mode=raw` it answers the XML alone, without the menus, and with `&element=supplier` it answers for a supplier invoice.
- Two resolvers on `EInvoicing`: `getEInvoiceXmlFilePath()` for the invoice sent to a customer, `getSupplierEInvoiceXmlFilePath()` for the document received from a supplier. Neither is `getEInvoiceFilePath()`, which names the file of the configured protocol and so answers a PDF under Factur-X; each mirrors the naming its writer uses.
- `Actions_EInvoicing::formBuilddocLineOptions()` — the core hook executed once per file line. It adds the usual preview picto on the line of the e-invoice file, with the `documentpreview` class the core binds in `lib_foot.js.php`, so **the XML opens in the same preview dialog as a PDF** rather than in a page of its own. Without javascript the `href` is simply followed and the same page answers as a full page.

The hook is executed after the core has closed the cell holding the file name, so the only place it can write is a cell of its own at the end of the line. The picto is therefore moved into that first cell on load, and the cell the hook had to open is dropped, so the line keeps the columns it had and the picto sits in the same column as the ones of the core. Without javascript nothing moves: the picto stays at the end of the line and still opens the file.

The file name keeps the download link the core gives it, untouched — the picto is an addition to the line, not a replacement.

**Both directions.** The same picto and the same page serve the invoice sent to a customer (`<ref>_cii.xml` on the customer invoice card) and the document received from a supplier (`<ref_supplier>_einvoice.xml` on the supplier invoice card), which is the same need read from the other end: checking what was actually exchanged, without downloading it first.

A document received from an access point usually arrives on a single line, which no one can read, so it is reindented for the display - and only then: a file that already carries its own line breaks, which is what this module generates, is shown exactly as it is on disk. The line above the document says when it was reindented, so what is on screen is never passed off as the bytes of the file. `LIBXML_NONET` because the document comes from outside: the parser must not go and fetch anything.

`htmlspecialchars()` is used rather than `dol_escape_htmltag()` on purpose, and the code says why: `dol_escape_htmltag()` drops the tags it does not keep (an element named `b` or `br` would disappear from the display), turns the newlines into a literal `\n`, and runs `html_entity_decode()` first, so `&#233;` in the file would be shown decoded instead of as written. The point of the page is to show the file as it is.

### What it looks like

The picto on the line of the generated file, in the column where the core puts the ones of the two PDF:

preview picto on the e-invoice XML line

Clicking it opens the preview dialog of the core, the same one a PDF opens in:

the e-invoice XML in the preview dialog

### Tested

On 18.0.10, 20.0.4, 23.0.3 and 24.0.0, through HTTP with a real session:

- both modes return 200 with no PHP notice, and the text printed back is **byte-identical** to the `_cii.xml` on disk (163 to 173 lines depending on the instance);
- the picto and the viewer answer on the **five types of customer invoice** — standard, deposit, credit note, replacement and situation — on each of the four cores: nothing here keys on the type, only on the file being there;
- on the supplier side, the picto is on the received XML of the supplier invoice card and the viewer answers with it, on 20, 23 and 24 (see below for 18);
- the four received documents of the bench arrive on one line and come out on 206 to 225, with the note saying so, and they hold the same document as the file once the whitespace between tags is removed; the generated XML of the same instances stays byte-identical and carries no note;
- the picto appears on exactly one line of the document list — the `_cii.xml` one — the two other files of the same invoice are left alone, and the three download links of the block are unchanged;
- clicking it in a browser opens the core preview dialog on the XML, titled with the file name, on the invoice card of 23.0.3, with the three pictos of the block ending up at the same abscissa and every line keeping its five cells;
- on three invoices that carry only a `_facturx.pdf`, nothing is added and the preview the core gives that PDF is still there;
- an invoice with no e-invoice file shows the "not generated yet" message, and `id=0`, `id=abc`, `id=999999` and a missing `id` all land on the standard "record not found" page, with no notice;
- a user without `facture->lire` nor `fournisseur->facture->lire` gets the standard access-refused page and no XML in the response, in both modes and on both sides; without a session, the login page and no XML.

Module PHPUnit: 16 files x 7 cores (18 to 24), 0 failure.

### Where it is not

The picto is on the document block of the invoice card. It is **not** on the Documents tab of the invoice: that tab renders through `FormFile::list_of_documents()`, whose only hook, `showFilesList`, replaces the whole list instead of completing a line, so there is no insertion point there for a module. The viewer answers on its own URL from anywhere.

On **18.0.10 and 19 the picto does not show on the supplier invoice card**, and it is not this code: the module was writing the received document outside the directory the card reads, so it was listed nowhere. That is #701, fixed by #702, and this PR does not depend on it — the viewer resolves the file the way the module wrote it, and keeps a fallback on the old location so a document received before #702 stays readable. Once #702 is in, the picto appears there like everywhere else.

### Note on #692

@mdeweerd opened #692 for the same issue while I was looking at it, and has since said he moved his version to a private module. This takes the other direction discussed there — a read-only wrapper rather than generating `.txt` twin files — so the two are alternatives, not additions. If #692 is preferred, close this one.


einvoicing/class/actions_einvoicing.class.php
```diff
@@ -1834,4 +1834,110 @@ public function replaceThirdparty($parameters, $object, &$action, $hookmanager)
...
+			$xmlfile = $einvoicing->getEInvoiceXmlFilePath(basename(dirname($relativepath)));
+			$urlparam = '';
+		} elseif ($modulepart == 'facture_fournisseur') {
+			// Received from a supplier: the file is named after the reference of the supplier, so the
+			// invoice has to be read to know it
+			if (!$user->hasRight('fournisseur', 'facture', 'lire') || substr($relativepath, -13) !== '_einvoice.xml') {
+				return 0;
+			}
+			require_once DOL_DOCUMENT_ROOT . '/fourn/class/fournisseur.facture.class.php';
+			$supplierinvoice = new FactureFournisseur($db);
```
eldy: I think in this context, you already have the supplier invoice. it should be  $object   ?

daGrumpf-bxp: Right on the extra read, wrong on where it would come from: in `formBuilddocLineOptions` the `$object` argument is not the invoice, it is the file of the line being rendered. The core builds the call in `FormFile::showdocuments()` as

```php
$parameters = array('colspan' => …, 'socid' => …, 'id' => …, 'modulepart' => $modulepart, 'relativepath' => $relativepath);
$res = $hookmanager->executeHooks('formBuilddocLineOptions', $parameters, $file);
```

— `$file`, one entry of the `dol_dir_list()` of the directory — and the same on 18.0.10, 20.0.4, 23.0.3 and 24.0.0 (`html.formfile.class.php`, lines 927 / 976 / 1110 / 1082). The invoice only reaches the hook as `$parameters['id']`, so there was no object to reuse.

The read itself was avoidable though, and it is gone. `_einvoice.xml` is the suffix `CIIProtocol::saveEInvoiceFileToSupplierInvoiceAttachment()` gives the file it writes beside the invoice, so the name of the line already identifies it; and `xmlpreview.php` fetches the invoice and resolves the path on its own before printing anything, so nothing was gained by resolving it twice. One less query per XML line, and the `$db` global left the method with it.

Re-tested on the bench, Dolibarr 18.0.10 / 22.0.5 / 23.0.3 / 24.0.0: the picto is still rendered on the customer invoice card and on the supplier invoice card, no PHP error, and the preview page still answers 200 with the CII in it.

daGrumpf-bxp:
Pushed a second commit: the source is now **coloured, numbered and foldable**, the way a browser shows an XML document it displays itself. Six hundred lines of one colour was not something anyone was going to read.

- namespace prefix told apart from the local name — on a CII document that alone is most of the readability
- attribute names apart from their values, comments and CDATA apart from the text
- the line number of the file in the margin
- a handle on every element whose end tag is further down, which folds it

`lib/xmlhighlight.lib.php` is deliberately self-contained — no class, no global, no other function of Dolibarr, no external library — so the same two functions can move to the core the day its native preview learns to handle XML.

**Nothing parses the document.** The source is tokenised as text and every token goes through `htmlspecialchars()`, so a truncated or invalid file is still shown as it is on disk instead of producing nothing, and no part of it can act as markup. That is what makes it safe to render at all: the core serves an XML as `application/octet-stream` on purpose, because an inline `text/xml` response from the Dolibarr origin can carry an `<?xml-stylesheet?>` and have the browser render arbitrary HTML in that origin.

`XmlSourceHighlightTest` (new, registered in `AllTests`) pins exactly those two properties over ten sources — a `<script>` in a text value, markup in an attribute value, an `xml-stylesheet` instruction, unbalanced tags, a truncated tag, a comment spanning several lines, a file that is not XML at all — plus the numbering and the fold targets: the only tags of the output are the `div` and the `span` the viewer writes, and the file comes back byte for byte once they are removed.

Run live on the bench, Dolibarr 18.0.10 to 24.0.0: 126/126 PHPUnit runs green (18 files × 7 instances), the picto still opens in the preview dialog of the core on the invoice card, and the same page answers as a full page on its own.


## What it looks like

Dolibarr 24.0.0, the invoice of the bench. The picto lands in the column the core uses for the files it previews itself, so the line reads as a native one:

The preview picto on the line of the CII file

Clicking it opens the preview dialog of the core, the same one a PDF opens in:

The coloured XML inside the preview dialog of the core

And the two first elements folded — the numbers stay the numbers of the file, 5 then 13 then 35:

Two elements folded, the line numbers still those of the file

---

@eldy, on your other point in #692 — *"the true real solution for long term is to have the native preview of dolibarr working into the dolibarr core for xml files"* — I agree, and this is written so that it can be exactly that. What the core would need is small:

1. `xml` in the `$mime_preview` list of `dolIsAllowedForPreview()` (`core/lib/functions.lib.php`), which is what today makes `document.php` force `application/octet-stream` and `FormFile::showPreview()` render no picto;
2. **but not by serving the file inline** — that is the reason it is excluded, and it should stay excluded as a direct response. The preview would point at a small viewer that prints the source escaped, which is what these two functions do.

So the core change is "a viewer page plus one entry in a whitelist", not a rewrite. I am happy to open it on `develop` once this lands here — as @mdeweerd notes, that means Dolibarr 25 and up, so the module keeps needing its own viewer for 18 → 24 either way. Say the word and I will prepare it.

PR #697, opened 2026-09-01
FIX #680 [einvoicing] The lists of Dolibarr stop paying for the joins of the module

### LINKED ISSUE #680: [einvoicing] Invoice list: unindexed LEFT JOIN on llx_einvoicing_extlinks adds seconds to every load and duplicates rows
### Bug

Module version: 1.1.0 (current `main`).

Three defects in the list hooks of `einvoicing/class/actions_einvoicing.class.php`. The first two
share the same root: the `LEFT JOIN` the module adds to the lists of the core is neither indexed
nor restricted to a single row.

## 1. `llx_einvoicing_extlinks` carries no index, and it is joined to the core lists

`ActionsEInvoicing::printFieldListFrom()` adds this join to the customer invoice list
(`einvoicing/class/actions_einvoicing.class.php:1356`):

```sql
LEFT JOIN llx_einvoicing_extlinks AS ext
       ON ext.element_id = f.rowid AND ext.element_type = 'facture'
```

The same method joins that table on the supplier invoice list (`:1360`), the product list (`:1364`)
and the thirdparty list, plus `llx_einvoicing_routing` on the thirdparty list (`:1352`).

None of these tables carries any index besides its `rowid` primary key. `einvoicing/sql/` ships a
`.key.sql` file for `llx_einvoicing_call`, `llx_einvoicing_document` and
`llx_einvoicing_extrafields`, but **not** for `llx_einvoicing_extlinks`,
`llx_einvoicing_lifecycle_msg` and `llx_einvoicing_routing` — which are exactly the three tables
the module joins into the lists of the core.

Resulting plan on the invoice list:

```
table  type  key          rows    Extra
f      ref   idx_entity   10116   Using index
ext    ALL   NULL         18627   Using where; Using join buffer (flat, BNL join)
```

`type: ALL`, `key: NULL`: the whole `extlinks` table is re-scanned for each block of invoice rows.

### The 25-rows-per-page limit does not protect from it

`compta/facture/list.php` builds its record count on the very same `FROM` clause. In 22.0.4:

```php
$sqlfields = $sql;                          // line 779, before the FROM hook
// line 803: printFieldListFrom appends the LEFT JOIN to $sql
$sqlforcount = preg_replace('/^'.preg_quote($sqlfields, '/').'/', 'SELECT COUNT(*) as nbtotalofrecords', $sql);   // line 1124
```

That `COUNT(*)` has no `LIMIT`: it walks the whole invoice set. The page size changes nothing, and
the join is paid twice per page load (count query + display query). Same code in 23.0.2
(lines 802 / 833 / 1188).

## 2. The join duplicates rows and inflates the record count

The join is not restricted to a single row. An element can carry several links: one per provider,
and a provider can be registered both directly and through a partner — a case
`EInvoicing::fetchLastknownInvoiceStatus()` explicitly handles (`$foundforanotherprovider`). Every
extra link produces an extra line in the list.

Measured on 20000 invoices of which 10000 carry 2 links: `COUNT(*)` returns **30000** instead of
20000, and an invoice holding 2 links is displayed **twice**.

A `GROUP BY` cannot repair that count: `list.php` strips the `GROUP BY` from the count query
(`preg_replace('/GROUP BY .*$/', '', $sqlforcount)`, line 1125 in 22.0.4).

Same problem on the thirdparty list: `LEFT JOIN llx_einvoicing_routing rt ON rt.fk_soc = s.rowid`
(`:1352`) filters neither on `routing_type` nor on `active`, so a thirdparty holding several
routing identifiers — or one `thirdparty` routing plus one `product` routing — is listed as many
times.

## 3. `printFieldListWhere()` filters on a column that does not exist

`einvoicing/class/actions_einvoicing.class.php:1392`:

```php
$this->resprints .= " AND ext.routing_id = '" . $db->escape(GETPOST('search_routing_id', 'alpha')) . "'";
```

`llx_einvoicing_extlinks` has no `routing_id` column — its columns are
`provider_sender_routing_id` and `override_routing_id`, while `routing_id` belongs to
`llx_einvoicing_routing` (aliased `rt`). Any request carrying `&search_routing_id=...` on the
invoice, supplier invoice, thirdparty or product list fails with:

```
ERROR 1054 (42S22): Unknown column 'ext.routing_id' in 'where clause'
```

The input is not reachable through the UI today (see the `completeArrayFields` point below), but
the parameter can still be passed in the URL.

## Suggested fix

**a) Ship the missing indexes.** New `einvoicing/sql/llx_einvoicing_extlinks.key.sql`,
`llx_einvoicing_lifecycle_msg.key.sql` and `llx_einvoicing_routing.key.sql`, plus the same
statements in `dolibarr_allversions.sql` so existing installations get them too:

```sql
ALTER TABLE llx_einvoicing_extlinks ADD INDEX idx_einvoicing_extlinks_element (element_type, element_id);
ALTER TABLE llx_einvoicing_extlinks ADD INDEX idx_einvoicing_extlinks_syncref (element_type, syncref);
ALTER TABLE llx_einvoicing_extlinks ADD INDEX idx_einvoicing_extlinks_flowid (flow_id);
ALTER TABLE llx_einvoicing_lifecycle_msg ADD INDEX idx_einvoicing_lifecycle_msg_element (element_type, element_id, lc_validation_status);
ALTER TABLE llx_einvoicing_lifecycle_msg ADD INDEX idx_einvoicing_lifecycle_msg_flowid (flow_id);
ALTER TABLE llx_einvoicing_routing ADD INDEX idx_einvoicing_routing_soc (fk_soc, routing_type, active);
```

**b) Restrict each join to a single row**, keeping the row
`EInvoicing::fetchLastknownInvoiceStatus()` would keep so that the list and the card never
disagree: the link of the currently configured provider first (the `ViaPartner` suffix does not
make it a different provider), then the most recent link.

```php
$providershort = preg_replace('/ViaPartner$/', '', getDolGlobalString('EINVOICING_PDP'));

$sql = ' AND ' . $alias . '.rowid = (SELECT sub.rowid FROM ' . $db->prefix() . 'einvoicing_extlinks as sub';
$sql .= " WHERE sub.element_type = '" . $db->escape($elementtype) . "' AND sub.element_id = " . $elementfield;
$sql .= " ORDER BY CASE WHEN sub.provider IN ('" . $db->escape($providershort) . "', '" . $db->escape($providershort . 'ViaPartner') . "') THEN 0 ELSE 1 END, sub.rowid DESC";
$sql .= ' LIMIT 1)';
```

and the equivalent on `llx_einvoicing_routing`, mirroring `EInvoicing::fetchDefaultRouting()`
(`routing_type = 'thirdparty' AND active = 1 ORDER BY is_default DESC, sub.rowid ASC LIMIT 1`).

Measured on the 20000 invoices / 30000 links dataset: `COUNT(*)` returns the correct 20000 and
runs in **0.0032 s**, against 0.045 s for the current (duplicating) join *with* the index — so
deduplicating is not a trade-off, it is also faster.

**c) Fix the `ext.routing_id` filter**: it should target `rt.routing_id`, and only on the
thirdparty list where `rt` is joined.

## Minor, same area

- **`completeArrayFields()` (`:1236`) never runs.** That hook is not executed anywhere in the
  Dolibarr core tree (checked on 23.0.2, and absent from the hooks executed by
  `compta/facture/list.php` in 22.0.4, which calls `printFieldList*`, `doActions`,
  `finalizeAmountOfInvoice` and `printFieldPreListTitle` only). So the `einvoicegenerated`,
  `pdp_syncstatus`, `routing_id` and `routing_product_id` entries it declares are never registered
  in `$arrayfields`. Consequences: the "E-invoice file" column can never be enabled by the user
  (`printFieldListTitle()` / `printFieldListValue()` test
  `!empty($parameters['arrayfields']['einvoicegenerated']['checked'])`, a key that never exists),
  the sync status column is always displayed (the tests use
  `empty($parameters['arrayfields']['pdp_syncstatus'])` as a fallback), and the
  `search_routing_id` input of `printFieldListOption()` is never printed. Either the hook name is
  wrong, or these guards should be dropped.
- Note that `'checked' => -1` would not have made those columns visible anyway:
  `Form::multiSelectArrayWithCheckbox()` normalises a negative `checked` to `0`.
- `printFieldListSelect()` adds `ext.*` and `rt.routing_id` for the `societelist` context (`:1325`)
  while `printFieldListFrom()` has no `societelist` branch, so both aliases would be undefined.
  `societelist` is not initialised anywhere in the core, so this is dead code today.
- `printFieldListFrom()` joins `extlinks` for `productservicelist` (`:1364`) while
  `printFieldListSelect()` selects nothing from it in that context: a pure-overhead join.
- `printFieldListSelect()` and `printFieldListFrom()` add the columns and the join
  unconditionally, testing neither `EINVOICING_DISABLE_SYNC_DOLI_TO_AP` nor whether the columns are
  checked, whereas `printFieldListOption()`, `printFieldListTitle()` and `printFieldListValue()` do
  test them. Unchecking the columns therefore cannot avoid the cost.

### Module

einvoicing

### Dolibarr Version

22.0.4

### Environment PHP

8.2.33

### Environment Database

MariaDB 10.11.18-0+deb12u1

1. Install einvoicing 1.1.0 and configure `EINVOICING_PDP` and `EINVOICING_PROTOCOL`.
2. Use a database holding a few thousand customer invoices, with rows in
   `llx_einvoicing_extlinks` for `element_type = 'facture'`.
3. Open **Billing | Customer invoices | List**, 25 rows per page.

**Observed:** the page takes seconds longer to load than with the module disabled. `EXPLAIN` on the
list query reports `type: ALL`, `key: NULL` and `Using join buffer (flat, BNL join)` on `ext`.
Adding a second `llx_einvoicing_extlinks` row for one invoice makes that invoice appear twice in
the list and increases the total record count by one.

**Expected:** the module adds a negligible cost to the list, and each invoice appears once.

Standalone reproduction of the timings, without a Dolibarr instance (MariaDB, temporary tables, so
nothing is written to a real schema):

```sql
SET SESSION max_recursive_iterations = 200000;

CREATE TEMPORARY TABLE f5 (rowid int PRIMARY KEY, entity int, datef date, KEY k(entity)) ENGINE=InnoDB;
INSERT INTO f5 WITH RECURSIVE s AS (SELECT 1 n UNION ALL SELECT n+1 FROM s WHERE n<16343)
  SELECT n, 1, DATE_ADD('2024-01-01', INTERVAL (n%700) DAY) FROM s;

CREATE TEMPORARY TABLE e5 (rowid int AUTO_INCREMENT PRIMARY KEY, element_id int,
  element_type varchar(50) NOT NULL, provider varchar(50) NOT NULL, syncstatus int) ENGINE=InnoDB;
INSERT INTO e5 (element_id, element_type, provider, syncstatus)
  WITH RECURSIVE s AS (SELECT 1 n UNION ALL SELECT n+1 FROM s WHERE n<16343)
  SELECT n, 'facture', 'Esalink', n%9 FROM s;

-- the join as shipped: 14.2 s
SELECT COUNT(*) FROM f5 f
  LEFT JOIN e5 ext ON ext.element_id = f.rowid AND ext.element_type = 'facture'
 WHERE f.entity = 1;

ALTER TABLE e5 ADD INDEX idx_element (element_type, element_id);

-- same query with the missing index: 0.03 s
SELECT COUNT(*) FROM f5 f
  LEFT JOIN e5 ext ON ext.element_id = f.rowid AND ext.element_type = 'facture'
 WHERE f.entity = 1;
```

Row duplication, same tables:

```sql
INSERT INTO e5 (element_id, element_type, provider, syncstatus)
  VALUES (42, 'facture', 'SuperPDPViaPartner', 3);

-- returns 2: invoice 42 is listed twice, and counted twice
SELECT COUNT(*) FROM f5 f
  LEFT JOIN e5 ext ON ext.element_id = f.rowid AND ext.element_type = 'facture'
 WHERE f.rowid = 42;
```

And the non-existent column:

```sql
-- ERROR 1054 (42S22): Unknown column 'ext.routing_id' in 'where clause'
SELECT COUNT(*) FROM llx_facture f
  LEFT JOIN llx_einvoicing_extlinks ext ON ext.element_id = f.rowid AND ext.element_type = 'facture'
 WHERE f.entity IN (1) AND ext.routing_id = 'X';
```

### Attached files

_No response_

daGrumpf-bxp:
Fixes #680.

The module joins `llx_einvoicing_extlinks` and `llx_einvoicing_routing` into the invoice, supplier
invoice and thirdparty lists of the core. Neither table carried an index, neither join was restricted
to a single row, and the search filter on the routing identifier read a column that does not exist.

Everything below was measured on a bench of eight instances, Dolibarr 17.0.4 to 24.0.0-beta, sharing
one checkout of the module, PHP 8.4 / MariaDB 11.4.

## 1. The missing indexes

Three new `.key.sql` files, and the same statements in `dolibarr_allversions.sql` for installations
that upgrade the core rather than reactivate the module. Every index answers a read the module
already performs: `(element_type, element_id)` on every card, `(element_type, syncref)` for
`fetchLastknownInvoiceStatus()` when the id is unknown, `flow_id` for the messages of the Access
Point, `(fk_soc, routing_type, active)` for `fetchDefaultRouting()`.

The record count of a list is the query that hurts, because `list.php` builds it on the same `FROM`
clause and gives it no `LIMIT`. Measured on 20 100 invoices carrying 30 101 links:

| state | `COUNT(*)` returns | time |
|---|---|---|
| as shipped, no index | 30 100 | **32.9 s** |
| as shipped, with the index | 30 100 | 0.046 s |
| deduplicated join, with the index | **20 100** | **0.024 s** |

The user-visible effect on the same data set, on Dolibarr 22.0.5: `/compta/facture/list.php?limit=25`
answered **HTTP 504 after 60 s**, three times in a row. With this branch it answers in **0.036 s**.
`EXPLAIN` goes from `type: ALL, key: NULL, rows: 496, Using join buffer (flat, BNL join)` on `ext`
to `type: eq_ref, key: PRIMARY, rows: 1`; on the count query MariaDB now eliminates the join
altogether, since it is provably at most one row and no column of it is read.

## 2. The join repeated rows and inflated the count

An element can carry several links: one per provider, and a provider registered both directly and
through a partner. The row kept is now the one `fetchLastknownInvoiceStatus()` would keep, so a list
and a card never disagree; same idea on the routing side with `fetchDefaultRouting()`.

On the bench, before → after, without touching the data:

- customer invoice list, Dolibarr 24: `COUNT(*)` 137 for 118 invoices → **118**; first page 100 row
  links for 90 distinct invoices → **100 / 100**.
- thirdparty list, Dolibarr 23 and 24: 13 links for 12 thirdparties → **12 / 12** (one thirdparty
  holding two routing identifiers).
- supplier invoice list: a second link added on one invoice made it appear twice (200 rows for 199
  invoices) → **200 / 200**, and the kept row is the one of the configured provider.

## 3. `ext.routing_id` does not exist

`printFieldListWhere()` filtered on `ext.routing_id`, a column of `llx_einvoicing_routing` (aliased
`rt`), not of `llx_einvoicing_extlinks`. Any URL carrying `&search_routing_id=` answered
`Unknown column 'ext.routing_id' in 'WHERE'` — reproduced on **all eight versions**, on the invoice,
supplier invoice, thirdparty and product lists. The filter now reads `rt.routing_id`, and only on the
thirdparty list where `rt` is joined.

Note that it therefore matches the routing identifier the list displays — the default one. Searching
a secondary identifier of a thirdparty returns nothing. An `EXISTS` on the whole table would answer
"which thirdparty holds this identifier" instead, at the price of a filter that no longer agrees with
the column; I kept the agreement, and I will gladly switch if you prefer the other reading.

## 4. The columns of the module were unreachable on the older cores

`completeArrayFields()` is the hook meant to declare a column, but the core only calls it from
Dolibarr 22 — and there only on the customer invoice list; the four lists get it from 23. Below that,
the declared fields were never registered, so the "E-invoice File" column could not be displayed at
all, and `arrayfields['einvoicegenerated']` — a key the thirdparty list never declares — gated its
routing column, which therefore never appeared on any version.

The core has passed that same `$arrayfields` **by reference** in the parameters of `doActions` for
much longer: since 17 on the customer invoice list, 19 on the supplier invoice list, 20 on the
thirdparty and product lists. The declaration moved to a method both hooks call, and the thirdparty
column is gated on the field name that list actually declares.

Column offered in the "select columns" menu, before → after:

| list | before | after |
|---|---|---|
| customer invoice, "E-invoice File" | 22, 23, 24 | **18 → 24** |
| thirdparty, "E-invoicing address" | declared on 23/24 but never rendered | **20 → 24**, rendered |

## 5. Consistency between the clauses

The `SELECT`, the `FROM`, the `WHERE` and the `GROUP BY` now answer the same question as the columns
they feed — a single `isListCompletedByModule()`. Consequences:

- the product list no longer joins `llx_einvoicing_extlinks`: nothing was ever selected from it there;
- with `EINVOICING_DISABLE_SYNC_DOLI_TO_AP` set, the invoice list now runs **zero** query touching
  `einvoicing_extlinks`, where the join used to be added regardless;
- the `societelist` branches are gone. `societe/list.php` calls
  `initHooks(array($contextpage, 'thirdpartylist'))`, so `thirdpartylist` is always present even with
  `?contextpage=societelist`: that branch only made the query select `pdplink_id` twice. Removing it
  is the one deletion here that is not strictly a bug fix — say the word and I put it back.

## Test batch

- Four lists × eight instances (17 → 24), with and without the search filters: 200 everywhere, no SQL
  error, and as many distinct rows as rows. On `main` the same sweep answers `DB_ERROR_NOSUCHFIELD`
  on both filter URLs of every version, and repeats rows on 22, 23 and 24.
- Both lists checked under `sql_mode=ONLY_FULL_GROUP_BY`.
- Both `EINVOICING_DISABLE_SYNC_*` options set, with the filters passed in the URL: no error.
- Module deactivated and reactivated on the seven instances 18 → 24: the six indexes are created, and
  replaying it a second time is a no-op.
- PHPUnit, every test file on every instance: 128 runs, 127 OK. The single failure
  (`EInvoicingSamplesTest` on the 17.0.4 instance, where the module is not enabled) fails identically
  on `main`.
- `phpcs` with the ruleset of the repository: clean.

### DIFF
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -1245,7 +1245,7 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 	 */
 	public function completeArrayFields($parameters, $object, &$action, $hookmanager)
 	{
-		if (in_array('invoicelist', explode(':', $parameters['context'])) && !getDolGlobalString('EINVOICING_DISABLE_SYNC_DOLI_TO_AP')) {
+		if (self::isListCompletedByModule($parameters['context'], 'invoicelist')) {
 			// Add fields to invoice list
 			$parameters['arrayfields']['einvoicegenerated'] = array(
 				'label' => 'EInvoiceFile',
@@ -1263,8 +1263,8 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 			);
 		}
 
-		if (in_array('thirdpartylist', explode(':', $parameters['context'])) && (!getDolGlobalString('EINVOICING_DISABLE_SYNC_DOLI_TO_AP') || !getDolGlobalString('EINVOICING_DISABLE_SYNC_AP_TO_DOLI'))) {
-			// Add fields to invoice list
+		if (self::isListCompletedByModule($parameters['context'], 'thirdpartylist')) {
+			// Add fields to thirdparty list
 			$parameters['arrayfields']['routing_id'] = array(
 				'label' => 'RoutingIdField',
 				'help' => 'SpecificRoutingFieldHelp',
@@ -1286,6 +1286,94 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 	}
 
 
+
+	/**
+	 * Tell whether the module completes a given list of the core with its own columns.
+	 *
+	 * The SELECT, the FROM, the WHERE and the GROUP BY must all answer the same thing: a column read
+	 * from, or a filter set on, a table that the FROM did not join turns the page of the core into an
+	 * SQL error. Same conditions as the ones printFieldListTitle() and printFieldListValue() use to
+	 * decide whether they display something.
+	 *
+	 * @param 	string	$context	Value of $parameters['context'] as the hook manager builds it
+	 * @param 	string	$list		List to test ('invoicelist', 'supplierinvoicelist' or 'thirdpartylist')
+	 * @return 	bool				True if the module adds its columns and its joins to that list
+	 */
+	protected static function isListCompletedByModule($context, $list)
+	{
+		if (!in_array($list, explode(':', $context), true)) {
+			return false;
+		}
+
+		if ($list == 'invoicelist') {
+			return !getDolGlobalString('EINVOICING_DISABLE_SYNC_DOLI_TO_AP');
+		}
+
+		if ($list == 'supplierinvoicelist') {
+			return !getDolGlobalString('EINVOICING_DISABLE_SYNC_AP_TO_DOLI');
+		}
+
+		return !getDolGlobalString('EINVOICING_DISABLE_SYNC_DOLI_TO_AP') || !getDolGlobalString('EINVOICING_DISABLE_SYNC_AP_TO_DOLI');
+	}
+
+	/**
+	 * Build the condition restricting the join on einvoicing_extlinks (aliased 'ext') to a single row.
+	 *
+	 * An element can carry several links: one per provider, and a provider can be registered both
+	 * directly and through a partner. Without this condition every extra link adds a line to the list
+	 * and, because the core builds its record count on the same FROM clause, an invoice is counted as
+	 * many times as it has links. The row kept here is the one EInvoicing::fetchLastknownInvoiceStatus()
+	 * would keep, so the list and the card never disagree.
+	 *
+	 * @param 	'facture'|'invoice_supplier'|'societe'	$elementtype	Value of einvoicing_extlinks.element_type, which also tells which list of the core is being built
+	 * @return 	string									Condition to append to the ON clause of the join
+	 */
+	protected static function getExtLinkJoinCondition($elementtype)
+	{
+		global $db;
+
+		// The 'ViaPartner' suffix tells how the provider is reached, not which provider it is.
+		$providershort = preg_replace('/ViaPartner$/', '', getDolGlobalString('EINVOICING_PDP'));
+		$sqlcurrentprovider = "sub.provider IN ('" . $db->escape($providershort) . "', '" . $db->escape($providershort . 'ViaPartner') . "')";
+
+		$sql = ' AND ext.rowid = (SELECT sub.rowid FROM ' . $db->prefix() . 'einvoicing_extlinks as sub';
+		$sql .= " WHERE sub.element_type = '" . $db->escape($elementtype) . "'";
+		if ($elementtype == 'societe') {
+			$sql .= ' AND sub.element_id = s.rowid';	// Alias of the thirdparty in the thirdparty list of the core
+		} else {
+			$sql .= ' AND sub.element_id = f.rowid';	// Alias of the invoice in both invoice lists of the core
+		}
+		$sql .= ' ORDER BY CASE WHEN ' . $sqlcurrentprovider . ' THEN 0 ELSE 1 END,';
+		// fetchLastknownInvoiceStatus() keeps the LAST link read for the configured provider, but the
+		// FIRST one read for another provider, so the two cases do not order the rowid the same way.
+		$sql .= ' CASE WHEN ' . $sqlcurrentprovider . ' THEN -sub.rowid ELSE sub.rowid END';
+		$sql .= ' LIMIT 1)';
+
+		return $sql;
+	}
+
+	/**
+	 * Build the condition restricting the join on einvoicing_routing (aliased 'rt') to a single row.
+	 *
+	 * A thirdparty can hold several routing identifiers, and a routing row for its default product on
+	 * top of them, so the join repeats the thirdparty in the list as many times. The row kept here is
+	 * the one EInvoicing::fetchDefaultRouting() and EInvoicing::fetchAllRoutings() put first.
+	 *
+	 * @return 	string				Condition to append to the ON clause of the join, correlated on s.rowid
+	 */
+	protected static function getRoutingJoinCondition()
+	{
+		global $db;
+
+		$sql = ' AND rt.rowid = (SELECT sub.rowid FROM ' . $db->prefix() . 'einvoicing_routing as sub';
+		$sql .= ' WHERE sub.fk_soc = s.rowid';	// Alias of the thirdparty in the thirdparty list of the core
+		$sql .= " AND sub.routing_type = 'thirdparty'";
+		$sql .= ' AND sub.active = 1';
+		$sql .= ' ORDER BY sub.is_default DESC, sub.rowid ASC';
+		$sql .= ' LIMIT 1)';
+
+		return $sql;
+	}
 
 	/**
 	 * Build the sub query returning the last lifecycle status validated by the Access Point for a supplier invoice.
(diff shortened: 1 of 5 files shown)


eldy:
PR with database structure change must be suggested in an independent PR.

The PR contains 5 different features. Must be spliited in 5 PR.

daGrumpf-bxp:
Done — split, and the database structure change is on its own.

| PR | concern |
|---|---|
| #713 | the three tables carry an index (**the DB structure change, alone**) |
| #714 | `ext.routing_id` does not exist: any `&search_routing_id=` is an SQL error |
| #715 | the `einvoicing_extlinks` join returns several rows and inflates the list and its count |
| #716 | the `einvoicing_routing` join does the same on the thirdparty list |
| #717 | the columns are unreachable on the cores that have no `completeArrayFields` hook |

All five are opened on `main` and are **mergeable in any order**: I merged them into `main` in all 120 permutations before opening, no conflict in any of them, and the result parses in every one. Each was also re-tested on its own on the bench of seven instances (Dolibarr 18.0.10 → 24.0.0), plus once with the five applied together.

Closing this one in their favour.

PR #699, opened 2026-09-01
FIX: do not forget $db in your object

altairis-noe:
FIX: do not forget $db in your object

$db property is defined but not set in PDPProvider::__construct().

So, when we call for a new HookManager in addProviderFromHooks() method, $this->db is null, and therefore an error 500 will occur if any module pretends to interact with this Hook !

### DIFF
--- a/einvoicing/class/providers/PDPProviderManager.class.php
+++ b/einvoicing/class/providers/PDPProviderManager.class.php
@@ -50,6 +50,8 @@ class PDPProviderManager
 		global $langs, $mysoc;
 		global $dolibarr_main_url_root;
 
+		$this->db = $db;
+
 		// Define $urlwithroot
 		$urlwithouturlroot = preg_replace('/'.preg_quote(DOL_URL_ROOT, '/').'$/i', '', trim($dolibarr_main_url_root));
 		$urlwithroot = $urlwithouturlroot.DOL_URL_ROOT; // This is to use external domain name found into config file


daGrumpf-bxp:
Confirmed, and thanks for catching it — the fix is right and in the right place (before `addProvidersFromHooks()`).

Some context on how it slipped through, since it is entirely on me:

- `public $db` has been declared on `PDPProviderManager` since the PDPConnect -> Einvoicing rename (`1aa10a2a`), but never assigned. Nothing noticed, because every other place in the class works on `global $db` — `getProvider()` still does `new $classnametouse($db)` from the global.
- The provider hook I added in `4438a3b4` (#448) is the first and only consumer of `$this->db`, at `new HookManager($this->db)`. So the hook manager was built with `null`, and `initHooks()` then instantiates the external module's `actions_*` class with that `null` — hence the 500 as soon as that module touches its own database.
- And `PDPProviderManagerTest` covers the *result* of the hook, not the hook: `providersList` is private and filling it for real needs a third-party module installed, so the tests inject the entries by reflection. No test ever ran a real `HookManager`, so no test could ever see this. You are the first person to plug an actual module into that context, and it broke on the very first thing it does.

So: the sample provider (`TestPDPProvider` + `doc/ADD-A-PDP-PROVIDER.md`) worked because it is instantiated by `getProvider()` with the global `$db` — the hole was on the declaring side, which nothing exercised. Sorry for the wasted debugging time.

LGTM.

altairis-noe:
Thanks @daGrumpf-bxp, for the detailed answer. I am not surprised how this glitch might have occured.

We stumbled upon this because we used another module with an 'all' selector for hooks in its module descriptor (which I will advise the editor not to do) 😅

Issue #703, opened 2026-09-01
[einvoicing] The version of the downloaded einvoicing module does not match the declared module version.


hansemschnokeloch:
### Bug

See screenshot, in Dolibarr the einvoicing module is announced as v1.1.0 but v1.0.4 is downloaded or installed

<img width="2303" height="1313" alt="Image" src="https://github.com/user-attachments/assets/012f8d94-2ad3-45aa-85ec-92accf746036" />

### Module

Einvoicing

### Dolibarr Version

24

### Steps to reproduce the behavior and expected behavior

Download einvoicing module from dolibarr

daGrumpf-bxp:
Confirmed, and here is what the two published packages actually contain.

**@hansemschnokeloch: could you tell us which link you followed?** The "Deploy/install external module" screen of Dolibarr, the *Download* button of the module catalogue, dolistore.com directly, or a GitHub archive? The three routes do not serve the same thing today, so knowing the exact one you used pins down which of them has to be fixed first.

### What each source serves right now

**1. Dolistore package** — `https://www.dolistore.com/_service_download.php?t=free&p=3309` (the URL behind `dolistore-download` in `einvoicing/index.yaml`)
- 3.65 MB, root folder `einvoicing/`, `VERSION` = **1.0.4**
- All 2111 files are byte-identical (git blob hashes) to `einvoicing/` at commit **`0e0481d` "Clean descriptor", 2026-08-19 18:12 +0200**; only dev-only files (`build/`, `doc/`, `test/`, `.gitignore`, `composer.disabled.lock` and 87 vendor dev files) are stripped
- Zip timestamps: **2026-08-19 18:57**, i.e. packaged ~45 min after that commit
- That commit is **not tagged** — there is no `einvoicing_1.0.4` tag, so the shipped package matches no released version
- `main` has since received **185 commits** touching `einvoicing/`

**2. GitHub tag archive** — `https://github.com/Dolibarr/dolibarr-community-modules/archive/refs/tags/einvoicing_1.1.0.zip`
- 11.3 MB, commit **`60cefed`, 2026-08-30 21:09 +0200**, `einvoicing/VERSION` = **1.1.0**
- But it is the **whole repository** (every module, plus `.DS_Store`), under a root folder `dolibarr-community-modules-einvoicing_1.1.0/` — so it cannot be fed to Dolibarr's module installer as-is

[...]

Issue #537, opened 2026-08-03
[einvoicing] Error on enabling the module in Dolibarr LTS 18.0.10


emheyarssi:
### Bug

When enabling the module an error 500 is returned due to the missing `CommonHookActions` class

### Module

einvoicing

### Dolibarr Version

18 LTS

### Environment PHP

7.4.33

### Steps to reproduce the behavior and expected behavior

Download einvoicing 1.0.3 module from dolistore, install it and enable it

daGrumpf-bxp:
Confirmed, and it is a packaging bug, not a source bug — so re-downloading from another place will not help.

### What happens

`einvoicing/class/actions_einvoicing.class.php` needs the `CommonHookActions` class, which core only
ships from Dolibarr **19.0** on (`htdocs/core/class/commonhookactions.class.php` is absent from the
`17.0` and `18.0` branches). That is why the module carries its own copy under `einvoicing/compat/`
and loads it on old versions (this is the fix that closed #153):

```php
if ((float) DOL_VERSION < 19) {
	dol_include_once('/einvoicing/compat/commonhookactions.class.php');
} else {
	require_once DOL_DOCUMENT_ROOT . '/core/class/commonhookactions.class.php';
}
```

That code is in `main` and has been since May. **The `compat/` directory is simply not in the
distributed zip.** The package builder `dev/build/makepack-modules.php` copies an explicit allow-list
of directories:

```php
$listOfModuleContent = [
	'admin', 'ajax', 'backport', 'class', 'css', 'core', 'img', 'js', 'langs',
	'lib', 'public', 'scripts', 'sql', 'tpl', 'vendor',
	'*.md', '*.json', '*.php', '*.yaml', 'COPYING', 'COPYRIGHT', 'modulebuilder.txt',
];
```

`compat` is not in it, so `dol_include_once()` finds nothing, and the very next page load fatals on
`class ActionsEInvoicing extends CommonHookActions`.

### Reproduced

Virgin Dolibarr **18.0.10**, no other module, `module_einvoicing-1.0.3.zip` unzipped into `custom/`,
module enabled:

```
DOL_VERSION = 18.0.10
core/class/commonhookactions.class.php present in core: NO
custom/einvoicing/compat/ present: NO
activateModule errors: (none)
--- now loading hooks like any page does ---
PHP Fatal error:  Uncaught Error: Class "CommonHookActions" not found in
[...]

PR #706, opened 2026-09-01
NEW: #703 [einvoicing] A tag publishes the installable package of the module

daGrumpf-bxp:
Refs #703 — not an automatic close: what #703 reported is a distribution lag, and the fix for that is an upload, not a commit.

**Rebased on `main` after @eldy's review.** The half of this PR that removed `einvoicing/build/buildzip.php` is gone — the build dirs were removed upstream in the meantime, and git dropped that commit as already applied. What is left is one file.

## The one file

`.github/workflows/module_einvoicing_release.yml`, triggered by a `einvoicing_*` tag (and by `workflow_dispatch`, to see what a release would ship without publishing anything).

It does not build the package its own way. It runs **`dev/build/makepack-modules.php makezip einvoicing`** — the script the maintainer runs — on the tag's own checkout, so the release carries the package that was tested and not a second build made another way.

Then, before publishing:

1. **the tag and `einvoicing/VERSION` must agree.** The tag carries the version after the underscore, the descriptor reads it from the `VERSION` file at runtime; two sources, so they are checked against each other before anything is published under that number.
2. **the package must be installable.** `VERSION`, `COMMIT`, `COPYING`, `index.yaml`, the descriptor, `compat/functions.lib.php` and `vendor/autoload.php` must be in it, and every entry must sit under a single `einvoicing/` root. That list is the #537 checklist: without `compat/`, the unconditional `require_once` of `lib/einvoicing.lib.php:30` is fatal, `admin/modules.php` answers **500**, and the user cannot even disable the module they just installed.

Finally it uploads the package as a build artifact, and attaches it to the GitHub release of the tag under the name the builder gave it — `module_einvoicing-<version>.zip`, the only shape the deployment screen accepts.

## Why an asset at all

The archive GitHub generates for a tag is the whole repository, under a root directory named after the tag. Dolibarr refuses it twice over: *Home - Setup - Modules* matches the file name against `/^(module[a-zA-Z0-9]*_|theme_|).*\-([0-9][0-9\.]*)(\s\(\d+\)\s)?\.zip$/i`, which `einvoicing_1.1.0.zip` does not satisfy, and unpacking it would not put an `einvoicing` directory into `htdocs/custom` anyway. The installable package is the one `makepack-modules.php` builds, it is committed under `dev/build/bin/` and uploaded to Dolistore by hand — which is how a 1.0.4 package came to be served while the catalogue announced 1.1.0.

## Open on purpose

Per the discussion below, the release tooling of this repository is still being written, and this workflow only covers the tag half of what is planned — not the rolling release. Adapting it (verify the committed zip instead of rebuilding it; or trigger on `main` instead of on a tag) is a small change, and closing it is fine too. It waits on the direction, not the other way round.

### DIFF
--- a/.github/workflows/module_einvoicing_release.yml
+++ b/.github/workflows/module_einvoicing_release.yml
@@ -1,0 +1,129 @@
+---
+# Attaches the installable package of the einvoicing module to the GitHub release of its tag, so a
+# version has one artifact everybody can point at.
+#
+# Why this exists: the archive GitHub generates for a tag is the whole repository, under a root
+# directory named after the tag. Dolibarr refuses it twice over - the deployment screen matches the
+# file name against /^(module[a-zA-Z0-9]*_|theme_|).*\-([0-9][0-9\.]*)(\s\(\d+\)\s)?\.zip$/i, which
+# "einvoicing_1.1.0.zip" does not satisfy, and unpacking it would not put an "einvoicing" directory
+# in htdocs/custom. The package people can install is the one dev/build/makepack-modules.php builds,
+# and nothing published it: it is committed under dev/build/bin/ and uploaded to Dolistore by hand,
+# which is how a 1.0.4 package came to be served while the catalogue announced 1.1.0 (issue #703).
+#
+# The job builds with that same script - the one that produces what ships - so the release carries
+# the package that was tested, not a second build made another way. It packages what the tag already
+# contains and never edits a file of the module.
+on:
+  push:
+    tags:
+      - 'einvoicing_*'
+
+  # Lets a maintainer build the package of any branch without publishing anything, to check what a
+  # release would ship.
+  workflow_dispatch:
+
+permissions:
+  contents: write
+
+concurrency:
+  group: release-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
+
+    steps:
+      # makepack-modules.php stamps the package with the commit it was built from, so the checkout
+      # needs a repository and not only the files.
+      - uses: actions/checkout@v6
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 8.3
+          extensions: zip
+
+      # The tag carries the version after the underscore ("einvoicing_1.1.0"), the module carries it
+      # in its VERSION file, and the descriptor reads that file at runtime. Two sources, so check
+      # they agree before anything is published under that number.
+      - name: Check the tag and the VERSION file agree
+        if: startsWith(github.ref, 'refs/tags/')
+        run: |
+          set -euo pipefail
+          file_version=$(tr -d '[:space:]' < einvoicing/VERSION)
+          tag_version=${GITHUB_REF_NAME#einvoicing_}
+          echo "VERSION file: $file_version, tag: $tag_version"
+          if [ "$file_version" != "$tag_version" ]; then
+            echo "::error::einvoicing/VERSION says $file_version but the tag says $tag_version"
+            exit 1
+          fi
+
+      - name: Build the module package
+        id: build
+        run: |
+          set -euo pipefail
+          version=$(tr -d '[:space:]' < einvoicing/VERSION)
+          package="dev/build/bin/module_einvoicing-$version.zip"
+          # The script leaves a package alone when a tag already names its version - which is always
+          # the case here, since a tag is what started the job. Removing it asks for the rebuild.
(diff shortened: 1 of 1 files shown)


daGrumpf-bxp:
Checks: `build` (phpcs), the PDF/A-3 job and the labeller are green. `Run phan` is red, but it is red on `main` too since 9e1edd07 (16:15 UTC today) and its four diagnostics are in `scripts/checksupplierinvoicestatus.php` and `class/einvoicing.class.php` — none in the two files this PR touches.

eldy:
The "buildzip" script was a relic from the age of the dinosaurs.
It has been abandoned in favor on makepack-dolibarrmodules.php
I removed all the build dirs.

eldy:
There was a difference between tag and zip because the tool to make package and to make release was not yet ready and we need to male a release urgently.
With makepack it should not happen anymore because the zip and tag are generated by same script in same process.
Still manually.

makepack will just be run to build the zip with param makeziptag by the maintener when he want to make a release.

For the moment what we target is
- build package by maintener on demand + push on the dolistore account of the maintener (we want a CLI tool for that). This will set a tag, build the zip and push on dolistore.
- having a rolling release (no tag) for the install/update from Dolibarr (this may be a github action for example and version should be increased compared to last tag)

daGrumpf-bxp:
Thanks — that settles it, and I have rebased the PR on `main` accordingly.

**The `buildzip.php` half is gone from the PR**: you had already removed the build dirs, so git dropped that commit as "already upstream" on the rebase. Good riddance — it was the relic, and the one that would have shipped a package without `compat/` and `VERSION`.

**What is left is one file**: `.github/workflows/module_einvoicing_release.yml`. It does not build anything its own way — it runs `dev/build/makepack-modules.php makezip einvoicing`, the same script you run, on the tag's own checkout, then checks the package (`VERSION`, `COMMIT`, `compat/`, `index.yaml`, `vendor/autoload.php`, everything under a single `einvoicing/` root — the #537 checklist) and attaches it to the GitHub release of that tag. So it publishes the artifact of *your* tool rather than a second one.

Given what you describe, that is a partial fit and I am happy to adapt it or park it, whichever you prefer:

- **the tag half** overlaps with `makeziptag` doing zip and tag in the same commit. If the zip is already committed by then, the job should probably *verify* it and attach it, not rebuild it. One-line change, say the word.
- **the rolling-release half is not covered** and is what you say you actually want next. The same job, triggered on push to `main` instead of on a tag, with the version bumped past the last tag, is the shape you described — I can turn it into that instead if that is the direction.
- and if the tooling is still moving too much for either to be useful now, closing this is fine by me. It is one file and it will keep.

No hurry on my side — I would rather it lands the way your release process ends up working than get merged early and get in the way.

daGrumpf-bxp:
Closing this, since #703 is closed and the tooling that landed since covers what it asked for.

**The two packages are now the same object.** The Dolistore download (`_service_download.php?t=free&p=3309`) and `dev/build/bin/module_einvoicing-1.2.0.zip` committed here are byte-identical:

| | Dolistore | this repository |
|---|---|---|
| size | 3 881 880 | 3 881 880 |
| md5 | `c10a4cf593fa811adbe175f932ef07c6` | `c10a4cf593fa811adbe175f932ef07c6` |
| sha256 | `d380210bf58951691e91c8a84b6004231f8ce7c5f2bc1ddd55da83480a5c339c` | same |
| entries | 2446 | 2446 |
| `VERSION` / `COMMIT` | 1.2.0 / `dd1cdc2e` | 1.2.0 / `dd1cdc2e` |

So the desynchronization reported in #703 — 1.0.4 served while the catalogue announced 1.1.0 — is gone, and the upload leg I was asking for is now `makepack-modules.php pushdolistore`.

**Against `main`:** the packaged `COMMIT` `dd1cdc2e` is your merge of 08/09 21:08, and the zip was committed one minute later in c24c52fb *"Version 1.2.0"*. `main` is at d7bfdd64 today, 56 commits further, 50 of them touching `einvoicing/` (32 files, +1960/-316), with `VERSION` still 1.2.0. That gap is the ordinary rolling-release lag, not a mismatch between channels.

**And the workflow no longer fits the process:**

- `makeziptag` builds the zip and sets the tag in the same run, so rebuilding on tag in CI would be a second build of a package that is already committed;
- 1.2.0 shipped without a tag, exactly as you described the rolling release, so the `einvoicing_*` trigger would not fire;
- the install path reads the committed zip — `externalModules.class.php` builds `.../raw/refs/heads/main/dev/build/bin/module_<mod>-<version>.zip` — and nothing reads a GitHub release asset.

Nothing worth carrying, then. Thanks for the direction.

PR #707, opened 2026-09-01
FIX #704 [einvoicing] The registry check keeps quiet on the fields INSEE does not disclose

### LINKED ISSUE #704: [einvoicing] erreur au contrôle des données du tier via API gouvernementales quand les données sont NON-DIFFUSIBLE
### Bug

Bonjour,

J'ai quelques clients qui ont demandé à ne pas publier certaines données sur leur entreprise.
Cela entraîne le retour d'une valeur NON-DIFFUSIBLE ce qui déclenche les alertes en cas de contrôle des données à la génération de la facture :
Le nom du tiers (NOM DE LA SOCIÉTÉ) diffère du nom officiel dans le registre national des entreprises françaises ([NON-DIFFUSIBLE]) ; pensez à le mettre à jour.
Warning API: Le code postal enregistré (69100) ne correspond pas à celui du registre national des entreprises françaises ([NON-DIFFUSIBLE]).
Warning API: La ville enregistrée (Villeurbanne) ne correspond pas à celle du registre national des entreprises françaises (LYON).

il faudrait ne pas lever d'alerte quand la valeur reçue de l'API est [NON-DIFFUSIBLE]

### Module

einvoicing

### Dolibarr Version

22.0.5

### Environment PHP

7.4

### Environment Database

MariaDB

option "Activer la validation des données des tiers via les API gouvernementales (SIREN, TVA)" : TRUE
génération d'une facture sur un client dont les données sont NON-DIFFUSIBLE

### Attached files

_No response_

daGrumpf-bxp:
Fixes #704.

### The report

With `EINVOICING_ENABLE_API_VALIDATION` on, generating an invoice for a customer whose INSEE data are not public raises warnings like:

> The third party name (MY COMPANY) differs from the official name in the French national business registry (**[NON-DIFFUSIBLE]**).
> The ZIP code on record (69100) does not match the one in the French national business registry (**[NON-DIFFUSIBLE]**).

### What is going on

`[NON-DIFFUSIBLE]` is not a value, it is a placeholder. Units that exercised their right to opposition (art. A123-96 of the French commercial code) carry the *partial diffusion* status (`statut_diffusion = "P"`, which replaced the old *non-diffusible* status in 2023, decree 2022-1014). `recherche-entreprises.api.gouv.fr` still returns them, but every protected field comes back as the literal string `[NON-DIFFUSIBLE]`: `nom_complet`, `nom_raison_sociale`, `sigle`, and on the establishment `adresse`, `code_postal`, `numero_voie`, `libelle_voie`, `cedex`, `nom_commercial`, `liste_enseignes`…

Comparing that placeholder against the third party record can only ever mismatch, so the customer got a warning on every single invoice. Nothing in EN 16931 or in the French CTC-FR rules requires the party name or address to match a registry — `BR-06`/`BR-07` only require a non-empty name — so those warnings had nothing to report.

Live example (SIREN `300025764`):

```json
{ "siren": "300025764", "statut_diffusion": "P",
  "nom_complet": "[NON-DIFFUSIBLE]",
  "siege": { "code_postal": "[NON-DIFFUSIBLE]", "libelle_commune": "RENNES" } }
```

### The change

A masked value is no longer cross-checked. What stays public for those units keeps being checked: the SIREN itself, `etat_administratif`, and the commune (`libelle_commune` is deliberately still disclosed). Nothing changes for a regular unit.

This only affects the warnings displayed to the user. The generated XML is untouched.

### Tests

Run on the sandbox (Dolibarr 18.0.10 → 24.0.0), against the real API:

| | before (`main`) | after |
|---|---|---|
| non-diffusible customer, town = registry commune | 2 warnings | **0** |
| non-diffusible customer, town differs | 3 warnings | **1** (the town one — public data, legitimate) |
| regular customer, all fields match | 0 | 0 |
| regular customer, all fields differ | 3 warnings | 3 warnings |
| SIREN absent from the registry | 1 warning | 1 warning |

Also run: the real path (a real third party on SIREN `300025764`, a real validated invoice, `checkRequiredinformations()` then generation) — 3 warnings before, 1 after; the module test suite on 18 → 24 (135/136, the remaining failure is the pre-existing 17-only one); the QA generation matrix (112 generations over 8 instances); the full send/receive lifecycle on 7 versions both ways; and import of a generated CII as a supplier invoice on 18/23/24, read back from a separate process.

The CII payload of one and the same invoice, generated by `main` and by this branch on the same instance, is byte for byte identical (checked with a negative control to make sure the comparison would have caught a difference).


daGrumpf-bxp:
Note on CI: the Phan job is red, but with the four findings that are already red on `main` (`einvoicing.class.php:1973-1974` and `ajax/checksupplierinvoicestatus.php:202-203`, all `PhanPossiblyUndeclaredVariable`), which #705 fixes. None of them is in the code this PR touches, and this branch adds none.

PR #708, opened 2026-09-01
FIX: #695 An ampersand in a text value empties the CII element that carries it

### LINKED ISSUE #695: [EInvoicing] Empty TradingBusinessName makes CII invoice invalid
### Bug

The EInvoicing module generates an empty XML element in a customer CII invoice:

<ram:TradingBusinessName/>

The seller company name is correctly configured in Dolibarr and is correctly exported into:

<ram:Name>ETHIQUE & TACT Publicité</ram:Name>

However, the empty TradingBusinessName element causes the invoice to fail the SuperPDP validator with:

[PEPPOL-EN16931-R008] Document MUST not contain empty elements.

Removing only this empty element makes the CII invoice fully valid according to all three SuperPDP validation stages.

### Expected behavior

If no trading business name is available, the module should omit the XML element instead of generating:

<ram:TradingBusinessName/>

Alternatively, it should populate it only when a non-empty value exists.

### Module

einvoicing

### Dolibarr version

23.0.3

### Electronic invoice format

CII — EN16931

### Access Point

SuperPDP

### Steps to reproduce

1. Configure the seller company name in Dolibarr.
2. Generate a customer electronic invoice in CII format.
3. Open the generated *_cii.xml file.
4. The seller SpecifiedLegalOrganization contains an empty TradingBusinessName element.
5. Validate the XML using the SuperPDP validator.
6. Validation fails with PEPPOL-EN16931-R008.
7. Remove the empty element and validate again.
8. The XML becomes valid.

### Suggested correction

Generate TradingBusinessName only if its value is not empty.

### Module

einvoicing

### Dolibarr Version

23.0.3

### Environment PHP

_No response_

### Environment Database

_No response_

1. Configure the seller company name in Dolibarr.
2. Generate a customer electronic invoice in CII format.
3. Open the generated *_cii.xml file.
4. The seller SpecifiedLegalOrganization contains an empty TradingBusinessName element.
5. Validate the XML using the SuperPDP validator.
6. Validation fails with PEPPOL-EN16931-R008.
7. Remove the empty element and validate again.
8. The XML becomes valid.

### Suggested correction

Generate TradingBusinessName only if its value is not empty.

### Attached files

_No response_

daGrumpf-bxp:
Fixes #695.

## What actually happens

`DOMDocument::createElement()` **parses** its second argument. A value holding a bare ampersand — `ETHIQUE & TACT Publicité`, an ordinary company name — raises `unterminated entity reference` and the element is created **empty**: the business term is silently dropped, and the Access Point then refuses the document under `PEPPOL-EN16931-R008`, *"Document MUST not contain empty elements"*.

```php
$doc->createElement('ram:Name', htmlspecialchars('ETHIQUE & TACT'));  // <ram:Name>ETHIQUE &amp; TACT</ram:Name>
$doc->createElement('ram:TradingBusinessName', 'ETHIQUE & TACT');     // <ram:TradingBusinessName/>   <-- issue #695
```

That is exactly what the reporter sees: `ram:Name` (BT-27) is correct and `ram:TradingBusinessName` (BT-28) is empty, a few lines apart in `buildParty()`, because only the first one went through `htmlspecialchars()`. The seller trading name is never empty in the data — it is filled with the company name — so **omitting the element when it is empty, as the issue suggests, does not fix this case on its own**: the value is not empty, it is lost on the way in.

The ampersand is the only character that breaks `createElement()`; `<`, `>` and quotes are handled.

## What this changes

1. **Every value that comes from the invoice or the company data is escaped** on its way into the document, in `CIIProtocol` and in `CdarHandler` (the lifecycle documents, where a party name or a reason would be lost the same way). Only literals of the file and the output of `number_format()` / `DateTime::format()` are written as they are, and the rule is written down in the `buildXML()` docblock so it stays true.
2. **Three optional terms are no longer written empty**: the trading name (BT-28 / BT-45), the payment means text (BT-82) and the account name (BT-85). EN 16931 asks for BT-28 only when it differs from the party name, so nothing to say means no element rather than an empty one.
3. **New test file** `test/phpunit/CIITextEscapingTest.php`: an invoice whose every text term holds an ampersand, asserting each term round-trips, that no profile produces an empty element, and that the optional terms are omitted when they have no value. It fails on the code before this patch.

## Measured, not deduced

A real invoice generated on Dolibarr 23, whose seller name, buyer name, bank account holder, line description and discount reason all hold an ampersand, sent to the public SuperPDP validator (`api.superpdp.tech`), same invoice before and after:

| document | empty elements | verdict |
|---|---|---|
| before the patch | 4 — BT-28, BT-45, BT-85, BT-104 | **INVALID**, 4 × `PEPPOL-EN16931-R008` |
| before the patch, same invoice without any ampersand | 0 | VALID |
| after the patch, ampersands everywhere | 0 | **VALID**, all five values intact |
| after the patch, BT-112 altered on purpose (control) | 0 | INVALID, `BR-CO-15` / `BR-CO-16` — the validator does bite |

Bench with Dolibarr 17, 18, 19, 20, 21, 22, 23 and 24 sharing the same module tree:

- **PHPUnit** file by file on the 8 cores, 17 files each: no failure (the one failure on 17 is `EInvoicingSamplesTest` on an instance where the module is disabled, and predates this branch).
- **Export**: sample generation, CII and Factur-X × standard / deposit / credit note on 18, 20, 23 and 24 — 24 documents, all OK.
- **Import**: the ampersand document read back with `createSupplierInvoiceFromSource()` on 23, then checked **from a second process** (nothing left in an open transaction); the discount reason arrives as `Geste commercial R & Cie D`, ampersand included.
- **End to end** through the Access Point sandbox, on a new invoice carrying an ampersand in its line: emitted, routed, received on the other instance, statuses 200 and 201 back, then a 205 emitted from the receiving side — the CDAR built by the patched `CdarHandler` is accepted and holds no empty element.
- **Limit cases**: a value that is only `&`, a value already holding `&amp;`, a value holding markup — no empty element, text preserved. A value that is not valid UTF-8 cannot be tested from the data: MariaDB refuses it in the column.

## Note for later, not in this PR

BT-28 is currently filled with the company name, so it repeats BT-27. EN 16931 defines it as *"a name by which the Seller is known, **other than** Seller name"*, to be used *"if different"*. The buyer side has a natural source in Dolibarr (`name_alias`), the seller side has none (`setMysoc()` only reads `MAIN_INFO_SOCIETE_NOM`), so the honest mapping would be `name_alias` for the buyer and nothing for the seller. That is a behaviour change, so it is left out of a bug fix.


daGrumpf-bxp:
The `Run phan` check is red on `main` itself, not because of this branch: the artifact of this PR's run (33543970235) and the artifact of the last push on `main` (33540908509) hold **the same four findings**, in two files this PR does not touch — `einvoicing/ajax/checksupplierinvoicestatus.php:202-203` (`PhanPossiblyUndeclaredGlobalVariable`) and `einvoicing/class/einvoicing.class.php:1946-1947` (`PhanPossiblyUndeclaredVariable`). `build`, `Factur-X container is PDF/A-3` and `assign-and-label` pass. Happy to send those four in a separate PR if that helps.

daGrumpf-bxp:
## Follow-up validation: 19 poisoned strings, one invoice each

The fix escapes every value written into the document, so it was worth re-testing with a much wider
set of characters than the ampersand of the report. Each test string was carried by **one invoice of
its own**, holding it in **14 business terms at once** (BT-13, 22, 27, 35, 37, 41, 44, 50, 52, 85,
104, 153, 154, 155), and every string is something a user can actually type or paste into Dolibarr:

`&` alone · `A & B & C` · `Caf&eacute; &amp; Co` · `<b>Fût</b> 30L` · `L'Été "Blonde" & Cie <bio>` ·
`L’Été « Blonde » — Œuf & Cie… 12 € ±3° n°7` · non-breaking spaces · a multi-line description with a
tab · `Fin ]]> & <![CDATA[` · `<?xml version="1.0"?><ram:Name>pwn</ram:Name>` ·
`a<b AND c>d & e -- ' OR 1=1; DROP` · CJK · Arabic/Hebrew · emoji · control characters pasted from a
PDF · leading/trailing spaces · a 450-character value · `50% {a} $var C:\dossier`.

| what was run | result |
|---|---|
| CII generation on Dolibarr 18, 20, 23 and 24 | 18/19 strings: well formed, **0 empty element**, all 14 terms round-trip unchanged |
| Factur-X (XML extracted from the PDF) | same as CII |
| profiles MINIMUM / BASICWL / BASIC / EN16931 / EXTENDED | no empty element, except the structural `ram:ApplicableHeaderTradeDelivery` of MINIMUM |
| **SuperPDP validator**, the 19 documents + 5 from the other cores | **VALID**, 207–210 checks each |
| control, on purpose invalid (BT-112 altered) | INVALID `BR-CO-15` / `BR-CO-16` — the validator does bite |
| import back with `createSupplierInvoiceFromSource()`, read from a second process | 15/16: allowance reason, line description and document number returned character for character |
| PHPUnit, 4 cores × 17 files | 68/68 |
| CDAR (`CdarHandler`), 15 strings | well formed, no empty element, the document reference escaped (`&amp;`, `]]&gt;`, `&lt;![CDATA[`) |
| end to end through the Access Point sandbox | ack `Ok`, invoice received on the other instance, statuses 200 and 201 back; `&`, `—`, `«»`, `…`, `€` arrive intact |

The same 19 invoices generated on the tree **before** this fix, on the same instance, for comparison:

| | before | after |
|---|---|---|
| 16 strings out of 19 | 5 empty elements, 3 terms lost, 5 × `unterminated entity reference` | 0 empty, nothing lost |
| `Caf&eacute;` and `a<b AND c>d` | **not even well-formed XML** (`Entity 'eacute' not defined`) | well formed |
| SuperPDP | INVALID, `PEPPOL-EN16931-R008`, or refused before parsing | VALID |

Worth noting when reading a SuperPDP report: `PEPPOL-EN16931-R008` is a **warning** filed in
`subreports[].messages[]`, not in `failures[]` — a reader that only looks at `failures[]` shows
"INVALID | -" without naming the rule.

## One gap this did not close

A **control character** (`0x0B`, `0x0C` — what a copy/paste from a PDF brings along) goes through
untouched: MariaDB accepts it in the column, `htmlspecialchars()` does not escape it (it only handles
`& < > " '`), `DOMDocument` writes it without a warning, and the produced document **is not valid XML
1.0**. Measured at three stages, identically on 18, 20, 23 and 24:

- `DOMDocument::loadXML()` → `PCDATA invalid Char value 11`;
- SuperPDP → `is_valid=false`, no subreport, `error: "line 19: illegal character code U+000B"`;
- the Access Point, on a real send → `HTTP 400 - CREATE_ERROR - XML syntax error on line 19: illegal character code U+000B`.

The CDAR breaks the same way when the value is the reference of the received document. This is **not
a regression of this fix** — the tree before it behaves identically — it is simply a character class
the escaping does not cover. Filtering the characters XML 1.0 forbids
(`\x00-\x08\x0B\x0C\x0E-\x1F`) where the escaping now happens would close it. Before opening
anything, it still has to be shown that such a byte can be entered **through the UI** (`GETPOST` may
well clean it): the bench writes the value through the model, not through a POST.

## Two side observations, both pre-existing

- `dol_string_nohtmltag()` on the line description (`buildinvoicelines.inc.php`) removes anything
  that looks like a tag, so `a<b AND c>d & e` reaches BT-154 as `ad & e`. Intended for the rich text
  editor of Dolibarr, but it silently eats plain text that merely contains `<`. BT-41 has the same
  behaviour, from the core itself (`getFullName()` ends with `dol_string_nohtmltag()`).
- Four-byte UTF-8 (emoji) cannot be stored at all when the database is utf8mb3; put through the
  builder in memory, it comes out intact.

PR #709, opened 2026-09-01
einvoicing: aggregate VAT breakdown by category/rate/exemption (fix duplicate ApplicableTradeTax rejected by PDP)

Pichinov-Jose:
## Problem

When an invoice mixes lines that carry a French VAT source code (e.g. `TVAFR20`) with lines that have **no** VAT code, at the **same** VAT rate and category, `buildinvoicelines.inc.php` builds **two** VAT breakdown groups, because the breakdown array is keyed by `tva_tx . vat_src_code`:

- lines at 20% with code `TVAFR20` → key `20 (TVAFR20)`
- lines at 20% without a code → key `20`

Both are then emitted as `<ram:ApplicableTradeTax>` groups with an identical `CategoryCode` (`S`) and `RateApplicablePercent` (`20.00`) but different `BasisAmount`. EN 16931 requires **one** VAT breakdown line (BG-23) per unique combination of category (BT-118), rate (BT-119) and exemption reason (BT-120/BT-121), so a conformant PDP / Access Point rejects the invoice for a semantic error — e.g. Factur-X EXTENDED rule **BR-FXEXT-S08b** (and the equivalent EN 16931 **BR-S-08**): *"duplicate VAT breakdown, the category taxable amount does not reconcile with the sum of the line net amounts"*.

Concrete case: an invoice whose product lines carry `TVAFR20` and whose service lines carry no code, all standard-rated at 20%, is emitted with two 20% `ApplicableTradeTax` groups and rejected `REJ_SEMAN`, even though every amount is correct.

## Fix

Key the VAT breakdown on the dimensions EN 16931 actually groups on — **VAT category code + VAT rate + exemption reason** — instead of the Dolibarr `vat_src_code`, which is a national dictionary code that is not represented in the emitted breakdown. Lines that share (category, rate, exemption) are aggregated into a single `ApplicableTradeTax`, as required by the standard.

```php
$vatBreakdownKey = $categoryVAT.'|'.$line->tva_tx.'|'.$exemptionReasonCode.'|'.$exemptionReason;
```

- No database / schema change.
- `vat_src_code` is still stored on the breakdown entry (it only feeds the debug XML comment); the aggregated entry keeps one representative value.
- Exempt breakdowns keep one group per exemption reason (the reason is part of the key), so exemptions are unaffected.
- Different categories or rates stay in separate groups (e.g. `S` 20% vs `AE` 20% reverse charge).

## Test

Verified on a real invoice that previously emitted two standard-rated 20% `ApplicableTradeTax` groups (BasisAmount 59.00 + 114.00) and was rejected `REJ_SEMAN` / BR-FXEXT-S08b. After the fix a single `ApplicableTradeTax` (BasisAmount 173.00) is emitted and the Access Point accepts the invoice, returning the full positive lifecycle (Deposited → Issued → Received → Made available).


eldy:
I see you made your PR in draft mode ?
Is it done on purpose ? If you think your PR is ready you can push it in ok mode.

PR #710, opened 2026-09-01
einvoicing: allow re-sending a rejected e-invoice after correction

Pichinov-Jose:
## Problem

Once an e-invoice has really been transmitted to the Access Point (it has a persistent `flow_id`), the invoice card locks regenerate / re-send. That lock is a sensible default and can be opted out of with `EINVOICING_ALLOW_RESEND_TRANSMITTED` (re-send) and `EINVOICING_ALLOW_REGEN_TRANSMITTED` (regenerate).

But even with the opt-out set, an invoice whose last lifecycle status is **Rejected (213, "Rejected by seller AP")** still cannot be re-sent: `STATUS_REJECTED` is absent both from the status list that shows the "Send to PDP" button and from the `send_to_pdp` action handler. In practice, after correcting the cause of a **format / semantic** rejection — where the invoice content is unchanged and only the generated XML needed fixing — the user must manually switch the e-invoice status back to "Generated" before the button both appears *and* actually transmits (the action silently no-ops otherwise). That extra manual step is confusing.

## Fix

Treat `STATUS_REJECTED` as a resendable state, consistently in the three places that gate re-send on the customer invoice card:

1. the **"Send to PDP" button** visibility list;
2. the `$resend` flag, so the button label reads "Re-send" rather than "Send";
3. the **`send_to_pdp` action handler** gate, so clicking actually transmits.

This remains governed by the existing transmitted-lock opt-out (`EINVOICING_ALLOW_RESEND_TRANSMITTED`): with the default lock in place the button stays hidden, so the safe default is unchanged. It only enables the natural *reject → correct → re-send the same invoice* flow for a rejection that is a formatting / semantic issue (the invoice is commercially correct) rather than one that would call for a credit note.


eldy:
I see you made your PR in draft mode ?
Is it done on purpose ? If you think your PR is ready you can push it in ok mode.

einvoicing/class/actions_einvoicing.class.php
```diff
@@ -408,10 +408,11 @@ public function addMoreActionsButtons($parameters, $object, &$action, $hookmanag
 					$einvoicing::STATUS_ERROR,
 					$einvoicing::STATUS_UNKNOWN,
 					$einvoicing::STATUS_AWAITING_VALIDATION,		// retry affordance (PA will refuse a duplicate)
-					$einvoicing::STATUS_AWAITING_ACK				// retry affordance (PA will refuse a duplicate)
+					$einvoicing::STATUS_AWAITING_ACK,				// retry affordance (PA will refuse a duplicate)
+					$einvoicing::STATUS_REJECTED					// resend after correcting a rejected e-invoice (gated by EINVOICING_ALLOW_RESEND_TRANSMITTED)
```
eldy, line 392: If this case is protected by the use of option EINVOICING_ALLOW_RESEND_TRANSMITTED, we should find a 
if EINVOICING_ALLOW_RESEND_TRANSMITTED somewhere in this part of code that defin which button to show ?

Pichinov-Jose: Thanks! The opt-out is the option `EINVOICING_ALLOW_RESEND_TRANSMITTED`, but it's read in a single place rather than inline here. A few lines above, `$locked = $einvoicing->isTransmittedLockActive($object->id, $object->ref)`, and this button block is gated by `!$locked`. `isTransmittedLockActive()` returns `false` as soon as `EINVOICING_ALLOW_RESEND_TRANSMITTED` is set (`einvoicing.class.php`, `isTransmittedLockActive()`), so the (re)send button stays hidden for a really-transmitted invoice unless the option opts out. The same helper backs the server-side `send_to_pdp` gate (the real enforcement / defense in depth), so keeping the option in one place stops the button visibility and the enforcement from drifting apart. I've pushed a comment clarification at the button block to make that explicit — happy to read the option inline here instead if you'd prefer.

eldy:
It seems another PR has adressed this trouble.
Can you check ?
If not, fix the conflict.

Pichinov-Jose:
Hi @eldy, thanks for checking.

I looked into it: no other merged PR covers this case. Current `main` still has no re-send path for a **rejected** e-invoice — `STATUS_REJECTED` is absent from both the single-invoice button gate and the mass-action gate in `actions_einvoicing.class.php`. So this PR is still needed.

I've also resolved the conflict: I merged the latest `main` and kept its `!einvoicingIsSendDisabled()` guard on the button, while preserving the `STATUS_REJECTED` re-send affordance (gated by `EINVOICING_ALLOW_RESEND_TRANSMITTED`). The branch is now mergeable and CI is green.

PR #712, opened 2026-09-01
einvoicing: store debug payloads UTF-8-safe (fix DB error 1366 on binary PDP responses)

Pichinov-Jose:
## Problem

When `EINVOICING_DEBUG_MODE` is enabled, the raw API request/response is persisted into utf8mb4 `TEXT` columns:

- `llx_einvoicing_call.request_body` / `.response` — `AbstractPDPProvider::logCall()`
- `llx_einvoicing_document.response_for_debug` — `SuperPDPProvider`, `EsalinkPDPProvider`

Some PDP responses contain **non-UTF-8 bytes** (e.g. a signed or compressed lifecycle payload). Stored verbatim they raise **`DB_ERROR_1366` (Incorrect string value)** on insert. Since that write happens while the flow is being processed, the whole synchronization of the flow aborts and the incoming flow stays stuck — its lifecycle status is never applied. Seen in production on incoming `CustomerInvoiceLC` lifecycle flows with debug mode on: the invoice card kept showing an old status because every re-sync failed on 1366.

## Fix

Add one helper on the shared base provider that makes any payload safe for a utf8mb4 TEXT column, and route the three debug writes through it:

```php
protected function makeStorableDebugPayload($payload)
{
    if (!is_string($payload) || $payload === '') {
        return (string) $payload;
    }
    if (preg_match('//u', $payload)) {   // already valid UTF-8
        return $payload;
    }
    return '[base64] ' . base64_encode($payload);
}
```

- Valid UTF-8 (the normal JSON/XML case) is stored unchanged.
- Binary / non-UTF-8 payloads are base64-encoded behind a `[base64] ` marker, so the debug trace is still captured and stays reversible, and the insert never fails.
- No schema change; debug mode can safely remain enabled.


eldy:
Good idea. Why not.

eldy (commented):
Good idea. Why not.

PR #716, opened 2026-09-02
FIX #680 [einvoicing] The join on einvoicing_routing stops repeating the thirdparties of the list

### LINKED ISSUE #680: [einvoicing] Invoice list: unindexed LEFT JOIN on llx_einvoicing_extlinks adds seconds to every load and duplicates rows

daGrumpf-bxp:
Part of #680. Split out of #697 at @eldy's request. #713, #714 and #715 are merged; this one and #717 are independent of each other and of what is already in `main`.

Updated after @eldy's review: the correlated subquery is gone, the join now carries the three criteria he asked for.

## The bug

A thirdparty can hold several routing identifiers, and a routing row for its default product on top of them. `printFieldListFrom()` joined `llx_einvoicing_routing` on `fk_soc` alone, so the thirdparty list of the core repeated the thirdparty once per row of the table, and its record count with it.

## The fix

```php
LEFT JOIN llx_einvoicing_routing rt ON rt.fk_soc = s.rowid
  AND rt.routing_type = 'thirdparty' AND rt.active = 1 AND rt.is_default = 1
```

`is_default = 1` designates exactly one row: `addRouting()` sets it on the first active routing of the type, and `setDefaultRouting()` / `setRoutingAsDefault()` both clear the flag on the other rows of the same type before setting it. The row it designates is the one the card of the thirdparty shows at the top of its list, and the one `fetchDefaultRouting()` answers. It stays a `LEFT JOIN`: a thirdparty holding no routing identifier keeps an empty column instead of leaving the list.

## `deleteRouting()` had to filter the type as well

There was one reachable state where `is_default = 1` designated no row at all. When the default was deleted from the card, `deleteRouting()` promoted the next entry with `WHERE fk_soc = X AND active = 1 ORDER BY rowid ASC LIMIT 1`, **without filtering `routing_type`**. A `product` routing with a lower rowid took the default, and the surviving `thirdparty` routing kept `is_default = 0` — so `fetchDefaultRouting()` answered nothing for a thirdparty that still held an active identifier and still emitted with it.

Reproduced on the seven instances, one thirdparty holding a `product` routing then two active `thirdparty` routings, default deleted from the card:

| | before the delete | after the delete, `main` | after the delete, this PR |
|---|---|---|---|
| `is_default` of the surviving `thirdparty` routing | 0 | **0** | **1** |
| `fetchDefaultRouting()` | `0225:111111111` | **nothing** | `0225:222222222` |
| column of the list | `0225:111111111` | **empty** | `0225:222222222` |

## Measured on the bench, before → after, without touching the data

Seven instances, Dolibarr 18.0.10 to 24.0.0. Thirdparty list of the core, `?limit=100`, over HTTP as a real user.

| instance | rows / distinct thirdparties, `main` | this PR |
|---|---|---|
| Dolibarr 23.0.4 | 33 / 32 | **32 / 32** |
| Dolibarr 24.0.0 | 15 / 14 | **14 / 14** |

The five other instances hold one routing per thirdparty and were already 1:1; they stay 1:1. The four lists the module completes answer `200` with no SQL error on all seven, and the search on the routing identifier of #714 still finds a thirdparty by **every** one of its active identifiers, not only by the default one the column now shows (negative control `ZZZ-INEXISTANT` = 0 thirdparty). PHPUnit 21/21 on Dolibarr 18 to 24.

### DIFF
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -1360,6 +1360,13 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 			if (in_array('thirdpartylist', $contexts, true)) {
 				$this->resprints .= ' LEFT JOIN ' . $db->prefix() . "einvoicing_extlinks as ext ON ext.element_id = s.rowid AND ext.element_type = 'societe'";
 				$this->resprints .= ' LEFT JOIN ' . $db->prefix() . "einvoicing_routing rt ON rt.fk_soc = s.rowid";
+				// A thirdparty can hold several routing identifiers, and a routing row for its default product
+				// on top of them, so joining on fk_soc alone repeats the thirdparty in the list as many times.
+				// The active default routing of type 'thirdparty' is the single row the list must show: it is
+				// the one the card of the thirdparty displays at the top of its list, and the one
+				// EInvoicing::fetchDefaultRouting() answers. The column stays empty for a thirdparty holding
+				// no routing identifier, so no row leaves the list.
+				$this->resprints .= " AND rt.routing_type = 'thirdparty' AND rt.active = 1 AND rt.is_default = 1";
 			}
 
 			if (in_array('invoicelist', explode(':', $parameters['context']))) {
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -3012,8 +3012,8 @@ class EInvoicing
 
 		$db->begin();
 
-		// Check if this entry was the default
-		$sql = "SELECT is_default FROM " . $db->prefix() . "einvoicing_routing";
+		// Check if this entry was the default, and for which type of routing
+		$sql = "SELECT is_default, routing_type FROM " . $db->prefix() . "einvoicing_routing";
 		$sql .= " WHERE rowid = " . (int) $rowid;
 		$resql = $db->query($sql);
 		if (!$resql) {
@@ -3023,6 +3023,7 @@ class EInvoicing
 		}
 		$obj = $db->fetch_object($resql);
 		$wasDefault = $obj ? (int) $obj->is_default : 0;
+		$routingtype = $obj ? $obj->routing_type : '';
 		$db->free($resql);
 
 		$sql = "DELETE FROM " . $db->prefix() . "einvoicing_routing";
@@ -3033,11 +3034,15 @@ class EInvoicing
 			return -1;
 		}
 
-		// Reassign default to oldest remaining entry if needed
+		// Reassign default to oldest remaining entry of the same type if needed. Without the filter on
+		// routing_type, a routing of the other type could take the default and leave the remaining entries
+		// of the deleted type without any, so fetchDefaultRouting() would answer nothing for a thirdparty
+		// that still holds an active routing identifier.
 		if ($wasDefault) {
 			$sql = "UPDATE " . $db->prefix() . "einvoicing_routing";
 			$sql .= " SET is_default = 1";
 			$sql .= " WHERE fk_soc = " . (int) $fk_soc . " AND active = 1";
+			$sql .= " AND routing_type = '" . $db->escape($routingtype) . "'";
 			$sql .= " ORDER BY rowid ASC LIMIT 1";
 			$db->query($sql); // Non-blocking: if no rows remain, nothing to reassign
 		}


einvoicing/class/actions_einvoicing.class.php
```diff
@@ -1360,6 +1360,14 @@ public function printFieldListFrom($parameters, $object, &$action, $hookmanager)
 			if (in_array('thirdpartylist', $contexts, true)) {
 				$this->resprints .= ' LEFT JOIN ' . $db->prefix() . "einvoicing_extlinks as ext ON ext.element_id = s.rowid AND ext.element_type = 'societe'";
 				$this->resprints .= ' LEFT JOIN ' . $db->prefix() . "einvoicing_routing rt ON rt.fk_soc = s.rowid";
```
eldy, line 1362: Why not just adding the condition 
`AND sub.routing_type = 'thirdparty' AND sub.active = 1`
on the line 
`$this->resprints .= ' LEFT JOIN ' . $db->prefix() . "einvoicing_routing rt ON rt.fk_soc = s.rowid";`

This is the proper way to avoid duplication when you make a left join and you need more criteria to guarantee that you select only 1 line

daGrumpf-bxp: That was my first attempt, and it turns out not to be enough here: the duplication does not come from the rows of the other type or from the inactive ones.

A thirdparty legitimately holds **several active routing identifiers of type `thirdparty`** at the same time — that is what the "add a routing identifier" action of the tab writes (`pdp_addrouting` → `addRouting($socId, $newRoutingId, $newRoutingInfo, 'thirdparty')`, `actions_einvoicing.class.php:797`), and `fetchAllRoutings()` is written to return that whole list, ordered `is_default DESC, rowid ASC`. There is no unique key on `(fk_soc, routing_type)` and there should not be one. So `AND rt.routing_type = 'thirdparty' AND rt.active = 1` narrows the join, but for such a thirdparty it still matches N rows.

Measured on the bench, seven instances, same `FROM` as `printFieldListFrom()`, no data touched:

| instance | no criteria (before) | `type + active` only | subquery (this PR) |
|---|---|---|---|
| 18.0.10 | 15 rows / 15 thirdparties | 15 / 15 | 15 / 15 |
| 19.0.4 | 4 / 4 | 4 / 4 | 4 / 4 |
| 20.0.4 | 16 / 16 | 16 / 16 | 16 / 16 |
| 21.0.4 | 5 / 5 | 5 / 5 | 5 / 5 |
| 22.0.5 | 11 / 11 | 11 / 11 | 11 / 11 |
| **23.0.4** | **32 / 31** | **32 / 31** | **31 / 31** |
| **24.0.0** | **14 / 13** | **14 / 13** | **13 / 13** |

The two instances that duplicate are exactly the two where one thirdparty holds 2 active `thirdparty` routings; on the five others every thirdparty holds one and the three variants agree.

So your criteria is necessary but not sufficient — something still has to designate **which** of the several identifiers the list shows. The subquery does that with the rule already used everywhere else in the module: `is_default` first, then lowest `rowid`, which is the row the card of the thirdparty shows at the top of its list and the one `fetchDefaultRouting()` answers when a default exists. It is covered by `idx_einvoicing_routing_soc (fk_soc, routing_type, active)`, already in `main`.

If you would rather not have a correlated subquery there, the two other shapes I see are:

- join on `AND rt.routing_type = 'thirdparty' AND rt.active = 1 AND rt.is_default = 1` — one row guaranteed, but the column then stays empty for every thirdparty that has no default, and `addRouting()` does not set one, so that is most of them;
- add `GROUP BY s.rowid` to the query — which is not mine to add from a hook, the `SELECT`/`FROM` belongs to the core page.

Happy to switch to either, or to keep this one. Tell me which you prefer and I push it.

eldy: Yes, the solution 
`join on AND rt.routing_type = 'thirdparty' AND rt.active = 1 AND rt.is_default = 1 — one row guaranteed, but the column then stays empty for every thirdparty that has no default, and addRouting() does not set one, so that is most of them;`
will return an empty value for field if there is no specific routing set, what it seems it is what we want ?
Because if waht we want is a line returned only if there is a specific routingdefined, we don't need a left join, we can do a inner join.

daGrumpf-bxp: You are right on the first half, and I was wrong: `addRouting()` **does** set the default. It counts
the active routings of that type first and inserts `is_default = 1` when the thirdparty has none
(`einvoicing.class.php:2974-2984`), so `is_default = 1` normally designates exactly one row.
Measured over the 8 instances of the bench: every thirdparty holding an active `thirdparty` routing
has exactly one default row — none has zero, none has two.

There is one reachable state where it designates none, and it is a bug of ours rather than of your
criteria. When the default is deleted from the card, `deleteRouting()` promotes the next one with
`WHERE fk_soc = X AND active = 1 ORDER BY rowid ASC LIMIT 1` (`einvoicing.class.php:3037-3042`) —
**without filtering `routing_type`**. So if the thirdparty also holds a `product` routing with a
lower rowid, that row takes the default and the surviving `thirdparty` routing keeps
`is_default = 0`.

Reproduced identically on 18.0.10, 23.0.4 and 24.0.0 — one thirdparty holding a `product` routing,
then two active `thirdparty` routings, then the default deleted from the card:

| join criteria | before the delete | after the delete |
|---|---|---|
| none (`main` today) | 3 rows | 2 rows |
| `type + active` | 2 rows | 1 row |
| `type + active + is_default` | 1 row, **column filled** | 1 row, **column empty** |
| subquery (this PR) | 1 row, column filled | 1 row, column filled |

In that state `fetchDefaultRouting()` answers nothing, while the card of the thirdparty and
`fetchAllRoutings()` both still show `0225:222222222` — which is the address the module actually
emits with. So the list would say "no routing" about a thirdparty that has one and sends with it.

On the inner join: I cannot use one from here. `printFieldListFrom` only appends to the `FROM` of the
core's own page, so an `INNER JOIN` would drop every thirdparty **without** a routing out of the
thirdparty list itself — it would remove rows from the list, not empty a column.

So your three criteria is the shape I would rather ship too, and it becomes correct as soon as that
promotion filters `routing_type` — one line in `deleteRouting()`. Your call: **(a)** I add the
`routing_type` filter and switch the join to `type + active + is_default`, or **(b)** I keep the
subquery, which is correct without touching the delete path. I lean towards (a): it fixes a real
defect on the way and leaves the simpler SQL in the hook.

eldy: We must switch on a)
the join to type + active + is_default
no reason to do subquery.

daGrumpf-bxp: Done, pushed as `13e92c2a`. The join is now

```php
LEFT JOIN llx_einvoicing_routing rt ON rt.fk_soc = s.rowid
  AND rt.routing_type = 'thirdparty' AND rt.active = 1 AND rt.is_default = 1
```

and `deleteRouting()` promotes the next entry of the **same type** (`AND routing_type = '...'` added to the `UPDATE`), which is what makes `is_default = 1` designate exactly one row in every reachable state.

Measured on the seven instances, thirdparty list of the core over HTTP as a real user, `?limit=100`:

| instance | `main` | this PR |
|---|---|---|
| 18.0.10 / 19.0.4 / 20.0.4 / 21.0.4 / 22.0.5 | already 1:1 | unchanged |
| 23.0.4 | 33 rows / 32 thirdparties | **32 / 32** |
| 24.0.0 | 15 / 14 | **14 / 14** |

And the delete path, one thirdparty holding a `product` routing then two active `thirdparty` routings, default deleted from the card:

| | before | after, `main` | after, this PR |
|---|---|---|---|
| `is_default` of the surviving `thirdparty` routing | 0 | 0 | **1** |
| `fetchDefaultRouting()` | `0225:111111111` | nothing | `0225:222222222` |
| column of the list | `0225:111111111` | empty | `0225:222222222` |

Also checked, since #714 and this one meet on the same list: the search on the routing identifier still finds a thirdparty by **every** one of its active identifiers, not only by the default one the column now shows — `315143296_2341` and `89230418900020_electricite` both return their thirdparty on 23 and 24, negative control `ZZZ-INEXISTANT` returns none. The four lists answer `200` with no SQL error on the seven instances, PHPUnit 21/21 on each.

PR #723, opened 2026-09-02
FIX [einvoicing] The four discount sentinels of the core no longer reach the customer

P-tite-tete:
## What this changes

A discount built from another piece — a credit note applied, a deposit deducted, an excess payment
carried over — carries no wording of its own. The core writes one of four sentinels in the description
of the discount, `insert_discount()` copies it into the description of the line, and
`pdf_getlinedesc()` resolves it against the piece it comes from when it prints.

Nothing resolved it for the e-invoice. The customer received the raw marker, in two places:

- **BT-97**, the reason of a document level allowance: `<ram:Reason>(CREDIT_NOTE)</ram:Reason>`
- **BT-153**, the item name of a deducted deposit line, which stays a line in the document (method 1 of
  XP Z12-014): `<ram:Name>(DEPOSIT)</ram:Name>`

They now read the same text as the PDF, naming the piece the amount comes from, and beside it the
invoice that piece corrects — a deduction that only says which credit note it comes from leaves the
customer to find which invoice that credit note was about.

## Why not a search through the text

The test is the one the core makes: an exact description on a line that really carries a discount
(`fk_remise_except`). Matching the text loosely is wrong in both directions — a line of work named
`Reprise (DEPOSIT) du chantier` would be renamed, and a description edited by hand would be missed.

And the four sentinels are not spelled alike: `(CREDIT_NOTE)` holds an underscore where `(EXCESS PAID)`
and `(EXCESS RECEIVED)` hold a space. A single pattern over `[A-Z_]+` misses two of the four.

The reference the text quotes follows the side the discount belongs to
(`ref_invoice_supplier_source` when `discount_type` is set, `ref_facture_source` otherwise); reading the
customer side unconditionally names nothing at all on a supplier discount.

## A line named `(DEPOSIT)` that deducts nothing

`main` treats any line whose description is exactly `(DEPOSIT)` as a deducted deposit without ever
looking at `fk_remise_except`, and flips its sign. A line of work an operator happened to name that way
took a 300 € invoice out at `-300.00` — basis, grand total and amount due — so the document declared
that the seller owed the customer the price of the job. This is the most expensive defect of the lot and
it was missing from this description until the review pointed it out.

The test here is the one the core makes, both halves of it: the description **and** the discount the
line points at. Such a line now keeps its name, its sign and its amount.

The second half lives in `einvoicingDiscountLabelOfLine()` rather than inline in the condition, because
the two call sites do not read a line the same way. The one writing BT-97 already stands inside a test
on `fk_remise_except`, and it must still resolve when the discount behind it cannot be fetched — that
is what the `...NoSource` wording exists for. Reached from a line carrying no discount at all, that same
wording would state that a line deducting nothing is a deposit deducted, which is a different situation
and a false one.

## A line with no name

While looking at what reaches BT-153: a line holding neither a product label nor a description is
issued with an empty name, and refused by the platform on **BR-25**, on a line number the seller then
has to go and find.

`validateInvoiceConfiguration()` now lists every such line before validation, and the generator refuses
to build the document, naming them all at once rather than one refusal per round trip. Title and
subtotal pseudo-lines are not concerned, and neither are discount lines, whose name is built from the
piece they deduct.

The shape that gets there is not an empty description: the free-line branch of the card refuses that on
every core from 18 to 24 (`empty($product_desc)`, renamed `$line_desc` in 22). It is a description made
of markup only — `<p><br></p>`, what the WYSIWYG editor leaves behind when the text is deleted. The POST
is not empty, so the card accepts the line, and `dol_string_nohtmltag()` then strips it to nothing.

## A last look before the document leaves

One shared list, `einvoicingDiscountSentinels()`, feeds both the resolution and a final check that
reports a sentinel having reached BT-153, BT-154 or BT-97. Two lists would let one spelling drift on one
side only, which is the shape of the original defect.

That check is an equality, never an inclusion, and it reports in the log rather than refusing: an
unresolved marker is ugly, not invalid, and holding an invoice back over it would cost the seller more
than it saves. It exists because the leak was found in two places, the second after the first was
fixed — an analysis can miss a construction path.

## Tests

`test/phpunit/DiscountSentinelTest.php`, registered in `AllTests.php`: the four sentinels and the piece
each names, the supplier side reference, the false positives that must go out untouched, the case with
no readable source piece, a line named after a sentinel while carrying no discount (each of the four,
with and without a discount behind the line), the corrected invoice named beside the piece, the deposit
date option, and the exact spelling of the four sentinels pinned against a future drift.

Verified end to end on an instance running Dolibarr 23.0.3: on the invoice that carried four
`(CREDIT_NOTE)` allowances, the regenerated CII differs from the previous one by those four reasons and
nothing else — no total moved.

## Housekeeping

Rebased on `main` (c06deafe).

The amounts half of this branch is gone. `main` now reads BT-113 from `getSommePaiement()`,
`getSumDepositsUsed()` and `getSumCreditNotesUsed()`, and joins the BT-25/BT-26 references on the type
of the source invoice — which is what the review suggested, and a better job than the three-value `IN`
over descriptions this branch carried: it holds by construction and hands over `f.type` for the
document type codes rather than deriving one back from a description string. Nothing of it is left here.

`AllTests.php` is out as well, `main` having replaced the hand written list with a `glob()`. The
`ChangeLog.md` entry is gone from the whole series and not only from its final state, the file being
regenerated from commit titles, and `langs/fr_FR/einvoicing.lang` is out too — translations other than
`en_US` come back through the translation platform.

What remains is the resolution of the four sentinels, the line named after one while deducting nothing,
the BR-25 refusal and the last look.

### DIFF
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -1491,7 +1491,9 @@ class EInvoicing
 				$hasLabel = trim((string) ($line->product_label ?? '')) !== '';
 				$hasDesc = trim(dol_string_nohtmltag((string) ($line->desc ?? ''), 0)) !== '';
 				if (!$hasLabel && !$hasDesc) {
-					$linesWithNoName[] = (int) ($line->rang ?: count($linesWithNoName) + 1);
+					// The rank places the line on the paper, the rowid is what a correction is addressed to.
+					// Naming both is what lets whoever reads this go straight to the line and fix it.
+					$linesWithNoName[] = ($line->rang ? '#'.((int) $line->rang) : '').' (id '.((int) $line->id).')';
 				}
 			}
 		}
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -1493,7 +1493,11 @@ class EInvoicing
 				if (!$hasLabel && !$hasDesc) {
 					// The rank places the line on the paper, the rowid is what a correction is addressed to.
 					// Naming both is what lets whoever reads this go straight to the line and fix it.
-					$linesWithNoName[] = ($line->rang ? '#'.((int) $line->rang) : '').' (id '.((int) $line->id).')';
+					// FactureLigne and FactureFournisseurLigne both hold the rank, their common parent
+					// does not declare it, and this reads whichever of the two the invoice carries.
+					// @phan-suppress-next-line PhanUndeclaredProperty
+					$rank = (int) ($line->rang ?? 0);
+					$linesWithNoName[] = ($rank ? '#'.$rank : '').' (id '.((int) $line->id).')';
 				}
 			}
 		}
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -1482,6 +1482,11 @@ class EInvoicing
 		// Title and subtotal lines are not concerned: they are pseudo-lines that never reach the
 		// document. A discount line is not concerned either, its name being built from the piece it
 		// deducts (see einvoicingDiscountLabel()).
+		//
+		// Customer invoices only, afterPDFCreation() gating on instanceof Facture: FactureFournisseurLigne
+		// fills ->description and not ->desc before 20.0, so extending this guard to supplier invoices
+		// needs a ?: $line->description or every free line of an 18.0/19.0 purchase invoice reads as
+		// having no name.
 		$linesWithNoName = [];
 		if (!empty($invoice->lines) && is_array($invoice->lines)) {
 			foreach ($invoice->lines as $line) {
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -1473,6 +1473,32 @@ class EInvoicing
 			}
 		}
 
+		// BR-25: every line of the document names what it invoices (BT-153). The name is built from the
+		// label of the product, or from the first line of the description when there is no product, so a
+		// line holding neither is issued with an empty name and the document is refused - and refused by
+		// the platform, after transmission, on a line number the seller then has to go and find. Every
+		// such line is listed here instead, before anything is sent.
+		//
+		// Title and subtotal lines are not concerned: they are pseudo-lines that never reach the
+		// document. A discount line is not concerned either, its name being built from the piece it
+		// deducts (see einvoicingDiscountLabel()).
+		$linesWithNoName = [];
+		if (!empty($invoice->lines) && is_array($invoice->lines)) {
+			foreach ($invoice->lines as $line) {
+				if ((int) $line->product_type == 9 || !empty($line->fk_remise_except)) {
+					continue;
+				}
+				$hasLabel = trim((string) ($line->product_label ?? '')) !== '';
+				$hasDesc = trim(dol_string_nohtmltag((string) ($line->desc ?? ''), 0)) !== '';
+				if (!$hasLabel && !$hasDesc) {
+					$linesWithNoName[] = (int) ($line->rang ?: count($linesWithNoName) + 1);
+				}
+			}
+		}
+		if (!empty($linesWithNoName)) {
+			$baseErrors[] = $langs->trans("FxCheckErrorLinesWithNoName", implode(', ', $linesWithNoName));
+		}
+
 		if (!empty($baseErrors)) {
 			$res = -1;
 			$message .= '<br> Error: ' . implode('<br> Error: ', $baseErrors);
--- a/einvoicing/langs/en_US/einvoicing.lang
+++ b/einvoicing/langs/en_US/einvoicing.lang
@@ -729,3 +729,11 @@ EInvoicePreviewXml=Show the e-invoice XML
 EInvoiceNoFileToPreview=No e-invoice XML has been generated for this invoice.
 EInvoiceNoReceivedFileToPreview=No e-invoice XML has been received for this invoice.
 EInvoiceXmlReindented=received on a single line, shown reindented
+
+# Wording of a discount line whose source piece cannot be named (BT-153 / BT-97)
+EInvDiscountOnInvoice=on invoice %s
+DiscountFromCreditNoteNoSource=Discount from a credit note
+DiscountFromDepositNoSource=Down payment deducted
+DiscountFromExcessReceivedNoSource=Payment in excess deducted
+DiscountFromExcessPaidNoSource=Payment in excess deducted
+FxCheckErrorLinesWithNoName=The following lines have neither a product label nor a description, so the item name of the e-invoice (BT-153) would be empty, which rule BR-25 refuses: %s. Enter a description on each of them.
--- a/einvoicing/langs/en_US/einvoicing.lang
+++ b/einvoicing/langs/en_US/einvoicing.lang
@@ -736,4 +736,4 @@ DiscountFromCreditNoteNoSource=Discount from a credit note
 DiscountFromDepositNoSource=Down payment deducted
 DiscountFromExcessReceivedNoSource=Payment in excess deducted
 DiscountFromExcessPaidNoSource=Payment in excess deducted
-FxCheckErrorLinesWithNoName=The following lines have neither a product label nor a description, so the item name of the e-invoice (BT-153) would be empty, which rule BR-25 refuses: %s. Enter a description on each of them.
+FxCheckErrorLinesWithNoName=The following lines have neither a product label nor a description, so the item name of the e-invoice (BT-153) would be empty, which rule BR-25 refuses: %s. Enter a description on each of them (rank in the document, then line id).
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -838,19 +838,22 @@ foreach ($linesData as $numligne => $vals) {
 	}
 }
 
+foreach ($globalDiscounts as $discountIndex => $vals) {
+	if (in_array((string) ($vals['reason'] ?? ''), $discountSentinels, true)) {
+		dol_syslog("EInvoicing: allowance ".$discountIndex." of ".$object->ref." carries the unresolved discount marker ".$vals['reason']." in BT-97. The discount source piece could not be read.", LOG_ERR);
+	}
+}
+
 // BR-25: a line with no name is not a document the platform accepts, so it is refused here rather than
 // after transmission, on a line number the seller would then have to go and find. Every such line is
 // named at once: sending them back one refusal at a time would be a round trip per line. This is the
 // same missing data the pre-check reports before validation (validateInvoiceConfiguration()); a
 // document reaching this point with one is one whose lines changed since, or one built by a path that
-// does not run the pre-check.
+// does not run the pre-check. Refused after both halves of the last look above, never between them: a
+// document carrying a nameless line and an unresolved marker in BT-97 would otherwise leave without the
+// marker ever being reported - the very case that last look exists to catch.
 if (!empty($linesWithNoName)) {
 	throw new Exception('MISSINGDATA[BR-25]: The line'.(count($linesWithNoName) > 1 ? 's ' : ' ').implode(', ', $linesWithNoName).' of '.$object->ref.' '.(count($linesWithNoName) > 1 ? 'have' : 'has').' no item name (BT-153). Enter a description on the line, or a label on the product it invoices.');
-}
-foreach ($globalDiscounts as $discountIndex => $vals) {
-	if (in_array((string) ($vals['reason'] ?? ''), $discountSentinels, true)) {
-		dol_syslog("EInvoicing: allowance ".$discountIndex." of ".$object->ref." carries the unresolved discount marker ".$vals['reason']." in BT-97. The discount source piece could not be read.", LOG_ERR);
-	}
 }
 
 // Rounding convention of the totals: Dolibarr sums the amounts already rounded on each line ("total of
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -807,6 +807,29 @@ if (!empty($object->situation_counter) && $object->situation_counter > 1
 	}
 }
 
+// Last look for a sentinel that reached a field the customer reads. Everything above resolves the four
+// of them, so anything left here is a way of building a document that this file does not know about -
+// which is not a supposition: the resolution was written for the reason of a document level allowance
+// and the item name of a deposit line was found carrying the sentinel afterwards, at the second look.
+//
+// The test is an equality, never an inclusion: a line of work named 'Reprise (DEPOSIT) du chantier' is
+// a legitimate text and must go out untouched. And it reports rather than refuses - a marker in an item
+// name is ugly, not invalid, and holding back an invoice over it would cost the seller more than it
+// saves.
+$discountSentinels = array_keys(einvoicingDiscountSentinels());
+foreach ($linesData as $numligne => $vals) {
+	foreach (array('prodname' => 'BT-153', 'proddesc' => 'BT-154') as $field => $businessTerm) {
+		if (in_array((string) ($vals[$field] ?? ''), $discountSentinels, true)) {
+			dol_syslog("EInvoicing: line ".$numligne." of ".$object->ref." carries the unresolved discount marker ".$vals[$field]." in ".$businessTerm.". The line is a discount whose source piece could not be read.", LOG_ERR);
+		}
+	}
+}
+foreach ($globalDiscounts as $discountIndex => $vals) {
+	if (in_array((string) ($vals['reason'] ?? ''), $discountSentinels, true)) {
+		dol_syslog("EInvoicing: allowance ".$discountIndex." of ".$object->ref." carries the unresolved discount marker ".$vals['reason']." in BT-97. The discount source piece could not be read.", LOG_ERR);
+	}
+}
+
 // Rounding convention of the totals: Dolibarr sums the amounts already rounded on each line ("total of
 // round", the default), unless MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND rounds the sum instead. The loop
 // above always applies the first one, so follow the instance or the document claims a cent less than
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -440,7 +440,8 @@ foreach ($object->lines as $line) {
 	// The second method need to use the field BT-113. We don't use it as we use the first method.
 	$depositFactRef  = null;
 	$depositFactDate = null;
-	if ($line->desc == '(DEPOSIT)') {
+	$lineDiscount    = null;	// Discount the line was built from, when it is a discount line
+	if ($line->desc == '(DEPOSIT)' && !empty($line->fk_remise_except)) {
 		$isDepositLine   = 1;
 		$depositFactRef  = "";
 		$depositFactDate = new DateTime();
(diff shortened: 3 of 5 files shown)


daGrumpf-bxp:
Ran this on the bench before reviewing the code, on two cores: **18.0.10** and **22.0.5**. Method: the same invoices are built once, then their CII is regenerated twice — once with `main` (b7850b7), once with `main` + this branch merged locally — by swapping the `custom/einvoicing` symlink between two exported trees, with `function_exists('einvoicingDiscountSentinels')` printed on every run as the negative control. Nothing else changes between the two columns. The core's own `getRemainToPay()` / `getSumDepositsUsed()` / `getSumCreditNotesUsed()` are printed beside each case as the reference.

## What the two trees produce

| Case | `main` | this PR |
|---|---|---|
| A. Credit note converted into a discount, applied as a line | BT-97 `<ram:Reason>(CREDIT_NOTE)</ram:Reason>` | `Discount from credit note DD22-AV-2609-0020 (on invoice DD22-FA-2609-0163)` — totals unchanged (800.00 / 960.00) |
| B. Deposit deducted (stays a line) | BT-153 `<ram:Name>(DEPOSIT)</ram:Name>` | `Down payments from invoice DD22-AC-2609-0020` — totals unchanged (700.00 / 840.00) |
| C. Excess payment of 600 held against the invoice (`fk_facture`) | BT-113 **0.00**, BT-115 **1200.00** — the core says remain to pay **600** | BT-113 600.00, BT-115 600.00 |
| D. Deposit of 480 held against the invoice (`fk_facture`) | BT-113 **0.00**, BT-115 **1200.00** — the core says remain to pay **720** | BT-113 480.00, BT-115 720.00 |
| E/F. Line with no readable name | line emitted **with no `<ram:Name>` element at all** | refused, by the pre-check and by the generator, naming rank and line id |
| G. Line of work literally named `(DEPOSIT)`, carrying no discount | Basis **-250.00**, Grand **-300.00**, Due **-300.00** on a 300 € invoice | Basis 250.00, Grand 300.00, Due 300.00 — but see below |
| H. Discount whose reason was typed by hand | `Geste commercial 2026` | `Geste commercial 2026` (untouched) |

Same results on 18.0.10 and 22.0.5. C and D are the ones with money in them: the customer was asked twice for what they had already paid, and the PR's figures land exactly on the core's `getRemainToPay()`.

## One thing to fix: case G

`main` treats *any* line whose description is exactly `(DEPOSIT)` as a deducted deposit, without looking at `fk_remise_except`, and flips its sign: a 300 € invoice goes out as **-300.00**. This PR fixes that (`&& !empty($line->fk_remise_except)`), and the fix is not mentioned anywhere in the description — it deserves to be, it is the most expensive of the lot.

But the new label helper then renames that same line to **"Down payment deducted"**. `einvoicingDiscountLabel()` decides on the description alone; the test the core makes, and the one this PR quotes, is `$desc == '(DEPOSIT)' && $object->lines[$i]->fk_remise_except`. The second half is missing at the BT-153 call site (the BT-97 one is already inside `if ($line->fk_remise_except)`), so a line carrying no discount at all falls into the `...NoSource` wording, which is meant for a discount whose *source piece* cannot be read — a different situation.

`testNothingElseIsResolved()` does not cover it: every case there passes a real discount and a non-sentinel description. A sentinel with no discount is the missing row.

## BT-113 and #342

Worth a sentence in the description, because at first read this looks like it re-opens #342 ("do not add deposit lines to the prepaid total"). It does not, and the bench says why: a deposit *inserted as a line* is linked by `fk_facture_line` (`Facture::insert_discount()` → `link_to_invoice($lineid, 0)`, all cores 18→24), while this SQL filters on `fk_facture`. The two sets are disjoint, case B keeps BT-113 at 0.00 under both trees, and the line method of XP Z12-014 is untouched.

On how the amount is selected: the core computes this same sum, and it selects on the **type of the source piece**, not on the description text —

```php
// DiscountAbsolute::getSumDepositsUsed()
" WHERE rc.fk_facture_source=f.rowid AND rc.fk_facture = ".((int) $invoice->id)
" AND f.type = ".(int) $invoice::TYPE_DEPOSIT
// DiscountAbsolute::getSumCreditNotesUsed(): f.type IN (STANDARD, CREDIT_NOTE, SITUATION)
```

and `getRemainToPay()` = `total_ttc − getSommePaiement() − getSumDepositsUsed() − getSumCreditNotesUsed()`. All three exist on 18 → 24. Reading them instead of a three-value `IN` over the description would give the same figures (measured: identical on both cases above), keep the module in step with the core's own arithmetic by construction, and hand over `f.type` for free — which is what the 386/381 choice is really about, rather than deriving the document type code back from the description string.

## Two reachability notes, since both cases sound impossible

- **A deposit held by `fk_facture`** (case D) is not reachable from the invoice card in the default setup: `$filtercreditnote` excludes `(DEPOSIT)%`. It is reachable through the REST API — `useCreditNote()` calls `link_to_invoice(0, $id)` with no filter on 18 → 24 — and from the card itself when `FACTURE_DEPOSITS_ARE_JUST_PAYMENTS` is set, where the filter becomes `fk_facture_source IS NOT NULL`. In that mode the option's own semantics ("subtracted from payments only") agrees with counting it in BT-113.
- **A line with neither label nor description** cannot be entered from the card either: the free-line branch refuses an empty description on every core from 18 to 24 (`empty($product_desc)`, renamed `$line_desc` in 22). The shape that reaches it is a description made of markup only — `<p><br></p>`, what the WYSIWYG editor leaves behind when the text is deleted: the POST is not empty, so the card accepts it, and `dol_string_nohtmltag()` strips it to nothing. That is case F above, built without touching SQL, and `main` emits the line with no `<ram:Name>` at all. Citing that shape in the description would make the BR-25 half a lot more convincing.

## Where the 386 actually shows up

The document type code only reaches the XML on the EXTENDED profiles — `CII-DT-018` bans a `TypeCode` on `InvoiceReferencedDocument`, and `CIIProtocol` emits it under `isExtendedProfile()` only. Measured with `EINVOICING_XML_PROFILE=EXTENDEDFR`: the PR emits `386` for the deposit and `381` for the excess received (on EN16931, no `TypeCode` at all, on either tree). Fine as it is, just not visible on the default profile.

## Housekeeping

- The branch no longer merges cleanly on `main`: conflicts in `lib/einvoicing.lib.php`, `lib/buildinvoicelines.inc.php` and both lang files (the new end-of-loop block lands where #674 put the situation deduction). A rebase is needed before merge.
- `langs/fr_FR/einvoicing.lang` — translations other than `en_US` are not carried in a PR here, they come back through the translation platform.
- `ChangeLog.md` is regenerated from commit titles (`.agents/AGENTS.md`), so the entry can be dropped.

P-tite-tete:
Thank you for the bench — the two columns on cases C and D are worth more than the reasoning that led
to them, and case G would not have been found by reading the diff.

**Case G is fixed.** You are right about where the defect is, and about what it costs: a line carrying
no discount fell into the `...NoSource` wording, which is meant for a discount whose source piece
cannot be read, and stated something false about a line deducting nothing.

The second half of the core's test now lives in `einvoicingDiscountLabelOfLine()` rather than inline in
the condition, because the two call sites do not read a line the same way. The one writing BT-97 stands
inside a test on `fk_remise_except` and must still resolve when the discount behind it cannot be
fetched — that is what `...NoSource` exists for — so the guard could not simply move into
`einvoicingDiscountLabel()` without re-opening the leak it was written to close. Putting it on the line
instead keeps both call sites honest and makes the case testable, which a bare `&&` in the condition
would not have: `testASentinelNameOnALineWithNoDiscountIsLeftAlone()` now covers each of the four
sentinels with and without a discount behind the line, which is the row you found missing from
`testNothingElseIsResolved()`.

**The sign fix is now in the description**, with the figures. You are right that it was the most
expensive defect of the lot and that it was mentioned nowhere; it was found while looking at something
else, which is exactly how it ended up undocumented.

**On reading the core's own sums** — `main` has since done it, and better than this branch did:
`getSommePaiement()` + `getSumDepositsUsed()` + `getSumCreditNotesUsed()` for BT-113, and the
BT-25/BT-26 references joined on the type of the source invoice rather than derived back from a
description string. The whole amounts half of this branch is therefore gone, `IN` over descriptions
included, and with it the paragraph about #342 — which only ever described the query this branch
carried. Nothing of it is left to review.

**The `<p><br></p>` shape is in the description too.** It is a more convincing account of the BR-25
half than an empty description, which no core from 18 to 24 accepts anyway.

**Housekeeping**: rebased on `main` (c06deafe). `AllTests.php` is out as well, `main` having replaced
the hand written list with a `glob()`. The `ChangeLog.md` entry is gone from the whole series and not
only from its final state, and `langs/fr_FR/einvoicing.lang` is out.

One thing I have not done: re-run your bench after the rebase. The unit tests are covered — CI is green
on 18.0.0 and 24.0.0, phan included — but the end to end figures on cases C and D are no longer this
branch's to produce, since that code is now `main`'s.

daGrumpf-bxp:
Re-ran the bench after the rebase, since the branch changed shape. Same method as the first round —
the same invoices regenerated under two trees swapped by the `custom/einvoicing` symlink,
`main` (dcf08348) against `main` + this branch merged locally, with
`function_exists('einvoicingDiscountSentinels')` printed on every run as the negative control — but on
**eight cores this time**: 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1 and 25.0.0-alpha.

## What the two trees produce

| Case | `main` | this PR |
|---|---|---|
| A. Credit note applied as a line (BT-97) | `<ram:Reason>(CREDIT_NOTE)</ram:Reason>` | `Discount from credit note DD22-AV-2609-0021 (on invoice DD22-FA-2609-0171)` — totals unchanged |
| B. Deposit deducted, stays a line (BT-153) | `<ram:Name>(DEPOSIT)</ram:Name>` | `Down payments from invoice DD22-AC-2609-0022` — totals unchanged |
| C/D. Excess and deposit held by `fk_facture` | BT-113 600.00 / 480.00 | identical — this is `main`'s code now |
| E/F. Line with no readable name | line emitted **with no `<ram:Name>` element at all** | refused by the pre-check and by the generator, rank and rowid named |
| **G. Line of work named `(DEPOSIT)`, no discount** | Basis **-250.00**, Grand **-300.00**, Due **-300.00** | **250.00 / 300.00 / 300.00**, and the line keeps its name |
| H. Discount whose reason was typed by hand | `Geste commercial 2026` | untouched |

Same result on all eight cores.

**Case G is fixed, both halves of it.** The line keeps the name it was given and the sign is right. The
guard on the line rather than on the description is the right place: `einvoicingDiscountLabelOfLine()`
makes the two call sites read the line the same way while leaving the `...NoSource` wording reachable
from BT-97, which a bare `&&` in the condition would not have.

**And case G is worth more than the description says.** The document `main` produces there is not
merely ugly — I put both files through the platform's validator: `main`'s `-300.00` invoice comes back
**rejected** (`res=-1`), this branch's comes back accepted (`res=1`). A 300 € invoice going out as a
document the platform refuses is the strongest line of this branch.

## What the new BR-25 refusal does *not* fire on

Built as a second bench, because a refusal added to the generation path is worth measuring from the
other side. None of these is refused, on either tree:

- a line invoicing a product with no description of its own — BT-153 is the product label
  (`Cle a molette 12mm`);
- a title pseudo-line with an empty description (`product_type` 9, `special_code` 104);
- a line of work quoting a sentinel inside a longer text — `Reprise (DEPOSIT) du chantier` goes out
  untouched, which is the equality-not-inclusion rule holding.

And the two shapes that must resolve do: a discount cut loose from its source piece comes out
`Discount from a credit note` (the `...NoSource` wording), and an `(EXCESS RECEIVED)` applied as a line
comes out `Payments in excess of invoice DD22-FA-2609-0377` in BT-97.

## Regression

- PHPUnit, whole grid, 8 cores × 30 files = 240 runs: no new failure. `SituationInvoiceTotalsTest`
  fails identically on `main`; the other reds of the parallel run are flakes and come back green run
  sequentially. `DiscountSentinelTest`: 8 tests, 45 assertions, OK on all eight cores.
- QA matrix, 126 generations over 9 instances, CII and Factur-X: `code=10` everywhere, identical
  between the two trees.
- `phpcs` clean on the four files; `phan` clean over the whole repository (`--analyze-twice`, real
  `ast` extension).
- Still merges cleanly on `main` at dcf08348 — a1a872f4 is the only commit to have touched these files
  since c06deafe and it does not collide.

## Three notes, none of them blocking

**1. The resolved text goes out half translated on a French document.** `DiscountFromCreditNote` comes
from the core in the language of the document, but `EInvDiscountOnInvoice` only exists in `en_US`, so
23.0.4 and 24.0.1 emit in BT-97:

```
Remise issue de l'avoir DD23-AV-2609-0052 (on invoice DD23-FA-2609-0451)
```

`en_US` only in a PR is the rule here and the platform delivers the rest, so this resolves on its own.
It is worth knowing all the same that this one is more visible than the usual untranslated string: it
lands in the middle of a sentence the core has already translated, in the XML the customer receives,
not in a screen.

**2. The last look runs in two halves and the refusal sits between them.** The `throw` for BR-25 is
placed after the loop over `$linesData` and before the loop over `$globalDiscounts`, so on a document
carrying both a nameless line and an unresolved marker in BT-97, the BT-97 report never happens — the
very case that last look exists to catch. Moving the `throw` below both loops costs nothing.

**3. A trap for later, not a defect now.** `validateInvoiceConfiguration()` reads `$line->desc`, which
`FactureFournisseur::fetch_lines()` does not set on 18.0 and 19.0 — it fills `->description` only, and
`->desc` appears from 20.0. On those two cores the pre-check would report every free line of a supplier
invoice as having no name. It is unreachable today, `afterPDFCreation()` gating on `instanceof Facture`,
so nothing to change here; worth a line so the guard is not later extended to supplier invoices without
a `?: $line->description`.

P-tite-tete:
Thank you for the second bench — the validator result on case G is worth more than the whole description of the branch, and I would not have thought to put the two documents through it. Two of the three notes are addressed in `9e2ccec5` and `d30057c2`.

**Note 2 — the refusal sitting between the two halves.** Fixed. You are right that the case the last look exists to catch was the one it could not report: the `throw` is now placed below the loop over `$globalDiscounts`, so both halves always run and a document carrying a nameless line *and* an unresolved marker in BT-97 names the marker in the log before the refusal is raised. The comment above the `throw` says why the placement matters, so it does not drift back between the loops later.

I have no unit test for it: `buildinvoicelines.inc.php` is an include that expects a full invoice object, and `DiscountSentinelTest` covers the helpers in `einvoicing.lib.php`. The ordering is guarded by the comment, not by a test — worth saying plainly rather than implying the grid covers it.

**Note 3 — `->desc` on a supplier invoice before 20.0.** Not a change of behaviour, a comment above the guard: `FactureFournisseurLigne` fills `->description` alone before 20.0, so extending the pre-check to supplier invoices needs a `?: $line->description` or every free line of an 18.0/19.0 purchase invoice reads as having no name. Unreachable today, `afterPDFCreation()` gating on `instanceof Facture`.

**Note 1 — the half-translated BT-97.** Left as it is: en_US only in a PR is the rule here. Your point that this one lands mid-sentence in the XML the customer receives, rather than on a screen, is a fair distinction and is worth carrying into the translation round.

What I re-ran on my side is narrower than your bench: `DiscountSentinelTest` (8 tests, 45 assertions, OK) on 23.0.4 only, and `php -l` on the two files. I have neither phpcs nor phan set up locally, and no multi-core grid. Still merges cleanly on `main` at dcf0834 — checked with `git merge-tree`, `einvoicing.class.php` auto-merges, `a1a872f4` touches lines well above this branch's.

daGrumpf-bxp:
Both commits re-measured, same method as before — the same invoices regenerated under two trees swapped
by the `custom/einvoicing` symlink, `main` (now **d2739fab**, it moved) against `main` + this branch
merged locally, `function_exists('einvoicingDiscountSentinels')` printed on every run as the negative
control, on the eight cores 18.0.10 → 25.0.0-alpha.

**Nothing moved.** Cases A→H and I→M come out exactly as in the previous round on all eight: BT-97 and
BT-153 resolved, E/F refused with the rank and the rowid, G at `250.00 / 300.00 / 300.00` with the line
keeping its name, H and M untouched, J on the `…NoSource` wording, K naming the invoice. PHPUnit grid:
248 runs against `main`'s 240, same 17 reds on both trees — `SituationInvoiceTotalsTest` on all eight
either way, the rest flakes of the parallel run that come back green one at a time, plus
`MandatoryThirdpartyExtrafieldTest` on 18.0 and `SupplierInvoiceHelperTest` on 20.0 which fail
identically on `main`. `DiscountSentinelTest`: 8 tests, 45 assertions, OK on all eight.
`php -l` and `dolics` clean on the four files. Still merges cleanly on `main` at d2739fab.

## Note 2 — and why you were right not to claim a test

The placement is correct now, and I went looking for the case to exercise it. It cannot be reached on
today's code, which is worth writing down because it is the real reason there is no test to write:

`einvoicingDiscountLabel()` returns a text for **all four** sentinels even with no discount object at
all — measured, `$outputlangs` in `en_US`:

```
(CREDIT_NOTE)      -> "Discount from a credit note"
(DEPOSIT)          -> "Down payment deducted"
(EXCESS RECEIVED)  -> "Payment in excess deducted"
(EXCESS PAID)      -> "Payment in excess deducted"
```

So `$discountReason` is empty exactly when the description is *not* a sentinel, and the fallback
`$discountReason ?: ($discount->description ?? 'REMISE')` can therefore never put a sentinel into
`'reason'`. The other `$globalDiscounts[]` site writes a translated string. The BT-97 half of the last
look is a net for a construction path that does not exist yet — which is what your comment now says,
and the comment is the right guard for it. No objection.

## One thing the second bench turned up: the last look calls case G a discount

The line of work an operator named `(DEPOSIT)`, the one this branch fixed so that it keeps its name and
its sign, is reported by the last look on its way out:

```
ERR EInvoicing: line 1 of DD23-FA-2609-0459 carries the unresolved discount marker (DEPOSIT)
    in BT-153. The line is a discount whose source piece could not be read.
```

It is not a discount, and there is no source piece — the branch's own `einvoicingDiscountLabelOfLine()`
established that two hundred lines above, on `fk_remise_except`. Only a log line, nothing reaches the
customer, and the document is correct. But it is the last look crying wolf on the one shape the branch
declares legitimate, in `LOG_ERR`, and a reader chasing that message would land on a line that has
nothing wrong with it.

The BT-153 half is missing the second half of the test the BT-97 half doesn't need. `$lineRowIds` is
already filled in that loop; a sibling would do:

```php
$lineDiscountIds[$numligne] = (int) $line->fk_remise_except;
```

and in the last look, `&& !empty($lineDiscountIds[$numligne])` on the `prodname`/`proddesc` report. The
equality against the sentinel list stays what it is — this only stops the report on a line that carries
no discount, which is the same rule as the resolution.

Measured on 23.0.4, the instance where the file log is on; the path is unconditional, so it fires on all
eight.

## Heads up, unrelated to the code

`.agents/AGENTS.md` grew a `## Comments` section on 05/09: never more than 5 lines, and never more than
the number of code lines added. This branch is at 121 comment lines for 109 of code (52%), with prose
blocks of 6, 8, 9 and 14 lines in `buildinvoicelines.inc.php` and `einvoicing.class.php` —
`d30057c2` took the BR-25 block from 9 lines to 14. I am not the one to enforce it and none of it is
wrong, but I went through the same pass on my own PRs last week and it is cheaper to do before the
review than after: the long reasoning moves to the PR body, the comment keeps the one sentence the
reader of the code cannot deduce.

P-tite-tete:
Your comment landed a few hours before eldy merged this, so it never got an answer here — both points are now in #936.

The last look crying wolf on case G is fixed the way you sketched it: `fk_remise_except` kept per line beside the rowid the loop already collects, and the BT-153/BT-154 report asking for it before firing. You were right that the report had to apply the same rule as the resolution rather than a second reading of the text. And thank you for measuring that the BT-97 half cannot be reached today — `einvoicingDiscountLabel()` answering for all four sentinels with no discount object is the part I had not checked, and it is what makes the asymmetry between the two halves correct rather than an oversight.

The comment pass is in the same PR: the blocks of nine and fourteen lines are back to five, in `buildinvoicelines.inc.php` and `einvoicing.class.php` and also in the prose above `einvoicingDiscountSentinels()` and `einvoicingDiscountLabelOfLine()`, with the reasoning moved to the PR body. Three files, 32 lines added against 61 removed. Point taken that this is cheaper before the review than after.

PR #728, opened 2026-09-02
FIX [einvoicing] Shipping/delivery subtotal detection uses removed fk_origin_line property

hregis:
## Summary
- `_isLineFromExternalModule()` (used to skip title/subtotal pseudo-lines when building invoice XML from a shipment or delivery) fetched the origin order line via `$line->fk_origin_line`.
- Dolibarr 20 renamed this property to `fk_elementdet` on `ExpeditionLigne`/`LivraisonLigne`, with no backward-compat alias kept in core.
- Since the rename, on Dolibarr 20+ this line always resolved to a null id, so the origin order line was never fetched correctly and subtotal-line detection silently broke for shipping/delivery-sourced invoices.
- This module declares `need_dolibarr_version = array(18, -3)`, so the fix reads whichever property the line object actually carries (`property_exists()` check) instead of assuming one name, keeping support for Dolibarr 18-19 as well as 20+.

## Test plan
- [x] `php -l` on the changed file (PHP 7.4 and 8.3)
- [x] Verified property rename timing against local Dolibarr checkouts: `fk_origin_line` present through Dolibarr 19, renamed to `fk_elementdet` starting Dolibarr 20, unchanged through 23/24
- [x] Added `test/phpunit/ExternalModuleLineDetectionTest.php`, a regression test covering the current-core property (`fk_elementdet`), the legacy one (`fk_origin_line`, Dolibarr 18-19), and that the current one takes priority when both are present. Run locally with DOLIBARR_HTDOCS pointed at real Dolibarr 23 and 24 instances: fails against the pre-fix code (`Undefined property: stdClass::$fk_origin_line` + wrong result), passes with the fix, on both versions.


daGrumpf-bxp:
I ran this against real Dolibarr installs before it lands, because the `shipping` / `delivery` branch looked unreachable to me. The property rename is real, but the branch that reads it cannot be entered — so I would rather delete it than teach it a second property name.

## 1. One call site, and it can never pass `shipping` / `delivery`

`_isLineFromExternalModule()` is called from exactly one place, `einvoicing/lib/buildinvoicelines.inc.php:392`:

```php
$isSubTotalLine = $this->_isLineFromExternalModule($line, $object->element, 'modSubtotal');
```

and `$object` is built 320 lines above it, at `buildinvoicelines.inc.php:72`:

```php
$tmpfacture = new Facture($db);
$object = $tmpfacture->fetch($invoice->id) > 0 ? $tmpfacture : $invoice;
```

`Facture::$element` is `'facture'` from 18 to 24, and its lines are `FactureLigne`. An invoice created *from* a shipment is no exception: `compta/facture/card.php` fetches the `Expedition` and copies its lines through `Facture::addline()`, so the objects walked here are never `ExpeditionLigne` / `DeliveryLine`.

## 2. Live run of the exact scenario, on four cores

Order carrying a `product_type 9` title line → shipment → invoice created from that shipment (`llx_element_element`: `shipping#N -> facture#M`, `commande#N -> facture#M`) with a title line on the invoice too → CII generation, on the pre-fix code:

| core | 2nd argument received | line classes | title line in the XML |
|---|---|---|---|
| 18.0.10 | `facture` | `FactureLigne` | skipped (2 lines → 1 `IncludedSupplyChainTradeLineItem`) |
| 20.0.4 | `facture` | `FactureLigne` | skipped |
| 23.0.4 | `facture` | `FactureLigne` | skipped (2 → 1) |
| 24.0.0 | `facture` | `FactureLigne` | skipped (2 → 1) |

Identical below and above the rename: subtotal detection is not broken on 20+ for shipment-sourced invoices, because that code does not run. The `FactureLigne` objects carry neither `fk_origin_line` nor `fk_elementdet`.

(One honest caveat: the 20.0.4 instance has no warehouse, so `Expedition::addline()` refused the line and the invoice ended up with the title line alone — 1 line → 0 line items, still correctly skipped.)

## 3. Positive control: the defect is real, but only reachable by reflection

Forcing the branch through reflection with a **real** `ExpeditionLigne` whose origin line is the order's title line:

* 18.0.10 → `true` (correct)
* 23.0.4 and 24.0.0 → `false` (wrong)

so the analysis itself holds. That is also the only way the added tests reach it: they pass `stdClass` mocks with the property set by hand, which proves the method's behaviour on a fabricated input, not that any input of that shape ever arrives.

## Two details the cores contradict

* `DeliveryLine` still declares `fk_origin_line` on 20.0.4, 23.0.4 and 24.0.0 (`delivery/class/delivery.class.php`). Only `ExpeditionLigne` was renamed (`fk_origin_line` on 18/19 → `fk_elementdet` from 20, and moved to `expedition/class/expeditionligne.class.php` in 23/24). The `delivery` half of the description never changed.
* On a real `ExpeditionLigne` the read is silent, not a notice: `DolDeprecationHandler::__get()` answers it, `isDynamicPropertiesEnabled()` returns `true` and `fk_origin_line` is not in `deprecatedProperties()`, so the value is plain `null`. The `Undefined property: stdClass::$fk_origin_line` quoted in the test plan comes from the mock, not from a core.

## Proposal: delete the branch instead

It arrived with the initial factorization (`1a4e957`, still under the `pdpconnectfr` name), copied from the core PDF builders where `$object` can legitimately be a shipment. Here it cannot. Rather than maintaining a two-name compatibility shim on a path that cannot execute, I would drop it, along with the then-unused `$element` argument:

```php
	/**
	 *    Check line type from external module ?
	 *
	 * @param  object $line       line we work on
	 * @param  string $searchName module name we look for
	 * @return boolean            true if the line is a special one and was created by the module we ask for
	 */
	private function _isLineFromExternalModule($line, $searchName)
	{
		if ((int) $line->product_type != 9) {
			return false;
		}
		// ... unchanged
```

with the call site becoming `$this->_isLineFromExternalModule($line, 'modSubtotal');`.

If the intent is the opposite — to let the XML builder accept a shipment or a delivery as source object some day — then this fix is right but incomplete, since nothing else in `buildinvoicelines.inc.php` handles a non-invoice object. Happy to open the delete PR if you agree.

Bench: Dolibarr 18.0.10, 20.0.4, 23.0.4 and 24.0.0, module on `main` (pre-fix code, verified identical to this PR's base).

daGrumpf-bxp:
Opened #730 with the delete, so the two options are side by side and the maintainers can pick. It carries the measurements above plus an A/B PHPUnit run (72 OK / 0 KO on 18.0.10, 20.0.4, 23.0.4 and 24.0.0, before and after). If the direction is instead to make the XML builder accept a shipment or a delivery as source object, close #730 and this one is the right base.

hregis:
Good catch, thank you for the thorough verification.

You're right on both counts:

- `_isLineFromExternalModule()` has exactly one call site (`buildinvoicelines.inc.php:392`), and `$object` there is always a `Facture` (see the explicit `@var Facture $object` and the `$tmpfacture = new Facture($db); $object = $tmpfacture->fetch(...) > 0 ? $tmpfacture : $invoice;` a few lines above). An invoice built from a shipment still walks `FactureLigne` objects, never `ExpeditionLigne`/`DeliveryLine` directly. So `$element` is always `'facture'`, and the `shipping`/`delivery` branch this PR touches cannot be reached today. My fix, while accurate about the property rename, was fixing a branch that never runs.
- I also mischaracterized "delivery" as affected by the same rename. It isn't: `DeliveryLine` (`delivery/class/delivery.class.php`) still carries `fk_origin_line` unchanged through 18/20/23/24 — only `ExpeditionLigne` was renamed to `fk_elementdet` (and moved to `expedition/class/expeditionligne.class.php` in 23/24, as you noted). I mixed the two up.

And fair point on the test: the `stdClass` mocks show the method's behavior on a fabricated shape, not that any real input reaches it that way — a real `ExpeditionLigne` degrades silently through `DolDeprecationHandler`, not with a notice.

Closing this in favor of #730, which addresses the actual (unreachable) state of the branch with live verification across four cores. Thanks for taking the time to run this against real installs.

Issue #731, opened 2026-09-02
[einvoicing] A document level charge (BG-21) of a received document never reaches the imported invoice


daGrumpf-bxp:
### Bug

A **document level charge** of a received e-invoice never reaches the imported supplier invoice, so
the invoice totals less than the document it came from — silently, with no warning anywhere.

Found while instrumenting #726, whose invoice loses two different amounts. This report is the second
one; the line-level half is #726 itself.

### What the norm says

EN 16931 puts two symmetric groups at document level:

- **BG-20**, a *document level allowance* — it **subtracts** from the total;
- **BG-21**, a *document level charge* — it **adds** to it. Freight, packaging, a handling fee.

and `BR-CO-13` makes a charge part of what is owed, exactly like a line is:

```
[BR-CO-13]-Invoice total amount without VAT (BT-109) = Σ Invoice line net amount (BT-131)
           - Sum of allowances on document level (BT-107) + Sum of charges on document level (BT-108).
```

A charge is fully specified: **BT-99** amount, **BT-102** VAT category, **BT-103** rate, **BT-104**
reason or **BT-105** reason code, summed into **BT-108**. `BR-36`, `BR-37` and `BR-38` make the first
three mandatory, and `BR-CL-20` restricts the reason code to UNCL 7161.

### Where it is lost

In CII an allowance and a charge are the **same element**, `ram:SpecifiedTradeAllowanceCharge`, told
apart only by `ram:ChargeIndicator/udt:Indicator`. Everything is parsed — indicator, reason, reason
code, amount, VAT category and rate — and then `CIIProtocol::createHeaderDiscounts()` keeps the
allowances and drops the charges:

```php
foreach ($headerAllowancesCharges as $index => $allowanceCharge) {
    // Skip charges
    if (($allowanceCharge['indicator'] ?? '') !== 'false') {
        continue;
    }
```

It is understandable rather than careless: an allowance has a home in Dolibarr — a `DiscountAbsolute`,
inserted as a discount line further down — while a supplier invoice holds **no positive amount at
document level**, so a charge had nowhere to go. But dropping it changes the amount owed.

The same gap exists on the way out: `buildAllowanceChargeNode()` takes an `$isCharge` flag, and both
of its callers pass `false`. The module never emits a BG-21 either.

### Steps to reproduce the behavior and expected behavior

A received document, EN 16931 or EXTENDED CTC-FR, whose header carries

```xml
[...]

Issue #726, opened 2026-09-02
Supplier invoice with a line without quantity


frederic34:
### Bug

<img width="1038" height="440" alt="Image" src="https://github.com/user-attachments/assets/14ac5c18-7b97-4f6f-8a38-d6db92a7a4ee" />

<img width="1407" height="264" alt="Image" src="https://github.com/user-attachments/assets/b9069728-3e92-43f3-9fa2-b5a07a47c9ad" />

<img width="800" height="323" alt="Image" src="https://github.com/user-attachments/assets/51317079-bf79-4978-9c30-8c4f489aeaa2" />

### Module

einvoicing

### Steps to reproduce the behavior and expected behavior

received a new invoice with bad total

daGrumpf-bxp:
Could you attach **the original XML** please? Without it I cannot tell whether the sender is at fault
or we are, and the norm answers differently in the two cases.

It is the file the module stored beside the supplier invoice when it imported the flow: *Documents*
tab of that supplier invoice, name ending in `_einvoice.xml`. If you would rather not publish a real
document, the two parts that decide are enough:

- the `<ram:GuidelineSpecifiedDocumentContextParameter><ram:ID>` of the header (the profile URN);
- the whole `<ram:IncludedSupplyChainTradeLineItem>` of the line numbered **899**, the "COMMENT" one
  at 12.00.

Your Dolibarr version and module version would help too.

## Why those two, and what the norm says

**BR-22** is fatal, and it is a presence test — a quantity of `0` satisfies it, an absent element
does not (FNFE schematron, `EN16931-CII-validation-preprocessed.sch`):

```xml
<assert id="BR-22" flag="fatal" test="(ram:SpecifiedLineTradeDelivery/ram:BilledQuantity)">
  [BR-22]-Each Invoice line (BG-25) shall have an Invoiced quantity (BT-129).</assert>
```

So on the **EN 16931** profile, a line carrying no `ram:BilledQuantity` at all makes the document
non-conformant, and the sender is the one to fix.

But your two lines are labelled **COMMENT**, and on the **EXTENDED CTC-FR** profile that is exactly
the case the French rules carve out. `BR-FREXT-BR-22` re-states BR-22 *conditionally*:

> [BR-FREXT-BR-22]-Each invoice item "INVOICE LINE" (BG-25) must contain the quantity of goods or
[...]

PR #736, opened 2026-09-02
FIX #735 [einvoicing] An invoice line charge (BG-28) stops being lost and corrupting the discount of its line

### LINKED ISSUE #735: [einvoicing] An invoice line charge (BG-28) is lost on import and corrupts the discount of its line
### Bug

An **invoice line charge** (BG-28) of a received document is not carried into the imported invoice, and
it also corrupts the discount of the line that holds it.

**Where this comes from, honestly:** not from a user report. It was found while auditing the four
allowance/charge groups after #731, by building a document for each of them. The invoice of #726 that
started all this carries **no** line level allowance or charge — I checked. So this is a defect proven
on a document I wrote, not one seen in the field, and whether it is worth fixing is a call for the
maintainers rather than mine. What follows is what I measured.

### What the norm says

EN 16931 mirrors at line level what it has at document level: **BG-27**, an invoice line allowance, and
**BG-28**, an invoice line charge (BT-141 amount, BT-142 base, BT-143 percentage, BT-144 reason,
BT-145 reason code). `BR-43` and `BR-44` make the amount and a reason mandatory.

The line net amount, **BT-131**, is what the rest of the document is built on — `BR-CO-10` sums it into
BT-106, and `BR-CO-13` turns that into the invoice total. `BR-24` makes BT-131 mandatory on every line.
Whatever the components, **BT-131 is the authoritative amount of the line**.

Note that in CII a line level allowance or charge carries no VAT of its own — `CII-SR-191` forbids
`ram:CategoryTradeTax` there — so it is taxed at the rate of the line it belongs to.

### Where it is lost

`CIIProtocol::resolveLineDiscountPercent()` filters the line entries down to the allowances, as its
comment says, and then rebuilds the gross price as if charges did not exist:

```php
// Keep only allowances (indicator = "false"), ignore charges (indicator = "true")
...
'priceWithoutDiscount' => (float) $lineTotalAmount + $totalDiscountAmount,
```

`$lineTotalAmount` is BT-131, which **already includes the charge**. So the charge is not merely
ignored: it is folded into the base the discount percentage is applied to, and the line comes out at
neither the right amount nor the right unit price.

One line, quantity 5 at a net price of 100.05, a line allowance of 10 % (50.03) and a line charge of
7.00, so `LineTotalAmount` = 500.25 − 50.03 + 7.00 = **457.22**:

```xml
<ram:SpecifiedLineTradeSettlement>
    <ram:CalculationPercent>10.00</ram:CalculationPercent>
    <ram:BasisAmount>500.25</ram:BasisAmount>
    <ram:ActualAmount>50.03</ram:ActualAmount>
    <ram:ReasonCode>95</ram:ReasonCode>
    <ram:ChargeIndicator><udt:Indicator>true</udt:Indicator></ram:ChargeIndicator>
    <ram:ActualAmount>7.00</ram:ActualAmount>
    <ram:ReasonCode>FC</ram:ReasonCode>
    <ram:Reason>Handling</ram:Reason>
  <ram:SpecifiedTradeSettlementLineMonetarySummation>
    <ram:LineTotalAmount>457.22</ram:LineTotalAmount>
  </ram:SpecifiedTradeSettlementLineMonetarySummation>
</ram:SpecifiedLineTradeSettlement>
```

Import it. **Expected**: 457.22 HT. **Observed**: **456.52 HT** — 0.70 short, and the unit price stored
is 101.45 where the document says 100.05.

Measured on Dolibarr **18.0.10, 23.0.4 and 24.0.0**, same result on the three. The document validates
against the FNFE schematron (`fnfempe/France_RFE`) before being imported.

### Note

With #729 applied this stops being silent — the import reports
`Line 1 of the received document announces a net amount (BT-131) of 457.22, but its quantity and unit
price rebuild 456.52` — but the amount is still wrong.

### Module

einvoicing

### Dolibarr Version

18.0.10, 23.0.4, 24.0.0

### Environment PHP

8.4

### Environment Database

MariaDB 11.4

daGrumpf-bxp:
Closes #735. **Rebased on `main` now that #729 and #732 are merged**, so it applies cleanly and there is
nothing left to resolve.

Third and last of the allowance/charge audit that started on #726. **It did not come from the reported
invoice** — that document carries no line level allowance or charge, I checked — it came from testing
the four groups one by one after #731. Whether it is worth carrying is therefore a call for you rather
than for me; what follows is what I measured.

## The defect

EN 16931 mirrors at line level what it has at document level: **BG-27**, an invoice line allowance, and
**BG-28**, an invoice line charge. A Dolibarr line holds a quantity, a unit price and a discount
percentage — it can express the allowance and has nowhere to put the charge.

`resolveLineDiscountPercent()` filters the entries down to the allowances, as its comment says, and then
rebuilds the gross price of the line from BT-131:

```php
// Keep only allowances (indicator = "false"), ignore charges (indicator = "true")
...
'priceWithoutDiscount' => (float) $lineTotalAmount + $totalDiscountAmount,
```

`$lineTotalAmount` is BT-131, and **BT-131 already contains the charge**. So the charge was not merely
ignored: it was folded into the base the percentage was applied to, and the line came out at neither
the amount nor the unit price the document gives.

## The change

- The charge amount is **taken back out of BT-131** before the gross is rebuilt, so the unit price
  stored is again BT-146 — 100.05 on the case below, where it used to be 101.45.
- The charge then **follows its line as a line of its own**, the way it would be keyed in by hand, so
  the two together are worth the BT-131 the document announces. Same shape as #732 does at document
  level, and for the same reason: BT-131 is the authoritative amount of the line (`BR-24` makes it
  mandatory, `BR-CO-10` sums it into BT-106, `BR-CO-13` turns that into the invoice total).
- It is taxed at the **rate of the line it belongs to**: in CII a line level charge carries no VAT of
  its own, `CII-SR-191` forbidding `ram:CategoryTradeTax` there.
- The label is the reason of the document (BT-144), or its reason code (BT-145) behind a translated
  label — `BR-44` accepts either — followed by the line it belongs to, so several charges stay
  readable. Two `en_US` keys.

A line with **no** charge resolves exactly as before: the subtraction is of zero. That is every
document received so far, and the tests pin it.

## Tested

`LineChargeTest` (new, registered in `AllTests`): 7 tests, 28 assertions, green on the seven supported
cores. It pins the unit price no longer being inflated, the charge becoming its own line at the line's
VAT rate, the labelling with and without BT-144, several charges on one line, a charge with no
allowance at all — and, as the control, that a line with an allowance only, or nothing, or a charge
worth zero, adds no line and resolves identically.

Live on the bench, documents imported through `CIIProtocol::createSupplierInvoiceFromSource()` and read
back from a second process. Both were validated first against the FNFE schematron
(`fnfempe/France_RFE`, EUPL) — **VALID**, so what follows is measured on conformant documents:

| document | announced | `main` (with #729 and #732) | this PR |
|---|---|---|---|
| a line of 5 × 100.05, allowance 50.03 **and charge 7.00** (BT-131 = 457.22) | 548.66 TTC | 547.82 ❌ | **548.66** ✅ |
| the same line with the **charge alone**, no allowance (BT-131 = 507.25) | 608.70 TTC | 600.30 ❌ | **608.70** ✅ |
| control: a document level allowance (BG-20), no line charge | 534.27 | 534.27 | **534.27** unchanged |
| control: a document level allowance **and** charge | 546.75 | 546.75 | **546.75** unchanged |
| control: the synthetic shape of #726 | 567.16 | 567.15 | **567.15** unchanged |
| control: the real invoice of the reporter | 1143.24 | 1143.24 | **1143.24** unchanged |

Identical on **Dolibarr 18.0.10, 23.0.4 and 24.0.0**. The stored line reads
`qty 5, pu_ht 100.05, total 450.22` plus `Handling - charge of line 1, qty 1, pu_ht 7.00, total 7.00`
— 457.22, which is the BT-131 of the document.

PHPUnit, the whole suite on the seven cores: **133 runs, 133 OK**. The four list hooks answer on the
four list contexts on all seven. `phpcs` clean with the Dolibarr ruleset, and Phan replayed locally
reports nothing this branch adds.

## Where it sits

Three defects were found on one invoice, and they are three separate pull requests because they are
three different methods:

| | issue | PR |
|---|---|---|
| a line whose quantity cannot rebuild its amount | #726 | #729 |
| a document level charge (BG-21) | #731 | #732 |
| **an invoice line charge (BG-28)** | **#735** | **this one** |

The first two are merged. This one is the hole the audit turned up next door.

For the record, the four groups now stand like this at import, each measured against a document
validated by the FNFE schematron:

| group | before the three PRs | today |
|---|---|---|
| **BG-27** invoice line allowance | correct | correct |
| **BG-20** document level allowance | correct | correct |
| **BG-21** document level charge | lost | fixed by #732 |
| **BG-28** invoice line charge | lost, and it corrupted its line | **this PR** |

Round trip through the Access Point sandbox on the rebased branch: `allfullx.sh` → **14/14 complete
cycles**, the seven versions in both directions, each with the whole lifecycle
`[200, 201, 205, 211, 212]`. None of those documents carries a line level charge, which is the point:
it is the non-regression control, and the change is inert on everything that does not hit it.


einvoicing/class/protocols/CIIProtocol.class.php
```diff
@@ -3346,6 +3354,73 @@ protected function determineDeliveryDatesAndCustomerOrderNumbers(&$customerOrder
...
+			return array();
+		}
+
+		// The import runs from a cron job as well as from a page, so the language file of the module is
+		// not necessarily loaded: without this the label of a charge would be its own translation key.
+		$langs->load('einvoicing@einvoicing');
+
+		$chargeLines = array();
+
+		foreach ($parsedLine['lineAllowances'] as $allowanceCharge) {
```
eldy, line 3386: You convert the allowance as a discount percent.
I want to be sure you know we can add linesthat are discount of a fixed amount in dolibarr.
Weadd a line int llx_facturfourn_det (line of invoice) with a link to llx_societe_remise via table llx_ociete_remise_supplier. This table llx_societe_remisecan be used to store commercial discount. It reduces the total amount of invoice, like do a negative line but with tracking that a flat discount was applied on invoice (not a relative discount on a given line).

So can you confirm we need to use a relative discount on a existing line here and not adding a second line that is a flat discount for the use case you want to manage ?

daGrumpf-bxp: Yes — confirmed, and here is the reasoning.

First, a scope note: the percent conversion you are commenting on is pre-existing code in `main` (`resolveLineDiscountPercent()`); this PR only subtracts the line **charge** from BT-131 before that base is computed. The flat-discount object is also already used by the module — document level allowances (BG-20) go through `DiscountAbsolute` (`CIIProtocol.class.php:3623`), which is exactly the right object there.

For what this PR fixes, the flat discount cannot be used:

- **Sign.** `FactureFournisseur::insert_discount()` writes `subprice = -$remise->amount_ht` (identical from 18.0 to 24.0). BG-28 is an invoice line **charge**, a positive amount, and the core has no "flat charge" counterpart.
- **Level.** `insert_discount()` creates an independent line (`qty=1`, `rang=-1`, `info_bits=2`) with no link back to the line it belongs to. EN 16931 puts BG-27/BG-28 **inside** BT-131 of their own line: BR-24 makes BT-131 mandatory, BR-CO-10 sums it into BT-106, and BR-CO-13 only brings in the **document** level BT-107/BT-108. The VAT breakdown rules (BR-S-08, BR-E-08, BR-AE-08…) say the same: the taxable amount per rate is Σ BT-131 plus/minus document level allowances and charges only — a line level one has no VAT of its own (CII-SR-191). Routing it to a document flat discount would leave every line's BT-131 wrong.
- **Semantics.** `llx_societe_remise_except` records a commercial discount held on the third party, consumable by any invoice. A line allowance on a received supplier invoice is not a credit we hold on that supplier; creating one would put a spendable discount on the account out of nothing.
- **Round trip.** On export the module already emits a Dolibarr line discount as BG-27 (`addLineDiscount()` on `SpecifiedLineTradeSettlement`, reason code 95). Import BG-27 → `remise_percent` is the exact inverse. Moving it to a flat discount would re-emit a received BG-27 as a BG-20 — a different document from the one received.

One thing worth being explicit about: neither shape is a perfect round trip. Strictly, Dolibarr's `remise_percent` maps to BT-147/BT-148 (item price discount), not to BG-27, and this PR's charge line re-emits as an extra invoice line rather than as BG-28. What both preserve is the amount, the VAT rate and the VAT base, which is what BR-CO-10/13 and the BR-*-08 family are checking. If you would rather revisit the allowance mapping itself, that is a separate change on existing behaviour and affects every document already imported — happy to open it as its own issue.

PR #741, opened 2026-09-03
FIX: #739 [einvoicing] A received invoice stops landing on a thirdparty that only shares the email address

### LINKED ISSUE #739: [EInvoicing] Supplier invoice matched to wrong third party based on email instead of legal identifier
### Bug

Module
einvoicing

Dolibarr Version
23.0.3

Bug
When importing a supplier e-invoice in CII format from a superPDP, Dolibarr may attach the invoice to an existing third party only because the email address matches, even when the legal identifiers do not match.

In my test case:

Auto creation of missing third parties is enabled.
The supplier in the CII file has:
Company name: TEST SUPPLIER
SIREN / legal identifier: 123456789
VAT ID: FR00123456789
Email: accounting@example.com
No third party with SIREN 123456789 exists in Dolibarr.
An unrelated test third party already exists in Dolibarr:
Name: TEST THIRD PARTY
No SIREN
No VAT ID
Same email: accounting@example.com

When the supplier invoice is synchronized from the PDP, Dolibarr creates the draft supplier invoice on TEST THIRD PARTY instead of creating a new supplier third party for TEST SUPPLIER.

Expected behavior

Matching should primarily rely on structured legal identifiers such as SIREN / legal organization identifier and VAT ID.

If no existing third party matches the supplier legal identifiers and automatic creation of missing third parties is enabled, Dolibarr should create a new supplier third party.

An email address alone should not be sufficient to associate an electronic supplier invoice with an existing third party when the legal identifier contained in the CII does not match.

Actual behavior

The invoice is attached to an unrelated third party apparently because the email address matches.

Steps to reproduce

Enable automatic creation of missing third parties in the EInvoicing module.
Create a third party named TEST THIRD PARTY.
Do not set any SIREN or VAT ID on this third party.
Set its email to accounting@example.com.
Import/synchronize a supplier invoice containing:
Supplier name TEST SUPPLIER
Legal identifier 123456789
VAT ID FR00123456789
Email accounting@example.com
Synchronize the invoice from the PDP.

Result

The draft supplier invoice is created on TEST THIRD PARTY.

Expected result

A new third party TEST SUPPLIER should be created and the invoice should be assigned to it.

Additional information

The incoming flow is successfully received as:

flowType: SupplierInvoice
flowDirection: In
flowSyntax: CII
acknowledgement.status: Ok

So the issue appears to be related to third-party matching during supplier invoice import rather than PDP transport or CII validation.

### Module

einvoicing

### Dolibarr Version

23.0.3

### Environment PHP

_No response_

### Environment Database

_No response_

Steps to reproduce:

1. Enable automatic creation of missing third parties in the EInvoicing module.
2. Create an unrelated test third party in Dolibarr, for example:
   - Name: TEST THIRD PARTY
   - No SIREN / legal identifier
   - No VAT ID
   - Email: accounting@example.com
3. Make sure that no third party exists in Dolibarr with the supplier legal identifier used in the incoming e-invoice.
4. Receive and synchronize a supplier e-invoice through the PDP with:
   - Supplier name: TEST SUPPLIER
   - SIREN / legal identifier: 123456789
   - VAT ID: FR00123456789
   - Email: accounting@example.com
5. The incoming flow is successfully received as a SupplierInvoice in CII format.
6. Dolibarr creates the supplier invoice draft, but assigns it to TEST THIRD PARTY.

Actual behavior:

Dolibarr matches the incoming supplier invoice with an unrelated existing third party apparently because the email address is the same.

The existing third party has:
- a different name,
- no SIREN,
- no VAT ID.

The supplier contained in the CII file has a valid legal identifier and VAT ID.

Expected behavior:

Dolibarr should first match suppliers using structured legal identifiers such as SIREN / legal organization identifier and VAT ID.

If no existing third party matches these legal identifiers, and automatic creation of missing third parties is enabled, Dolibarr should create a new supplier third party.

An email address alone should not be sufficient to associate an incoming electronic supplier invoice with an existing third party when the legal identifiers do not match.

### Attached files

_No response_

daGrumpf-bxp:
Fixes #739.

Reproducing the report turned up **two** defects on the same lookup, both booking a received
supplier invoice on a company that did not send it. Fixing only the reported one would leave 18
and 19 broken in a worse way, so both are here — one commit each.

### 1. An email address is used as an identity

`_syncOrCreateThirdpartyFromEInvoiceSeller()` looks the seller up by its structured identifier
(SIREN/SIRET at step 1, VAT at step 2). When those find nothing, step 3 still handed the email
address to `findNearest()`, whose email branch answers any third party carrying it. The reporter's
`TEST THIRD PARTY`, with no SIREN and no VAT, thus collected an invoice issued by SIREN 123456789.

An email address is not a legal identity: several third parties can share one, and a third party
can carry the address of someone who is not the seller. So it is **no longer a match criterion
unless an administrator asks for it**, through a new option of the setup page —
`EINVOICING_THIRDPARTIES_MATCH_ON_EMAIL`, off by default, next to the other options that decide
what a received document may create or change. Turning it on restores the previous behaviour
whole, which is what an installation whose third parties carry no legal identifier needs; the
help text says plainly what it costs.

### 2. `findNearest()` answers `1`, and `1` is read as a third party id

`Societe::fetch()` answers **1** on a match up to Dolibarr 19, and **the row id** from 20 on. The
compat `findNearest()` in `einvoicing/lib/einvoicing.lib.php` forwarded that answer raw, and the
caller books the invoice on the returned value — so on 18 and 19 every received invoice whose
seller was not found by identifier landed on **the third party of id 1**, whatever company that is.
It now answers `$tmpthirdparty->id`, negative codes untouched.

`createSupplierInvoiceFromSource()` on a CII whose seller is `T739 SUPPLIER` / SIREN 739000001 /
VAT FR99739000001, against a pre-existing supplier with no identifier and the same email, on
18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4 and 24.0.0.

| core | before | after, option off (default) | after, option on |
|---|---|---|---|
| 18, 19 | invoice on third party **id 1** (unrelated) | new supplier created | invoice on the **email twin** |
| 20, 21 | new supplier created | new supplier created | new supplier created |
| 22, 23, 24 | invoice on the **email twin** | new supplier created | invoice on the **email twin** |

The third column is the point of the option: it is the whole of the previous behaviour, back on
demand. 20 and 21 do not move in any column because their `Societe::findNearest()` is still the
stub that only does the exact all-fields match; 22 implemented the step-by-step search, email
included. Controls:

- **document with no identifier at all** → new supplier created with the option off, matched by
  email with the option on, and on 18/19 it now reaches the intended third party instead of id 1;
- **creation disabled** (`EINVOICING_THIRDPARTIES_AUTO_GENERATION` off) → the import is refused
  with the message naming the supplier on the seven cores, where it used to book on the wrong
  company silently;
- **the setup page over HTTP on 23.0.4** → the option renders with its help text, its toggle
  writes then removes the constant, and an import run right after the toggle follows what the page
  was set to;
- PHPUnit 147/147 on the seven versions, before as after;
- 112 generations (`CII` + `FACTURX`, 8 instances) clean before as after — no emission path is
  touched;
- two full lifecycles from a fresh invoice through the sandbox access point (to 18 and to 23),
  acknowledged both times.


daGrumpf-bxp:
Two points left open by this PR, for the record. Both are of the same family as the defect fixed
here, neither is a regression introduced by it, and I would rather have them decided than slipped
in silently.

### 1. The seller name is also passed as `ref_ext`, and that stage is now the deciding one

The call passes the seller name three times over — as `$ref`, as `$ref_ext` and as `$ref_alias`:

```php
$thirdparty->findNearest(
    0,
    $sellerInfo['sellername'] ?? '',   // $ref
    $sellerInfo['sellername'] ?? '',   // $ref_ext
    '', '', '', '', '', '', '',
    $emailForLooseMatch,               // $email
    $sellerInfo['sellername'] ?? ''    // $ref_alias
);
```

The last stage of `findNearest()` is an OR over the three (`societe.class.php` around l. 2363 on
23.0.4, and the compat copy in `einvoicing/lib/einvoicing.lib.php`):

```sql
... AND ( s.nom = 'NAME' OR s.name_alias = 'NAME' OR s.ref_ext = 'NAME' )
```

With the email now dropped for an identified seller, that OR becomes the **deciding** stage of the
cascade: the exact all-fields match needs `nom` AND `name_alias` AND `ref_ext` to all equal the
name, barcode is empty, the profid stage gets nothing, and the email stage is what this PR closes.
So a third party whose `ref_ext` happens to equal the seller's name now collects the invoice.
`ref_ext` is a free field, not shown on the card, typically written by whatever import created the
record — it is a weaker identity claim than the email address this PR just refused. Dropping the
`$ref_ext` argument would leave name and alias, which is what the loose match is meant to be.

### 2. Read to the letter, #739 asks for no loose match at all here

The report states: *"If no existing third party matches the supplier legal identifiers and
automatic creation of missing third parties is enabled, Dolibarr should create a new supplier third
party."* This PR keeps the match by name for an identified seller whose identifier matched nothing.

That is defensible — suppliers entered by hand carry no SIREN, and refusing the name match would
create a duplicate for each of them. But it is a deliberate divergence from what the reporter asked
for, and a homonym with no identifier still collects the invoice. Worth a word from
@olivierMBTS on whether the name match is wanted in that case, and from @eldy on the scope, rather
than my deciding it.

Both are outside the scope of the two commits here, and I am happy to open a follow-up for
whichever of the two you want.

### Not an issue, checked

- `EINVOICING_THIRDPARTIES_COMPLETE_INFO` writing the real supplier's SIREN onto the wrongly matched
  third party — which would make the mis-attachment self-confirming, the next flow matching
  "cleanly" by SIREN at step 1 — is closed by construction: the guard counts as an identification
  only the schemes step 1 actually searches
  (`!empty($this->_mapGlobalIdSchemeToIdprof($idScheme, $sellerCountryCode))`), which are exactly
  the ones the update step writes. No gap between the two.
- `$is_client` / `$is_supplier` left at 0, so the loose match can land on a plain prospect: that one
  is deliberate and documented in the code ("ex: a prospect or customer receiving its first supplier
  invoice"), so it must stay as is.

daGrumpf-bxp:
Second commit reworked: the email is no longer dropped by a rule the code decides on its own, it
is dropped unless an administrator turns it back on.

The first version guessed — it kept the email when the document carried no structured identifier
and dropped it when it did. That guess was defensible but invisible: nothing in the interface said
which of the two an installation was getting, and no one could ask for the other. It is now a
plain option of the setup page, `EINVOICING_THIRDPARTIES_MATCH_ON_EMAIL`, off by default, in the
"Synchronization of purchase invoices" block beside the options that decide what a received
document may create or change. Off, the email is never a match criterion; on, it is one exactly as
before. The measurements are in the description, the option-on column included.

Two consequences worth stating rather than leaving to be found:

- **an installation whose suppliers carry no legal identifier now creates a duplicate** where it
  used to match by email, until an administrator ticks the option. That is the price of a default
  that never mis-attributes, and it is one click to undo, which the previous version's guess was
  not;
- **with the option on, `EINVOICING_THIRDPARTIES_COMPLETE_INFO` makes a mis-attachment
  self-confirming**: I measured the seller's SIREN and VAT number being written onto the third
  party matched by email, so the next flow matches "cleanly" by identifier at step 1 and the
  option can no longer be blamed. The first version's guard closed that by construction; the
  option cannot, so the help text warns instead. If you would rather the two options refused to be
  on together, say so and I will add the check.

The point 1 of my previous comment is unchanged and, if anything, sharper: with the email gone by
default, the `ref_ext` leg of the last `findNearest()` stage becomes the deciding one, and
`ref_ext` is a weaker identity claim than the email address this PR just refused. Point 2 — whether
a name match is wanted at all for a seller that carries an identifier — is still a call for
@olivierMBTS and @eldy, not mine.
---

**Added: the control that says the default passes for the right reason.** Same document (SIREN
739000001, VAT FR99739000001), option off, three shapes of the pre-existing third party, all three
carrying the seller's email address:

| pre-existing third party | 18 | 19 | 20 | 21 | 22 | 23 | 24 |
| the **email only**, a different name | new | new | new | new | new | new | new |
| email **+ the seller's own name** | it | it | new | new | it | it | it |
| email **+ `ref_ext` = the seller's name** | it | it | new | new | it | it | it |

The first row is the fix. The two others are positive controls: the loose match *can* reach that
third party when a criterion it still trusts answers, so the first row is the email being refused
and not a bench that never matches anything. 20 and 21 never move — the stub again.

The second control also settles point 1 of my previous comment, which I had only read in the code:
**`ref_ext` really does catch, on five cores**. A third party whose `ref_ext` happens to equal the
seller's name collects the invoice — a free field, not shown on the card, usually written by
whatever import created the record, and a weaker identity claim than the email address this PR just
refused. On 18/19/22/24 `EINVOICING_THIRDPARTIES_COMPLETE_INFO` then writes the real seller's SIREN
and VAT number onto it, so the next flow matches "cleanly" at step 1. Dropping the `$ref_ext`
argument from the call would leave name and alias, which is what a loose match is meant to be — say
the word and it is a one-line follow-up.

olivierMBTS:
Thanks for the fix and for reproducing the issue.

Regarding the remaining question: when the incoming invoice contains a structured legal identifier (SIREN/SIRET/VAT) and no existing third party matches that identifier, I would prefer Dolibarr to create a new supplier rather than fall back to a loose name match.

In my opinion, a company name is not reliable enough as an identity criterion for e-invoicing. Different legal entities may have very similar or identical names, while the structured legal identifier is authoritative.

A name-based fallback could still make sense when the incoming document does not contain any structured legal identifier, but when one is present and does not match, I think automatic creation is safer.

daGrumpf-bxp:
Point 1 is now a PR of its own: #743 drops the `$ref_ext` argument, leaving name and commercial
alias, which is what a loose match is meant to be.

It sits on `main` and does not wait for this one — the two touch the same call but different
arguments. I tried the merge of the two locally: no conflict, and `php -l` clean on the result
(the comments are deliberately placed apart so they do not collide).

Measured on the seven cores, with a pre-existing third party carrying no email address and the
seller name in one single field: `ref_ext` = the name stops collecting the invoice (a new supplier
is created, 7/7), while `nom` = the name and `name_alias` = the name still do, unchanged. Point 2
— whether a name match is wanted at all for a seller that carries an identifier — is still yours
to call, @olivierMBTS and @eldy.

olivierMBTS:
I think there may be an even safer approach for point 2.

If the incoming supplier carries a structured legal identifier (SIREN/SIRET/VAT) and no exact legal identifier match is found, Dolibarr could avoid automatically assigning the invoice based only on name/name_alias.

Instead, if a loose name match is found, Dolibarr could propose that third party to the user and offer two choices:

- use the suggested existing third party;
- create a new third party from the e-invoice data.

This would keep the benefit of name matching for avoiding duplicates, while preventing an incorrect automatic assignment.

In my opinion, this would be the safest behavior for e-invoicing.

olivierMBTS:
I think there may be an even safer approach for point 2.

If the incoming supplier carries a structured legal identifier (SIREN/SIRET/VAT) and no exact legal identifier match is found, Dolibarr could avoid automatically assigning the invoice based only on name/name_alias.

Instead, if a loose name match is found, Dolibarr could propose that third party to the user and offer two choices:

- use the suggested existing third party;
- create a new third party from the e-invoice data.

This would keep the benefit of name matching for avoiding duplicates, while preventing an incorrect automatic assignment.

In my opinion, this would be the safest behavior for e-invoicing.

There is also an important consequence of a wrong automatic match in the current workflow:

Once the incoming e-invoice was created as a draft supplier invoice on the wrong third party, I was not able to simply correct the supplier.

I also could not delete the draft invoice, because Dolibarr requires the received e-invoice to be approved or rejected first.

And when I tried to reject it, the CDAR status could not be sent to the PDP and Dolibarr returned an error about an invalid electronic address (MDT-73).

So a wrong automatic third-party match can leave the user with a received e-invoice attached to the wrong supplier and no simple way to correct or remove it.

This is another reason why I think a user confirmation step would be safer when only a loose match is available.

eldy:
> I think there may be an even safer approach for point 2.
> 
> If the incoming supplier carries a structured legal identifier (SIREN/SIRET/VAT) and no exact legal identifier match is found, Dolibarr could avoid automatically assigning the invoice based only on name/name_alias.
> 
> Instead, if a loose name match is found, Dolibarr could propose that third party to the user and offer two choices:
> 
> * use the suggested existing third party;
> * create a new third party from the e-invoice data.
> 
> This would keep the benefit of name matching for avoiding duplicates, while preventing an incorrect automatic assignment.
> 
> In my opinion, this would be the safest behavior for e-invoicing.

To be honnest i would even never allow the non ID matching.

olivierMBTS:
One additional workflow issue I noticed during this test:

When an incoming e-invoice is imported as a draft supplier invoice, I think Dolibarr should allow the local draft to be deleted as long as it has not been validated yet, without forcing the user to approve or reject the electronic invoice.

It would also be very useful to have a "reprocess / re-import incoming e-invoice" action.

In my test case, once the invoice had been imported on the wrong third party, I could not:
- change the supplier;
- delete the draft;
- re-import the same incoming flow after correcting the third-party data.

This makes recovery from an import or matching issue difficult.

Separating the local Dolibarr draft lifecycle from the electronic invoice lifecycle would make the workflow much safer.

daGrumpf-bxp:
@eldy — done, the PR is now your ruling and nothing else: no non-ID matching at all. The
`findNearest()` step is removed, and with it the compat copy in `einvoicing/lib/einvoicing.lib.php`,
whose only caller it was. The option I had added to make the email match an administrator's choice
is gone too — it was a way of keeping something you say should not be kept.

Measured on the seven cores (18.0.10 → 24.0.0), a pre-existing third party carrying the seller's
email, `nom`, `name_alias` or `ref_ext` — one single field each time — used to collect the invoice
on 18/19/22/23/24 and now gets a new supplier created instead, 7/7 in all four shapes. Creation
disabled: refused with the message naming the seller, 7/7.

@olivierMBTS — the one thing your reading kept and this one does not: a document that carries **no
structured identifier at all**. Measured, importing the same such document twice: 22/23/24 used to
recognize the supplier created by the first import, and now create a second one, every time. On a
CTC-FR flow that should not happen — BT-29/BT-30/BT-31 are there — but it is the price, so it is
stated rather than left to be found. One condition brings it back if the board wants it.

Your other point is a real one and outside this PR: once a received e-invoice has been booked on
the wrong third party, the draft can be neither re-assigned nor deleted (approval or rejection
required first, and the rejection fails on MDT-73). That is a trap of its own and deserves its own
issue — I can open it with the reproduction if you have not already.

#743 is closed: it dropped the `$ref_ext` argument of a call this PR deletes.

daGrumpf-bxp:
Apologies for the noise on this thread — I edited the title and description of this PR **after it
was merged**, believing it was still open. Both are restored to what was actually merged. The
branch is back to its merged head as well.

What happened: your ruling landed at 16:35:43, the merge at 16:35:55, and I was rewriting the
branch to implement the ruling at the same moment. Nothing of the merged content was affected — the
merge commit and `main` never moved — but the two comments above referring to "this PR" removing
the loose match describe code that is **not** in this PR and not in `main`.

The ruling itself is not dropped: it goes to a new PR against current `main`, which now carries
both this one and #743. I will link it here.

daGrumpf-bxp:
@eldy — your ruling is #746, on top of current `main`: the `findNearest()` step goes whole, so the
name and the commercial alias stop attaching a received document, as the email address and
`ref_ext` already do since this PR and #743. The compat `findNearest()` of the module lib goes with
it (that step was its only caller), and so does the `EINVOICING_THIRDPARTIES_MATCH_ON_EMAIL` option
this PR added — it offered a match you say should not exist.

Measured on the seven cores: `nom` and `name_alias` still hand the invoice to a third party that
merely carries them, on 18/19/22/23/24, and stop doing so. The one thing it costs is in the
description of #746 and is @olivierMBTS's case: a document carrying **no** structured identifier at
all now creates a supplier at every import instead of recognizing the one it created before. One
condition brings that back if you want it.

olivierMBTS:
Thanks, that sounds good.

I have not opened a separate issue yet for the draft supplier invoice being stuck on the wrong third party and the MDT-73 rejection error.

Please feel free to open it with the reproduction steps from this thread.

I would also add that, ideally, an imported supplier invoice that is still only a local draft should be deletable without forcing an approval/rejection status, and there should be a way to reprocess/re-import the incoming e-invoice after fixing the matching data.

daGrumpf-bxp:
@olivierMBTS — the MDT-73 half of what you reported is #747.

I reproduced it on a really received invoice against the sandbox access point: with no routing
identifier recorded for the vendor, a refusal (210) comes back as
`L'adresse électronique (MDT-73) est invalide` and nothing else — the message names neither the
vendor, nor the address that was refused, nor what to do. #747 makes it say both, and name the
field that fixes it (the routing identifier on the third-party card, tab E-invoicing), with a
different sentence depending on whether one is recorded and wrong, or missing altogether. The
control is in the PR: with the right address the access point accepts the CDAR and answers about
another rule, and that answer is forwarded untouched.

One thing worth knowing that came out of measuring it: the electronic address carried by the
document you receive (BT-34) is **not** necessarily the address its sender exchanges under — on my
bench the document says `000000001` where the access point knows that vendor as `315143296_2341`.
So the address really has to be recorded on the vendor card; the module can only guess it.

The other half — a draft supplier invoice that cannot be deleted without an approval or a refusal,
its third party that cannot be changed, and no way to re-import a received flow after fixing the
matching data — is a separate matter and I have not touched it. Say the word and I open the issue
with the reproduction, or open it yourself if you would rather word it.

olivierMBTS:
Yes please, I think this deserves a separate issue.

Please feel free to open it with the reproduction steps from this thread.

The main workflow issue for me is that, once a received e-invoice has been imported as a draft supplier invoice on the wrong third party, there is currently no easy recovery path:

- the supplier cannot be changed;
- the draft cannot be deleted without first approving or rejecting the electronic invoice;
- and there is no way to reprocess/re-import the same incoming flow after correcting the third-party data.

Ideally, I think a received supplier invoice that is still only a local draft should be deletable without forcing an approval/rejection status.

It would also be very useful to have a "reprocess / re-import incoming e-invoice" action after correcting the matching data.

This would avoid having to reject a valid electronic invoice and ask the supplier to issue a new one just because the first local import was attached to the wrong third party.

Thanks again for looking into this.

daGrumpf-bxp:
@olivierMBTS — opened as #755, and #757 answers it.

The trap turned out to be exactly as you described it, and the message the module answers names a way out that does not exist: an approval (205) and a refusal (210) were both sent for real to the sandbox access point and accepted, and the deletion is refused again afterwards, with that same "please approve or reject this invoice". Deleting the flow record and synchronizing again does not help either: it is refused as soon as it is not the most recent record, and when it is, the import finds the supplier invoice that is still there and links straight back to it - same wrong third party.

So #757 does the two things you asked for: the draft of a received e-invoice can be deleted (the received document itself is kept, detached, so it stays in the flow list), and an "import this document again" action re-runs the import from the access point once the third party data is fixed - it deletes the draft, asks for the flow again, and lets the matching resolve the vendor a second time. Measured on 18.0.10 through 24.0.0, and on a full round trip: putting the SIREN back on the right third party and importing again moves the invoice to it.

One thing it deliberately does not do is change the vendor of the invoice in place: the lines of an imported invoice are bound to the vendor they were resolved against (supplier references, product mapping, the purchase order BT-13 matched), so moving the third party alone would leave an invoice whose lines belong to someone else. Importing the document again redoes all of it.

PR #743, opened 2026-09-03
FIX: #739 [einvoicing] ref_ext is not an identity claim for the seller of a received document

daGrumpf-bxp:
Follow-up to #741, on the same lookup and the same issue (#739). It stands on `main` on its own
and does not need #741 to be merged first — the two touch the same call but different arguments,
and the merge of the two was tried locally: it applies clean and the result parses.

### What it does

`_syncOrCreateThirdpartyFromEInvoiceSeller()` passes the seller name of a received document
**three times over** to `findNearest()` — as `$ref`, as `$ref_ext` and as `$ref_alias`. The last
stage of that function ORs the three (`societe.class.php` around l. 2363 on 23.0.4, and the compat
copy in `einvoicing/lib/einvoicing.lib.php`):

```sql
... AND ( s.nom = 'NAME' OR s.name_alias = 'NAME' OR s.ref_ext = 'NAME' )
```

So a third party whose `ref_ext` happens to equal the seller name collects the invoice. Name and
commercial alias are fair criteria for a loose match — they are what a human reads on the card.
`ref_ext` is not: a free field, absent from the card, typically written by whatever import created
the record, and it makes no claim to **be** that company. It is no longer passed.

This matters more since #741, which stops the email address from being a criterion: `ref_ext` then
becomes the deciding leg of the cascade, and it is a weaker claim than the email address #741
refuses.

`createSupplierInvoiceFromSource()` on a CII whose seller is `T739 SUPPLIER` / SIREN 739000001 /
VAT FR99739000001, against a pre-existing supplier carrying **no email address** and the seller
name in one single field, on 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4 and 24.0.0.

| the pre-existing third party carries | before | after |
|---|---|---|
| `ref_ext` = the seller name | it collects the invoice (18, 19, 22, 23, 24) | **a new supplier is created**, 7/7 |
| `nom` = the seller name | it collects the invoice (18, 19, 22, 23, 24) | unchanged |
| `name_alias` = the seller name | it collects the invoice (18, 19, 22, 23, 24) | unchanged |

The last two lines are the point: they say the loose match still answers on the criteria it is
meant to answer on, so the first line is `ref_ext` being dropped and not a cascade that stopped
working. 20 and 21 create a new supplier in every line, before as after — their
`Societe::findNearest()` is still the stub that only does the exact all-fields match.

- PHPUnit 147/147 on the seven versions, before as after;
- 112 generations (`CII` + `FACTURX`, 8 instances) clean before as after — no emission path is
  touched;
- merged with #741 locally: no conflict, and `php -l` clean on the result.


daGrumpf-bxp:
Superseded by #741 and closed.

@eldy settled the underlying question there — *"to be honest i would even never allow the non ID
matching"* — so #741 now removes the whole `findNearest()` step instead of one of its arguments.
This PR edits a call that no longer exists after it.

The measurement stands and moved there: `ref_ext` = the seller name used to collect the invoice on
18/19/22/23/24 and no longer can. Reopen this one if the board would rather keep the loose match
and only drop `ref_ext` from it.

daGrumpf-bxp:
Correction to my comment above: this PR was **merged**, not superseded — it landed while I was
writing. Please ignore the "closed / reopen it" wording, there is nothing to reopen.

What is true: @eldy's ruling on #741 (*"i would even never allow the non ID matching"*) removes the
whole `findNearest()` step, so the call this PR fixed will go away in the follow-up PR against
current `main`. That does not undo anything here — `ref_ext` is out of `main` either way.

Issue #718, opened 2026-09-02
[einvoicing] Any import failure that is not explicitly postponed still stops the whole incoming synchronization


daGrumpf-bxp:
### Bug

#675 asked two things. The first — a message that names the problem and what to do — was answered by
#676. The second was left open in that issue, and is the one this report is about:

> **Should one document still stop the queue?** [...] That is a behaviour change on the shared loop
> of every provider, so it is your call rather than mine.

#676 answered it, but **for one case only**. It introduced a `postponeflow` flag that lets
`syncFlows()` carry on, and wired it into the single situation it was written for: a received invoice
referencing a document the buyer does not hold. Every other failure still stops the batch, and the
flag is opt-in at the deepest point of the import rather than a decision of the loop that owns the
batch.

The numbers, on `main` at a46effc2:

| | `return ['res' => -1` sites | of which carry `postponeflow` |
|---|---|---|
| `CIIProtocol.class.php` | 26 | 1 |
| `FacturXProtocol.class.php` | 16 | 1 |

Plus the `catch (Exception $e)` of the loop itself (`SuperPDPProvider.class.php:2029`,
`EsalinkPDPProvider.class.php` same shape): any PHP exception raised while importing one flow also
increments `$error`.

And `$error > 0` means the batch is over — `SuperPDPProvider.class.php:2034`, and
`EsalinkPDPProvider.class.php:1129`:

```php
if ($error > 0) {
    if (in_array($rescode, array('THIRDPARTY_NOT_FOUND','PRODUCT_NOT_FOUND'))) {
        $results_messages[] = "Aborting synchronization due to a business error. There is a manual action to do.";
    } else {
        $results_messages[] = "Aborting synchronization due to errors.";
    }
    break;
}
```

So 40 of the 42 failure returns of the import, and every unforeseen exception, still freeze the
incoming synchronization of the whole instance — for every supplier — until somebody intervenes by
hand.

### What it cost this week, measured

A concrete instance of the class, found on the bench and reproduced on the seven supported cores.

A situation invoice n°2 carries a document level allowance (BT-107, the deduction of the situations
already invoiced) — a shape **this module itself now emits**, since the fix for #674. On the
receiving side that allowance becomes a discount exception, and on Dolibarr 18 and 19 its creation
failed on a foreign key. What the receiver got:

```
ERROR_SYNCFLOW - Failed to synchronize flow <id>: Failed to create supplier invoice from E-invoice
[...]

PR #744, opened 2026-09-03
FIX #742 [einvoicing] A received Factur-X in the EXTENDED-CTC-FR profile can be imported again

### LINKED ISSUE #742: [einvoicing] A received Factur-X in the EXTENDED-CTC-FR profile cannot be imported and stops the whole incoming synchronization
A user running the module in production reported that the incoming synchronization stopped
altogether. It aborts on the third flow of the batch and never gets past it, run after run:

Image

```
Synchronisation interrompue sur le flux # 3 / 100 (Flux i_422783)
Informations techniques : Exception occurred while synchronizing flow i_422783:
Cannot determain the profile by urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr
Aborting synchronization due to errors.
```

The received document is a Factur-X PDF whose guideline (BT-24) is the French profile
`urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr`.

## Root cause

`FacturXProtocol::doCreateSupplierInvoiceFromSource()` opens the received PDF with the reader of
`horstoeko/zugferd`, which resolves the profile by matching that URN against its own table
(`ZugferdProfiles::PROFILEDEF`). That table only holds the Factur-X / ZUGFeRD profiles — MINIMUM,
BASIC WL, BASIC, EN 16931, EXTENDED, XRECHNUNG — and has **no entry for EXTENDED-CTC-FR**, so
`ZugferdProfileResolver::resolve()` throws `ZugferdUnknownProfileException`.

`einvoicing/class/protocols/FacturXProtocol.class.php:452`

```php
$document = ZugferdDocumentPdfReader::readAndGuessFromFile($tempFile);   // throws here
$embeddedXml = ZugferdDocumentPdfReaderExt::getInvoiceDocumentContentFromFile($tempFile);

$parsedHeader = [];
$parsedLines = [];
if (!getDolGlobalInt('EINVOICING_USE_EXTERNAL_FACTURX_READER')) {   // the default
    $parsedHeader = $this->parseInvoiceHeader($embeddedXml);        // parser of the module
    $parsedLines  = $this->parseInvoiceLines($embeddedXml);
} else {
    $document->getDocumentInformation(...);                        // the only user of $document
}
```

`$document` is used **only** inside the `else`, that is behind
`EINVOICING_USE_EXTERNAL_FACTURX_READER` — a constant that is defined nowhere in the module, so
that branch never runs in practice. What the default path actually uses,
`getInvoiceDocumentContentFromFile()`, only reads the PDF/A-3 attachment and never looks at the
profile: it handles EXTENDED-CTC-FR perfectly well.

So the whole import is stopped by a call whose result is thrown away.

The same gap of the library is already documented on the generation side, and handled there:
`CtcFrPdfMerger` exists precisely because `ZugferdDocumentPdfMerger` cannot resolve
EXTENDED-CTC-FR, and `FacturxTcpdfMerger::resolveProfileParameters()` wraps the resolver in a
`try/catch` with a fallback on the EXTENDED parameters. The reading side never got the same
treatment.

Worth noting: EXTENDED-CTC-FR is the profile **this module itself emits** —
`CIIProtocol` builds it for `EINVOICING_XML_PROFILE=EXTENDEDFR`, `XmlPatcher::URN_EXTENDED_CTC_FR`
forces it, and `AbstractPDPProvider::FLOW_PROFILE_BY_GUIDELINE` declares `Extended-CTC-FR` to the
platforms for it (#395). The module produces a profile it cannot read back.

## Why the whole batch dies with it

The exception is caught in `syncFlow()` and rethrown, then caught in `syncFlows()`, which does
`$error++` and `break`. Nothing was written to `llx_einvoicing_document` for that flow, so it is
not among the "already processed" ones on the next run: it comes back at the same position and
blocks again. The block is permanent, and every flow queued behind it — 97 of them in the report
above — is never read.

The mechanism meant for this, `postponeflow`, is opt-in on a few `-1` returns and is never reached
by an exception. That part is #718.

Nothing is lost: no invoice is consumed or deleted on the platform side, and the backlog is caught
up as soon as the flow can be read.

## Reproduction

Fully reproducible with a document the module generates itself:

1. `EINVOICING_PROTOCOL=FACTURX`, `EINVOICING_XML_PROFILE=EXTENDEDFR`
2. validate an invoice → `<ref>_facturx.pdf`, whose BT-24 is
   `urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr`
3. feed that PDF to `FacturXProtocol::createSupplierInvoiceFromSource()`

```
=== dd23 (Dolibarr 23.0.4) ===
  file    : fx-ctcfr.pdf (108521 bytes)
  import  : EXCEPTION horstoeko\zugferd\exception\ZugferdUnknownProfileException:
            Cannot determain the profile by urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr
```

Same result on **18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4 and 24.0.0** — 7 of 7 blocked.
A Factur-X in the EN 16931 profile imports normally on all of them, which is why this went
unnoticed: only the French profile is affected.

## Scope

Any inbound Factur-X in the French profile. There is no way around it on the Esalink provider,
whose `SupplierInvoice` case only asks for the `Converted` document, which is a Factur-X PDF, so
every received invoice goes down this path. On SuperPDP it depends on the syntax the access point
is configured to convert to — and a customer who deliberately configured `Converted = Factur-X`,
which is a legitimate setup, has all of their incoming invoices blocked.

PR follows.

daGrumpf-bxp:
Fixes #742.

A user in production had their incoming synchronization stopped dead: it aborted on the third flow
of the batch and never got past it, run after run.

```
Synchronisation interrompue sur le flux # 3 / 100 (Flux i_422783)
Exception occurred while synchronizing flow i_422783:
Cannot determain the profile by urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr
Aborting synchronization due to errors.
```

## What was happening

`FacturXProtocol::doCreateSupplierInvoiceFromSource()` opened the received PDF with the reader of
`horstoeko/zugferd`, which resolves the profile by matching the guideline URN (BT-24) against its
own `ZugferdProfiles::PROFILEDEF`. That table holds the Factur-X / ZUGFeRD profiles only — MINIMUM,
BASIC WL, BASIC, EN 16931, EXTENDED, XRECHNUNG — and has no entry for EXTENDED-CTC-FR, so
`ZugferdProfileResolver::resolve()` threw `ZugferdUnknownProfileException`.

The exception is rethrown by `syncFlow()` and caught in `syncFlows()`, which does `$error++` and
`break`. Nothing had been written to `llx_einvoicing_document` for that flow, so it was not among
the already-processed ones on the next run: it came back at the same position and blocked again,
and the 97 invoices queued behind it were never read. Nothing is lost — the backlog is caught up
as soon as the flow can be read — but the queue does not move on its own.

EXTENDED-CTC-FR is the profile **this module itself emits**: `CIIProtocol` builds it for
`EINVOICING_XML_PROFILE=EXTENDEDFR`, `XmlPatcher::URN_EXTENDED_CTC_FR` forces it, and
`AbstractPDPProvider::FLOW_PROFILE_BY_GUIDELINE` declares `Extended-CTC-FR` to the platforms for it
(#395). The module produced a profile it could not read back.

## The fix

The reader that throws is used **only** inside the `else`, that is behind
`EINVOICING_USE_EXTERNAL_FACTURX_READER` — a constant defined nowhere in the module, so that branch
never runs in practice. The default path parses the embedded CII with the parser of the module, and
the extraction it relies on, `getInvoiceDocumentContentFromFile()`, only reads the PDF/A-3
attachment and never looks at the profile: EXTENDED-CTC-FR goes through it unharmed.

So the whole import was stopped by a call whose result was thrown away. It is now built inside the
branch that uses it. One statement moved, no behaviour added.

The same gap of the library is already handled on the generation side — `CtcFrPdfMerger` exists
precisely because `ZugferdDocumentPdfMerger` cannot resolve EXTENDED-CTC-FR, and
`FacturxTcpdfMerger::resolveProfileParameters()` wraps the resolver in a `try/catch` with a fallback
on the EXTENDED parameters. The reading side had never got the same treatment.

Fixture generated by `main` itself: `EINVOICING_PROTOCOL=FACTURX`,
`EINVOICING_XML_PROFILE=EXTENDEDFR`, invoice validated → `<ref>_facturx.pdf` whose BT-24 is
`urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr`, then handed to
`createSupplierInvoiceFromSource()`. Every imported invoice is read back **from a second process**,
the control that catches a transaction left open.

| Dolibarr | main (before) | this branch |
|---|---|---|
| 18.0.10 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |
| 19.0.4 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |
| 20.0.4 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |
| 21.0.4 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |
| 22.0.5 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |
| 23.0.4 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |
| 24.0.0 | `ZugferdUnknownProfileException` | imported, 1 line, 120.00 TTC |

Controls, all on the same rebased base:

- **No regression on the profiles that already worked**: a Factur-X in EN 16931 imports identically
  before and after (same invoice, same line, same total, read back from the other process).
- **The moved statement is really reached on its branch**: with
  `EINVOICING_USE_EXTERNAL_FACTURX_READER=1`, the EXTENDED-CTC-FR document throws again — that
  branch is where the reader now lives — while the EN 16931 one imports through the duplicate
  parser. The opt-in path is untouched, not dead.
- PHPUnit of the module, file by file (never `AllTests.php`), on 18.0.10 and 23.0.4: **21 OK / 0 KO**,
  same counts as on `main`.
- `dolics` (phpcs 3.x + the ruleset of the repo) clean, `php -l` clean.

## Scope

Any inbound Factur-X in the French profile. There is no setup that avoids it on the Esalink
provider, whose `SupplierInvoice` case only asks for the `Converted` document, which is a Factur-X
PDF. On SuperPDP it depends on the syntax the access point converts to, and a customer who
deliberately configured `Converted = Factur-X` — a legitimate setup — had every incoming invoice
blocked.

Not addressed here, and left to #718: an exception during a flow still aborts the whole batch
instead of postponing that one flow.


daGrumpf-bxp:
The failing `phan` check is not from this branch: it comes from `main` itself. `PhanUndeclaredClassMethod: Call to method switchEntity from undeclared class \ActionsMulticompany` in `actions_einvoicing.class.php`, a file this PR does not touch — the run of `main` at 5e6b128 ("Try a native swith of entity") and at e72f02e (the base of this branch) both fail on the same line, while 0c9f61b before them was green.

`build`, `Factur-X container is PDF/A-3` and `assign-and-label` are green here.

Issue #309, opened 2026-06-29
[einvoice] Warn when the supplier name on an inbound e-invoice does not match the linked third party


ABISEsalink:
# [einvoice] Warn when the supplier name on an inbound e-invoice does not match the linked third party

## Context

French e-invoicing reform (*réforme de la facturation électronique* / RFE).

When an inbound supplier invoice (Factur-X / CII) is processed, the `einvoice` module resolves the supplier through its **electronic billing address** (*adresse de facturation électronique*, routing identifier — `schemeID="0225"`) and/or its **legal identifier** (SIREN, `schemeID="0002"`), then links it to an existing Dolibarr third party (`societe`).

Identification and routing therefore rely on a **structured identifier**, never on the company name. As a consequence, the third party can be correctly resolved even when the human-readable name carried by the invoice differs from the name stored in Dolibarr — and right now nothing tells the user about that discrepancy.

## User story

```gherkin
As a customer
When I receive a supplier invoice
And the electronic billing address of the supplier is known and linked to a third party
But the name of the supplier in the invoice does not match the name known in Dolibarr
Then I should get a warning
```

## Steps to reproduce

1. Have a Dolibarr third party linked to an electronic billing address (`0225`) / SIREN (`0002`).
2. Receive a supplier invoice (CII / Factur-X) whose seller identifier matches that third party.
3. Make the invoice's `SellerTradeParty/Name` differ from the third party's name in Dolibarr.

## Current behaviour

The invoice is silently attached to the matched third party. No discrepancy is reported, even when the names are obviously different.

Example (see attached screenshot):

| Source | Identifier (`0002` / `0225`) | Name |
|---|---|---|
| Dolibarr third party | `123456125` | `Societe MAUVAIS NOM` |
| Inbound CII invoice | `123456125` | `Societe BIDULE CHOUETTE` |

<img width="911" height="914" alt="Image" src="https://github.com/user-attachments/assets/fb794516-3208-4e3c-92ec-8891271f7a82" />

Relevant XML excerpt:

```xml
<ram:SellerTradeParty>
  <ram:ID>123456125</ram:ID>
  <ram:GlobalID schemeID="0225">123456125</ram:GlobalID>
  <ram:Name>Societe BIDULE CHOUETTE</ram:Name>                       <!-- BT-27 Seller name -->
  <ram:SpecifiedLegalOrganization>
[...]

PR #746, opened 2026-09-03
FIX: #739 [einvoicing] A structured legal identifier attaches a received document to a third party, a name only on request

daGrumpf-bxp:
Applies what @eldy settled on #741 — *"to be honest i would even never allow the non ID matching"* —
on top of current `main`, and keeps step 3 behind a hidden option as @eldy asked in the review below.

### Where the lookup stands today

`_syncOrCreateThirdpartyFromEInvoiceSeller()` looks the seller up by its structured identifier
(SIREN/SIRET at step 1, VAT at step 2) and, when those find nothing, falls back on `findNearest()`.
That call used to pass four descriptive fields; #741 took the **email address** out of it and #743
took **`ref_ext`** out. The **name** and the **commercial alias** are still in, and the last stage of
`findNearest()` ORs them, so a third party carrying either still collects the invoice.

A name is not an identity either: distinct legal entities collide on it, and under EN 16931 it is a
descriptive field of the document (BT-27, BT-28), never a routing one. So the loose match is **off by
default**: a received document is attached to a third party by a structured legal identifier, and a
seller whose identifier matches nothing is unknown — it is created, or the document is refused when
the automatic creation of third parties is disabled.

### The hidden option, as asked in review

The step itself is kept. It is now conditioned by **`EINVOICING_THIRDPARTIES_MATCH_ON_NAME`**, which
**no setup page offers**: the specific needs that rely on a name fallback keep it, an expert enables
it knowingly, and no end user meets it. Same idiom as `EINVOICING_TOLERANT_SUPPLIER_REF_MATCH`, and
the reason is written where the step is.

`EINVOICING_THIRDPARTIES_MATCH_ON_EMAIL`, added by #741, keeps its exact meaning **inside** that step
— the email is added to the criteria only when it is on — and follows it out of the setup page. It
has to: it now lives *inside* the hidden condition, so on its own it decides nothing — measured, with
it on and `MATCH_ON_NAME` absent a new supplier is created 7/7. A switch in the setup page that does
nothing until an invisible constant is set is exactly the bad use this review is closing, so the two
options are hidden together, and its two `en_US` keys go with the setup entry that was their only
reader (same shape as `EINVOICING_TOLERANT_SUPPLIER_REF_MATCH`, which carries no lang key either).

`$matchedByStructuredIdentifier` is unchanged, and stays false on a step-3 match: it is the condition
of the non-blocking name-mismatch warning of #309, which is exactly what such a match calls for.

`createSupplierInvoiceFromSource()` on a CII whose seller is `T739 SUPPLIER` / SIREN 739000001 / VAT
FR99739000001, against a pre-existing supplier carrying the seller's name, its commercial alias or
its email address in **one single field**, on 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4 and
24.0.0.

| the pre-existing third party carries | option absent | option set |
|---|---|---|
| `nom` = the seller name | new supplier created, 7/7 | it collects the invoice again (18, 19, 22, 23, 24) |
| `name_alias` = the seller name | new supplier created, 7/7 | it collects the invoice again (18, 19, 22, 23, 24) |
| the same **email address**, `MATCH_ON_EMAIL` on | new supplier created, 7/7 | it collects the invoice again (18, 19, 22, 23, 24) |
| the same **email address**, `MATCH_ON_EMAIL` off | new supplier created, 7/7 | new supplier created, 7/7 |

20 and 21 create a new supplier in every line, before as after: their `Societe::findNearest()` is
still the stub that only does the exact all-fields match. With
`EINVOICING_THIRDPARTIES_AUTO_GENERATION` off, the import is refused with the message naming the
seller, 7/7, whatever the option.

Importing the same document twice in a row, to check this does not become a duplicate factory:

| document | second import, option absent | second import, option set |
|---|---|---|
| carries an identifier | finds the supplier the first import created, 7/7 | finds it too, 7/7 |
| carries **no identifier at all** | a second supplier is created, 7/7 | finds the first one again (18, 19, 22, 23, 24) |

The second line is the whole of what the default costs, and it is the case @olivierMBTS wanted a name
fallback for — that is now what the hidden option buys back, measured rather than assumed.

### The rest of the batch

PHPUnit on the seven versions: **140 OK / 7 KO**, the same count and the same file as current `main`
run side by side (`EInvoicingSamplesTest`, a VAT rate that comes out `0.00` instead of `20.00` since
55d35ce — unrelated to this PR, and reproduced on pure `main`). 112 generations (`CII` + `FACTURX`, 8
instances) clean — no emission path is touched.

⚠️ The sandbox access point delivered nothing during this session: two lifecycles could not be
closed, and a control cycle run on pure `main` did not close either, so it is not this branch. I will
close a full cycle and report here as soon as the access point routes again.

Fixes the remaining part of #739.


einvoicing/class/protocols/CommonProtocol.class.php
```diff
@@ -574,69 +581,6 @@ private function _syncOrCreateThirdpartyFromEInvoiceSeller($sellerInfo, $priorit
 			}
 		}
 
-		// Step 3: If not found, try to find by findNearest function
```
eldy, line 577: I would keep this code available, however, i would just add a condition on a hidden option and i woul dnot show this option, so:
- We do not loose the Step 3 that may be usefull for some specific need.
- We do not  show end user this feature so no risk of a bad use. Only expert playing with hidden option can enable it.

daGrumpf-bxp: Done, that is what the branch does now: step 3 is back, unchanged, conditioned by the hidden option `EINVOICING_THIRDPARTIES_MATCH_ON_NAME` — no entry in the setup page, same idiom as `EINVOICING_TOLERANT_SUPPLIER_REF_MATCH`, and the reason is written where the step is. Default installs never take that step; an expert who needs it enables the constant knowingly.

`EINVOICING_THIRDPARTIES_MATCH_ON_EMAIL` (added by #741) keeps its exact meaning *inside* that step, and is hidden with it: it now sits **within** the hidden condition, so on its own it decides nothing — with it on and `MATCH_ON_NAME` absent, a new supplier is created 7/7. A visible switch that does nothing until an invisible constant is set would be the very bad use you are closing here, so both options are hidden together, and its two `en_US` keys leave with the setup entry that was their only reader (`EINVOICING_TOLERANT_SUPPLIER_REF_MATCH` carries none either).

Measured on a real import on 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4 and 24.0.0, against a supplier carrying the seller name, its commercial alias or its email in one single field:

- option absent: a new supplier is created in every shape, 7/7 — including with `MATCH_ON_EMAIL` on;
- option set: the existing third party collects the invoice again on 18, 19, 22, 23 and 24, by name and by alias, and by email when the email option is on too (20 and 21 never did: their `Societe::findNearest()` is the stub that only does the exact all-fields match).

And the case the fallback is really for — a document carrying no structured identifier at all, imported twice: option absent it creates a second supplier, option set it finds the first one again.

PR #750, opened 2026-09-03
CRITICAL FIX #749 [einvoicing] Since #709 every invoice carrying VAT is refused by the platform, and a situation invoice claims more than Dolibarr bills

### LINKED ISSUE #749: CRITICAL [einvoicing] Since #709 every invoice carrying VAT is refused by the platform, and a situation invoice claims more than Dolibarr bills
### Bug

The VAT breakdown of every generated CII carries a rate of `0.00`, and a situation invoice states an
amount **larger than the one Dolibarr bills**. Both come from the same cause, introduced by the merge
of #709 (`55d35ce`, 3 Sep).

`EInvoicingSamplesTest` catches the first one and is red on `main`:

```
--- Expected
+++ Actual
         <ram:CategoryCode>S</ram:CategoryCode>
-        <ram:RateApplicablePercent>20.00</ram:RateApplicablePercent>
+        <ram:RateApplicablePercent>0.00</ram:RateApplicablePercent>
```

Bisected, not guessed: `2d2e8ad` (the commit just before the merge) is green, `55d35ce` is red on all
5 sample cases.

### Cause

#709 is right about what it changes: a VAT breakdown group (BG-23) is identified by category (BT-118),
rate (BT-119) and exemption reason (BT-120/BT-121), never by the Dolibarr `vat_src_code`. It rekeys
the `$taxBreakdown` accumulator accordingly, from `"20 (TVAFR20)"` to `"S|20.0000||"`.

But that key was being read as data in two other places, which the pull request did not follow:

**1. The rate emitted comes from the array key** — `CIIProtocol.class.php`:

```php
foreach ($invoiceData['taxBreakdown'] as $rate => $vals) {   // $rate is 0, 20.0, ...
    $settlement->appendChild($this->buildTaxNode($doc, $rate, $vals, ...));
}
```

and in `buildTaxNode()`:

```php
$floatrate = preg_replace('/\(.*\)/', '', (string) $rate);   // 'x.x (CODE)' -> 'x.x'
$tax->appendChild($doc->createElement('ram:RateApplicablePercent', number_format((float) $floatrate, 2, '.', '')));
```

`(float) "S|20.0000||"` is `0`. Hence BT-119 = `0.00` on every document, whatever the real rate. The
group carries the right `tva_tx`, it is simply not the value being read.

**2. The deduction of previous situations files under the old key** —
`buildinvoicelines.inc.php`, the block added for #674:

```php
$keyforvatrate = $line->tva_tx . ($line->vat_src_code ? ' (' . $line->vat_src_code . ')' : '');
if (!isset($taxBreakdown[$keyforvatrate])) {
    continue;                      // always true now: the breakdown is keyed 'S|20.0000||'
}
```

so the document level allowance (BT-107) that brings a situation invoice back onto its instalment is
silently dropped.

### What it does to a situation invoice

Bench, situation 1 at 30 % then situation 2 at 70 % of a 1000.00 line, generation only:

| | `2d2e8ad` (before #709) | `55d35ce` (main today) |
|---|---|---|
| Dolibarr `total_ttc` | 480.00 | 480.00 |
| BT-119 rate | 20.00 | **0.00** |
| BG-20 allowance (BT-107) | 1, amount 300.00 | **0** |
| BT-109 `TaxBasisTotalAmount` | 400.00 | **700.00** |
| BT-112 `GrandTotalAmount` | 480.00 | **840.00** |
| BT-115 `DuePayableAmount` | 480.00 | **840.00** |

Same on 18.0.10, 23.0.4 and 24.0.0. **The e-invoice claims 840.00 where Dolibarr bills 480.00** —
360.00 too much, on a document that goes to the platform and to the customer.

An ordinary invoice keeps the right totals (they are summed elsewhere) but still declares a 0.00 VAT
rate against a non-zero VAT amount, which contradicts BR-CO-17 and the S category rules and will not
survive a validator.

### Module

einvoicing

### Dolibarr Version

18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.0 — the code is version independent

### Environment PHP

8.4.25

### Environment Database

MariaDB 11.4.13

daGrumpf-bxp:
Fixes #749. Found while running the suite on another branch: `EInvoicingSamplesTest` is red on `main`.

@Pichinov-Jose — #709 itself is right, and this keeps what it fixed. It rekeyed the breakdown on what
identifies a BG-23 group in EN 16931 (category, rate, exemption reason) instead of the Dolibarr
`vat_src_code`. What it could not see is that the old key was being read back **as data** in two other
places.

## 1. BT-119 was emitted from the array key

`CIIProtocol.class.php`:

```php
foreach ($invoiceData['taxBreakdown'] as $rate => $vals) {   // $rate is 0, 20.0, ...
    $settlement->appendChild($this->buildTaxNode($doc, $rate, $vals, ...));
}
```

```php
$floatrate = preg_replace('/\(.*\)/', '', (string) $rate);   // 'x.x (CODE)' -> 'x.x'
$tax->appendChild($doc->createElement('ram:RateApplicablePercent', number_format((float) $floatrate, 2, '.', '')));
```

`(float) "S|20.0000||"` is `0`, so every generated document declared `RateApplicablePercent = 0.00`
against a non-zero `CalculatedAmount`. The rate now comes from `$vals['tva_tx']`, which the group has
always carried; the `preg_replace` that stripped the ` (CODE)` suffix of the former key goes away with
it, and so does the `$rate` parameter.

## 2. The deduction of previous situations filed under the old key

`buildinvoicelines.inc.php`, the block added for #674:

```php
$keyforvatrate = $line->tva_tx . ($line->vat_src_code ? ' (' . $line->vat_src_code . ')' : '');
if (!isset($taxBreakdown[$keyforvatrate])) {
    continue;                      // always true now
}
```

The document level allowance (BT-107) that brings a situation invoice back onto its instalment was
silently dropped.

Both places now build the key through `einvoicingVatBreakdownKey()`. A key that identifies a group is
easy to fill under one shape and read back under another — that is exactly what happened twice here,
so it gets one definition.

Three trees compared on the same instance, symlink swapped between each run: `2d2e8ad` (just before
the merge of #709), `55d35ce` (`main` today) and this branch.

**The case #709 is about** — two lines, same category and same rate, one carrying `vat_src_code`
`TVAFR20` and one without:

| | `2d2e8ad` | `55d35ce` (main) | this branch |
|---|---|---|---|
| BG-23 groups | **2** (the bug #709 fixed) | 1 | 1 |
| RateApplicablePercent | 20.00 | **0.00** | 20.00 |
| BasisAmount | 100.00 + 50.00 | 150.00 | 150.00 |

**A situation invoice** — situation 1 at 30 %, situation 2 at 70 % of a 1000.00 line:

| | `2d2e8ad` | `55d35ce` (main) | this branch |
|---|---|---|---|
| Dolibarr `total_ttc` | 480.00 | 480.00 | 480.00 |
| BT-119 rate | 20.00 | **0.00** | 20.00 |
| BG-20 allowance (BT-107) | 1, 300.00 | **0** | 1, 300.00 |
| BT-109 `TaxBasisTotalAmount` | 400.00 | **700.00** | 400.00 |
| BT-112 `GrandTotalAmount` | 480.00 | **840.00** | 480.00 |
| BT-115 `DuePayableAmount` | 480.00 | **840.00** | 480.00 |

Reproduced identically on **18.0.10**, **23.0.4** and **24.0.0**, and both cases come back to the
`2d2e8ad` values on this branch while the single-group merge of #709 is kept.

PHPUnit of the module, file by file (never `AllTests.php`), on 18.0.10 and 23.0.4:
**21 OK / 0 KO** — against 20 OK / 1 KO (`EInvoicingSamplesTest`) on `main` at 55d35ce.

### DIFF
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -437,7 +437,7 @@ foreach ($object->lines as $line) {
 	// Keying the breakdown by vat_src_code split otherwise identical "S"/rate lines into two
 	// ApplicableTradeTax groups (e.g. products carrying code TVAFR20 vs services without code), which
 	// the PDP rejects (BR-FXEXT-S08b / BR-S-08: duplicate breakdown, taxable base does not reconcile).
-	$vatBreakdownKey = $categoryVAT.'|'.$line->tva_tx.'|'.$exemptionReasonCode.'|'.$exemptionReason;
+	$vatBreakdownKey = einvoicingVatBreakdownKey($categoryVAT, $line->tva_tx, $exemptionReasonCode, $exemptionReason);
 
 	// if ($line->subprice < 0 || $line->subprice_ttc < 0) {
 	// 	throw new Exception("NEGATIVE_UNIT_PRICE_NOT_ALLOWED: Unit price in lines can't be negative. Try to edit the line with ID " . $line->id);
@@ -766,9 +766,12 @@ if (!empty($object->situation_counter) && $object->situation_counter > 1
 			continue;
 		}
 
-		// The allowance reduces the basis of a VAT rate of THIS invoice, so it is filed under the rate
-		// of the line as it stands now, which is the one the breakdown knows.
-		$keyforvatrate = $line->tva_tx . ($line->vat_src_code ? ' (' . $line->vat_src_code . ')' : '');
+		// The allowance reduces the basis of a VAT breakdown group of THIS invoice, so it is filed under
+		// the group of the line as it stands now, which is the one the breakdown knows. The key has to be
+		// built exactly the way the breakdown above built it, hence the shared helper: filed under any
+		// other shape, the deduction lands on a group that does not exist and is silently dropped.
+		$tmpcategory = $this->getCategoryRate($line, $mysoc, $object);
+		$keyforvatrate = einvoicingVatBreakdownKey($tmpcategory['categoryVAT'], $line->tva_tx, $tmpcategory['ExemptionReasonCode'], $tmpcategory['ExemptionReason']);
 		if (!isset($taxBreakdown[$keyforvatrate])) {
 			continue;
 		}
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -2446,13 +2446,13 @@ class CIIProtocol extends AbstractProtocol
 			}
 
 			// VAT array by rate (tax breakdown)
-			foreach ($invoiceData['taxBreakdown'] as $rate => $vals) {		// $rate is 0, 20.0, ..., $vals is an array
+			foreach ($invoiceData['taxBreakdown'] as $vals) {		// $vals is one VAT breakdown group (BG-23)
 				// Add comment
 				$comment = $doc->createComment('VAT rate: '.$vals['tva_tx'].', VAT src code: '.$vals['vat_src_code'].', ExemptionReasonCode: '.$vals['ExemptionReasonCode']);
 				$settlement->appendChild($comment);
 
 				$settlement->appendChild(
-					$this->buildTaxNode($doc, $rate, $vals, $invoiceData['invoiceCurrency'], $invoiceData['vatDueDateTypeCode'] ?? '') 	// ApplicableTradeTax
+					$this->buildTaxNode($doc, $vals, $invoiceData['invoiceCurrency'], $invoiceData['vatDueDateTypeCode'] ?? '') 	// ApplicableTradeTax
 				);
 			}
 
@@ -3005,13 +3005,12 @@ class CIIProtocol extends AbstractProtocol
 	 * Build a tax node.
 	 *
 	 * @param \DOMDocument 		$doc 		Document to create nodes in
-	 * @param float|string      $rate 		Tax rate
-	 * @param array       		$vals 		Array containing tax values
+	 * @param array       		$vals 		One VAT breakdown group (BG-23): rate, category, exemption and totals
 	 * @param string       		$currency 	Currency code
 	 * @param string       		$dueDateTypeCode 	BT-8, VAT point date code ('5', '29' or '72'), empty to omit it
 	 * @return \DOMElement
 	 */
-	private function buildTaxNode($doc, $rate, $vals, $currency, $dueDateTypeCode = '')
+	private function buildTaxNode($doc, $vals, $currency, $dueDateTypeCode = '')
 	{
 		$tax = $doc->createElement('ram:ApplicableTradeTax');
 
@@ -3035,9 +3034,9 @@ class CIIProtocol extends AbstractProtocol
 			$tax->appendChild($doc->createElement('ram:DueDateTypeCode', htmlspecialchars((string) $dueDateTypeCode)));
 		}
 
-		$floatrate = preg_replace('/\(.*\)/', '', (string) $rate);		// If $rate is 'x.x (CODE)', we change it into 'x.x'
-
-		$tax->appendChild($doc->createElement('ram:RateApplicablePercent', number_format((float) $floatrate, 2, '.', '')));
+		// BT-119 comes from the group itself, never from the key it is filed under: that key identifies
+		// the group (category, rate and exemption reason together) and is not a number.
+		$tax->appendChild($doc->createElement('ram:RateApplicablePercent', number_format((float) $vals['tva_tx'], 2, '.', '')));
 
 		return $tax;
 	}
--- a/einvoicing/lib/einvoicing.lib.php
+++ b/einvoicing/lib/einvoicing.lib.php
@@ -965,3 +965,25 @@ function einvoicingCheckoutCommit($repodir)
 
 	return (preg_match('/^[0-9a-f]{40,}$/', $commit) ? substr($commit, 0, 7) : '');
 }
+
+/**
+ * Key identifying the VAT breakdown group (BG-23) a line belongs to.
+ *
+ * EN 16931 identifies a breakdown group by its VAT category code (BT-118), its rate (BT-119) and its
+ * exemption reason (BT-120/BT-121), and by nothing else - in particular not by the Dolibarr
+ * vat_src_code, which used to split otherwise identical groups in two and had the platform reject the
+ * document (BR-S-08: the taxable base does not reconcile).
+ *
+ * It exists as a function because the same key is built in more than one place: a breakdown filled
+ * under one shape and read back under another silently loses what was filed under it.
+ *
+ * @param	string		$categoryVAT			VAT category code of the line (BT-118)
+ * @param	float|string	$rate				VAT rate of the line (BT-119)
+ * @param	string		$exemptionReasonCode	Exemption reason code (BT-121), empty when there is none
+ * @param	string		$exemptionReason		Exemption reason text (BT-120), empty when there is none
+ * @return	string								Key of the group in the breakdown accumulator
+ */
+function einvoicingVatBreakdownKey($categoryVAT, $rate, $exemptionReasonCode = '', $exemptionReason = '')
+{
+	return $categoryVAT.'|'.$rate.'|'.$exemptionReasonCode.'|'.$exemptionReason;
+}
--- a/einvoicing/test/phpunit/SituationInvoiceTotalsTest.php
+++ b/einvoicing/test/phpunit/SituationInvoiceTotalsTest.php
@@ -1,0 +1,367 @@
+<?php
+/* Copyright (C) 2026 Pierre Grasswill
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program. If not, see <https://www.gnu.org/licenses/>.
+ * or see https://www.gnu.org/
+ */
+
+/**
+ *      \file       test/phpunit/SituationInvoiceTotalsTest.php
+ *      \ingroup    test
+ *      \brief      A generated document states the amount Dolibarr bills, not one of its own.
+ *      \remarks    No validator can catch this. A document that overstates its amounts is
+ *                  arithmetically consistent with itself - the lines, the breakdown and the totals
+ *                  all agree - so the CEN, Factur-X and CTC-FR schematrons answer VALID on it. The
+ *                  only thing it contradicts is the invoice it came from, and that is outside every
+ *                  one of them.
+ *
+ *                  #709 did exactly that: it filed the deduction of the previous situations (#674)
+ *                  under one key shape and read it back under another, so the document level
+ *                  allowance (BT-107) was silently dropped. A second situation of 70 % on a 1000.00
+ *                  line announced 840.00 where Dolibarr bills 480.00 - 360.00 too much, on a
+ *                  document that goes to the platform and to the customer, and every validator said
+ *                  the document was fine.
+ *
+ *                  This builds that very cycle - two situations, in the database, in the transaction
+ *                  the test class rolls back - and requires the totals of the document to be the
+ *                  totals of the invoice.
+ */
+
+
+// This script must only be run from the command line.
+if (PHP_SAPI !== 'cli') {
+	echo "Error: this script must be run from the command line (CLI), not through a web server.\n";
+	exit(1);
+}
+
+global $conf, $user, $langs, $db;
+
+// Load Dolibarr environment. Same resolution as the other test files of the module.
+$dolibarrHtdocs = getenv('DOLIBARR_HTDOCS');
+if (!$dolibarrHtdocs) {
+	$dolibarrHtdocs = dirname(__FILE__) . '/../../htdocs';
+}
+if (!file_exists($dolibarrHtdocs . '/master.inc.php')) {
+	throw new \RuntimeException('Could not locate master.inc.php under "' . $dolibarrHtdocs . '/". Set the environment variable (export DOLIBARR_HTDOCS=...) to the htdocs directory of the Dolibarr instance to test against.');
+}
+require_once $dolibarrHtdocs . '/master.inc.php';
+require_once DOL_DOCUMENT_ROOT . '/user/class/user.class.php';
+require_once DOL_DOCUMENT_ROOT . '/societe/class/societe.class.php';
+require_once DOL_DOCUMENT_ROOT . '/compta/facture/class/facture.class.php';
+
+/**
+ * @var Conf $conf
+ * @var DoliDB $db
+ * @var Translate $langs
+ * @var User $user
+ */
+
+dol_include_once('einvoicing/class/protocols/CIIProtocol.class.php');
+require_once __DIR__ . '/CommonClassTestCompat.inc.php';
+
+
+/**
+ * Class SituationInvoiceTotalsTest
+ *
+ * Builds a situation cycle of two invoices and checks the document generated for the second one
+ * against what the core says the customer owes.
+ */
+class SituationInvoiceTotalsTest extends CommonClassTest
+{
+	const RAM = 'urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100';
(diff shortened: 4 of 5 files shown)


daGrumpf-bxp:
Confirmed against a real Access Point, not only against the bench.

A full end-to-end cycle on the sandbox — a new invoice issued, routed by the platform, received by
another instance, lifecycle statuses back — was run twice, the only difference between the two runs
being the commit of this pull request.

**Without it**, the platform refuses the flow outright:

```
--- flow i_431169 at the Access Point:
  http=200
  acknowledgement = {"status":"Error"}
```

The document it refused carries exactly what is described above:

```xml
<ram:CalculatedAmount>40.00</ram:CalculatedAmount>
<ram:RateApplicablePercent>0.00</ram:RateApplicablePercent>
```

**With it**, the same invoice goes through and comes back:

```
  RESULT REF=TRI-FA-2609-0604 FLOW=i_431179 code=15
  sync 1 "syncedFlows":2
  dd23 supplier invoice 8458 ref_supplier=TRI-FA-2609-0604 ht=200.00 ttc=240.00
  lifecycle_msg 200 flow=ie_1262089
  lifecycle_msg 201 flow=ie_1262090
  received=yes
```

So this is not only a unit-level discrepancy: as `main` stands, an invoice emitted by the module is
rejected by the platform.

daGrumpf-bxp:
To be precise about what was tested, since `EInvoicingSamplesTest` is only what *detected* this: the
two documents above are **ordinary customer invoices**, not samples — same third party, two service
lines, 200.00 HT / 40.00 VAT / 240.00 TTC, created and validated on the bench, generated by the
module and transmitted. The only difference between them is the commit of this pull request.

Their header VAT breakdown, everything else being identical:

```xml
<!-- TRI-FA-2609-0603, generated by main, refused by the platform -->
<ram:ApplicableTradeTax>
  <ram:CalculatedAmount>40.00</ram:CalculatedAmount>
  <ram:BasisAmount>200.00</ram:BasisAmount>
  <ram:CategoryCode>S</ram:CategoryCode>
  <ram:RateApplicablePercent>0.00</ram:RateApplicablePercent>
</ram:ApplicableTradeTax>

<!-- TRI-FA-2609-0604, generated with this pull request, accepted -->
<ram:ApplicableTradeTax>
  <ram:CalculatedAmount>40.00</ram:CalculatedAmount>
  <ram:BasisAmount>200.00</ram:BasisAmount>
  <ram:CategoryCode>S</ram:CategoryCode>
  <ram:RateApplicablePercent>20.00</ram:RateApplicablePercent>
</ram:ApplicableTradeTax>
```

Both submitted to the SuperPDP validator
(`POST https://api.superpdp.tech/v1.beta/validation_reports`), which names the rules:

| | `main` (TRI-FA-2609-0603) | this pull request (TRI-FA-2609-0604) |
|---|---|---|
| XSD `CII_D22B_uncoupled` | 1 check, 0 failure | 1 check, 0 failure |
| `FACTUR-X_EN16931.xslt` | 122 checks, **3 failures** | 122 checks, **0 failure** |
| `BR-FR-Flux2-Schematron-CII_WARNING.xslt` | 80 checks, 0 failure | 80 checks, 0 failure |
| verdict | **`is_valid: false`** | **`is_valid: true`** |

The three failures on the document `main` produces:

```
BR-CO-17  VAT category tax amount (BT-117) = VAT category taxable amount (BT-116)
          x (VAT category rate (BT-119) / 100), rounded to two decimals.
BR-S-09   The VAT category tax amount (BT-117) in a VAT breakdown (BG-23) where VAT category
          code (BT-118) is "Standard rated" shall equal …
BR-S-08   For each different value of VAT category rate (BT-119) where the VAT category code
          (BT-118) is "Standard rated", the VAT category taxable amount (BT-116) shall equal …
```

The pair is its own negative control: the same validator, on the same invoice, says no to one and
yes to the other.

Note that only the **header** breakdown is wrong — the line level `ApplicableTradeTax` still reads
20.00, which is why this is easy to miss when reading the XML quickly.

daGrumpf-bxp:
Raising the priority of this one, with the reason.

`main` has carried this since 55d35ce was merged (3 Sep, 20:25). Anything packaged from `main` after
that point ships it — I hit it on packages built from `main` this evening, and checked by running the
packaged code rather than by reading it: the XML it generates is refused by the validator on **both**
an ordinary invoice and a situation invoice.

What it means for anyone on `main` today:

- every invoice **carrying VAT** is refused by the platform (BR-CO-17, BR-S-09, BR-S-08). Not a wrong
  invoice that gets through — one that cannot be sent at all;
- invoices that are **fully exempt** are unaffected: 0.00 is the correct rate there;
- and a **situation invoice** carries the wrong amount in its XML: `GrandTotalAmount` and
  `DuePayableAmount` state the cumulative instead of the instalment (840.00 against the 480.00
  Dolibarr bills, on the bench case). Measured on the Factur-X path too, where the readable page shows
  480.00 while the embedded XML says 840.00 — a document that says two different things, and nothing
  validates it when it travels outside the platform.

No tag contains 55d35ce at the time of writing, so no release is affected yet. **Worth not tagging
`main` until this is in.**

Nothing here is a criticism of #709, which fixes a real rejection of its own and whose key is kept as
it is. The two places that read that key as data were simply not visible from it.

daGrumpf-bxp:
@eldy heads-up: **`main` is broken right now**, and has been since 55d35ce was merged today at
**20:25 (+0200)**.

Every invoice carrying VAT that the module generates from `main` is **refused by the platform** —
`RateApplicablePercent` is emitted as `0.00` against a non-zero VAT amount, so BR-CO-17, BR-S-09 and
BR-S-08 all fail. Verified end to end on the sandbox (the flow comes back
`acknowledgement: {"status":"Error"}`) and against the SuperPDP validator (`is_valid: false`,
3 failures), on ordinary customer invoices, not on the samples.

A situation invoice is worse: its XML states 840.00 where Dolibarr bills 480.00.

This pull request is the fix, tested on 18.0.10 / 23.0.4 / 24.0.0 and green end to end; the detail and
the validator reports are in the comments above.

Two things while it is not merged:

- **please do not tag `main`** — no tag contains 55d35ce yet, so no release is affected so far;
- anyone packaging from `main` since 20:25 is shipping it. Worth a word to whoever pulls from it.

#709 itself is not in question: its key is right and this keeps it. Two places were reading that key
as a value, which its pull request could not see.

daGrumpf-bxp:
Corroboration from another bench, measured tonight while running the lifecycle of #757 on the public-URL bench.

Same instance, same emission script, same minute, three trees:

| tree | BT-119 emitted | acknowledgement of the access point |
|---|---|---|
| `main` | `0.00` | `Error` |
| `main` + another branch that touches nothing in the generation | `0.00` | `Error` |
| this PR | `20.00` | `Ok` |
| this PR + that other branch | `20.00` | `Ok` |

The refusal is immediate - a few hundred milliseconds after the submission - and the access point gives no detail beyond `acknowledgement: {"status":"Error"}`; the only difference between the two documents is the one line `<ram:RateApplicablePercent>`. 15 emissions on `main` tonight, 15 refused; with this PR in place the full lifecycle goes through to 212.

So on `main` as it stands, no invoice carrying VAT leaves the platform, and that also makes any end-to-end recette of another branch impossible without stacking it on this one.

daGrumpf-bxp:
Added the two regression tests to this branch (`8b7bd08`), one per half of the defect. Both are red on
`main` as it stands and green on this branch — measured on three cores, not supposed.

**`VatBreakdownStatesItsRateTest`** reads the five specimen documents as they are generated and
requires each VAT breakdown to state the rate that produced its tax amount (BR-CO-17), and no two
breakdowns to share a category and a rate — BR-S-08, which is what #709 set out to fix. It compares
against no reference document, so regenerating the fixtures cannot make it green again.

**`SituationInvoiceTotalsTest`** builds the case no validator can see: a cycle of two situations
created in the database, inside the transaction the test class rolls back, and the document generated
for the second one has to state the amounts the core bills — 400.00 net and 480.00 gross where the
lines say 700.00, the difference carried as the document level allowance of #674. On `main` it reads
700.00 and 840.00. A document that overstates its totals is arithmetically consistent with itself, so
the CEN, Factur-X and CTC-FR schematrons all answer VALID on it; the invoice it came from is the only
thing it contradicts, and that is outside every validator.

Runs, with the PHPUnit job of #760 doing the running:

- on `main` (this defect still in): `EInvoicingSamplesTest`, `SituationInvoiceTotalsTest` and
  `VatBreakdownStatesItsRateTest` fail on 18.0.10, 23.0.4 and 24.0.0 —
  https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/33810893272
- on this branch: 23 test files pass on the three cores —
  https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/33810782627

Two notes from making them run on a bare instance, which is what the CI has: the buyer is created
with an explicit customer code (the demo database numbers its customers with a module that refuses a
third party without one), and the seller identifiers are pinned on `$mysoc` in memory and restored
after, because a demo company whose SIREN is `123456` stops the generation before the first total is
read.

einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -766,9 +766,12 @@
...
 
-		// The allowance reduces the basis of a VAT rate of THIS invoice, so it is filed under the rate
-		// of the line as it stands now, which is the one the breakdown knows.
-		$keyforvatrate = $line->tva_tx . ($line->vat_src_code ? ' (' . $line->vat_src_code . ')' : '');
+		// The allowance reduces the basis of a VAT breakdown group of THIS invoice, so it is filed under
+		// the group of the line as it stands now, which is the one the breakdown knows. The key has to be
+		// built exactly the way the breakdown above built it, hence the shared helper: filed under any
+		// other shape, the deduction lands on a group that does not exist and is silently dropped.
+		$tmpcategory = $this->getCategoryRate($line, $mysoc, $object);
+		$keyforvatrate = einvoicingVatBreakdownKey($tmpcategory['categoryVAT'], $line->tva_tx, $tmpcategory['ExemptionReasonCode'], $tmpcategory['ExemptionReason']);
```
eldy, line 774: This is strange that we don't have the $line->vat_src_code anymore in the breakdownkey.

In Dolibarr, we can have 2 vat rate 0, with different code so depending on the code the einvoice_vatex (the column einvoice_vatex in tha table of vat rates llx_c_tva) will differ.

This means the vatex code (reason for vat excemption does not need to be in the grouping vat criteria ?
Is this done on purpose ? May be it is the ExemptionReasonCode that replace this in the key ?

daGrumpf-bxp: On purpose, yes — and yes, it is the exemption reason that stands in its place.

The key is `categoryVAT | rate | ExemptionReasonCode | ExemptionReason`, i.e. BT-118, BT-119, BT-121
and BT-120: the four things EN 16931 identifies a BG-23 group by. `vat_src_code` is not dropped, it is
*resolved* one step earlier. `getCategoryRate()` reads the dictionary line the invoice actually used —
`_getVatDictionaryEntry($vat_rate, $vat_src_code)` — and takes `einvoice_vatex` for BT-121 and the
`note` for BT-120. So two rate 0 lines with different codes arrive at the key with different values and
stay in two groups; what the key holds is what the document will state, not what the database calls it.

What no longer splits is two lines whose codes resolve to the *same* category, rate and reason. That is
#709: `S`/20 with `TVAFR20` and `S`/20 without a code gave two `ApplicableTradeTax` that no field of the
document tells apart, which BR-S-08 refuses.

Measured just now on the bench, module at `241f0e61` (`main`, #750 in), one invoice per case:

| case | document level BG-23 groups |
|---|---|
| two lines at 0 %, codes `E-CGI261-1` and `E-CGI261-4`, VATEX `VATEX-FR-CGI261-1` and `VATEX-FR-CGI261-4` | **2** — `E`/0.00/100.00 each, BT-121 and BT-120 of its own dictionary line |
| two lines at 20 %, one with `TVAFR20` and one without | **1** — `S`/20.00, basis 200.00, tax 40.00 |

Identical on **18.0.10**, **23.0.4** and **24.0.0**. Below 24 there is no `einvoice_vatex` column, so
BT-121 comes from `MAIN_VAT_EXEMPTION_CODE_FOR_<rate>_<CODE>`, which is per code as well — the two
groups come out the same on those cores. Bench script: `vatex750.php`, it inserts the two dictionary
lines, generates, reads back `ApplicableHeaderTradeSettlement` and removes them again.

Two edges worth naming, since they are what makes this safe:

- when neither a note nor a code can be found for an `E`/`AE`/`G`/`K` line, the module throws
  `MISSINGSETUP` naming the code it could not translate (BR-E-10 and siblings), rather than emit a
  group with an empty reason. So two codes cannot quietly merge on an empty key;
- `Z` is the one place two different codes do share a group, and that is deliberate: BR-Z-09 and
  BR-Z-10 forbid an exemption reason on a zero rated line, so nothing in the document tells them
  apart and one group is the only valid answer.

daGrumpf-bxp: Not on purpose in the sense you are asking, and your instinct is the right one: the code does not need
to be in the key, but what replaced it should not be there either.

How it works today: `vat_src_code` is not dropped, it is resolved one step earlier. `getCategoryRate()`
reads the dictionary line the invoice actually used — `_getVatDictionaryEntry($vat_rate, $vat_src_code)` —
and takes `einvoice_vatex` for BT-121 and the `note` for BT-120, both of which then sit in the key
`categoryVAT | rate | ExemptionReasonCode | ExemptionReason`. So two 0 % lines with different codes do
come out as two breakdown groups.

And that is the problem, because EN 16931 does not group the same way for every category:

- `S` groups **per rate**: `BR-S-08` asserts `every $rate in ../ram:RateApplicablePercent satisfies
  (../ram:BasisAmount = sum of the lines at that rate)`. Two `S`/20 groups are refused — that is #709,
  and it stays fixed.
- `E`, `Z`, `AE`, `K`, `G`, `O` take **exactly one group per category, whatever the reason**. `BR-E-01`
  asserts `count(//ram:ApplicableHeaderTradeSettlement/ram:ApplicableTradeTax[ram:CategoryCode='E']) = 1`,
  and `BR-E-08` sums *every* line of category `E` against that single `BasisAmount`.

Two invoices generated on the bench for this, and passed through both validators — 2 lines at 0 % with
`E-CGI261-1` and `E-CGI261-4` (different `einvoice_vatex`), and 2 lines at 20 % with and without
`TVAFR20`:

| document | FNFE chain, run locally | api.superpdp.tech |
|---|---|---|
| 2 lines at 0 %, 2 exempt codes | **INVALID** — 120 checks, 3 failures: `BR-E-08` ×2 and `BR-E-01` | `is_valid: false`, the same 3 |
| 2 lines at 20 %, one with `TVAFR20` one without | VALID — 0 failure | `is_valid: true` |

Same chain both sides (XSD `CII_D22B_uncoupled`, `FACTUR-X_EN16931.xslt`, then
`BR-FR-Flux2-Schematron-CII_WARNING.xslt`, which reports 0 failure on both).

Two things this is **not**:

- not a regression of #709 or #750. Replayed on `2d2e8ad`, before #709, the same invoice already came
  out with two `E` groups: the old key split on `vat_src_code`, the new one splits on the reason that
  code resolves to. Same wrong count, different cause.
- not reachable with the dictionary as it ships: the French one has a single 0 % line, no code and no
  `einvoice_vatex`. It becomes reachable the moment an installation fills in the column 24 added, which
  is exactly the setup your question describes.

What EN 16931 has no room for is two exemption reasons on one invoice. The key wants to be
`category | rate`, and the single group then needs one reason — quote the one of the first exempt line
and say nothing of the second, or refuse the invoice naming both codes. That is a call I would rather
not make alone inside a merged pull request: I will open an issue with the two documents and both
reports attached, unless you would rather keep it on this thread.

daGrumpf-bxp: Checked against French law, and it runs the other way from EN 16931, so the 24 model is not the odd one out: CGI annex II art. 242 nonies A puts the rate on each line (I-8°) and asks, in case of exemption, for "la référence à la disposition pertinente" that founds it (I-12°), while BOFiP BOI-TVA-DECLA-30-20-20-10 § 260 groups that mention by "chaque rubrique de fournitures soumises au même taux d'imposition **ou à la même mesure d'exonération**" and § 490 attaches the reference to the operation, not to the invoice.
So two exemption measures on one invoice are expected in law, each carrying its own legal basis — which is exactly what `einvoice_vatex` per dictionary line lets an operator express.
EN 16931 allows exactly one: `BR-E-01` counts a single `E` breakdown and `BR-E-08` sums every `E` line into it, so the document the law asks for cannot be written in the syntax.
And no French rule closes the gap in the other direction: on the very document the CEN layer refuses, `BR-FR-Flux2-Schematron-CII` reports 0 failure.
Which way should the module fall — one group carrying the reason of the first exempt line, or a refusal naming both codes and asking the operator to split the invoice?

PR #751, opened 2026-09-03
FIX [einvoicing] Both providers read the shape of a received flow they can import, not only the Converted one

daGrumpf-bxp:
Follow-up of #718 and of the UBL invoice @hgy29 attached there. **Supersedes #748**, which fixed the
crash but not what caused the flow to be unreadable in the first place — close that one if this is
merged, they touch the same lines.

@eldy — this is about the *direction* of `72b19fac` ("Standardize code Esalink vs SuperPDP"), not
about standardizing, which is right. That commit commented out the SuperPDP loop and copied the
Esalink shape into SuperPDP. The loop was there on purpose, and it was the thing that protected the
import from exactly the case reported on #718.

## What the single 'Converted' fetch costs

```php
$flowResponse = $this->fetchFlowData($flowId, 'Converted');
```

The `Converted` document is the invoice rewritten into the syntax configured **on the Access Point
account**, outside Dolibarr. Set to UBL — a syntax that belongs to the French socle and has no reader
in this module — the whole inbox of the instance becomes unreadable, *including the invoices whose
issuer sent a CII or a Factur-X*: the platform hands over its UBL rewrite of them and nothing else is
looked at. `getProtocol('UBL')` returns null, and the import is called on that null, which raises an
`Error` — not an `Exception`, so neither `catch` in the chain takes it and the synchronization request
dies on a fatal.

That is the report on #718: an inbox frozen on one document, with no message saying why.

## Why the loop, and why on both providers

The three shapes of a flow are independent: `Converted` is the platform's rewrite, `Original` is what
the issuer really sent, `ReadableView` is the human readable copy — which, on an access point that
builds it as a Factur-X PDF, carries the same data again. A miss on one says nothing about the others.

`Converted` still comes first, because it is the one that shields the import from an issuer emitting a
syntax with no reader here. It is simply no longer the only thing tried.

And this belongs on Esalink too: its API documents the very same enum. From the Esalink/Hubtimize
API specification v1.2.6.2, `GET /flows/{flowId}`, parameter `docType`:

> `- Metadata [Default] … - Original: the document that has been initially sent/provided by the
> emitter - Converted: the document that has been optionally converted by the system - ReadableView:
> the document that has been optionally generated as the readable file`
>
> `enum: [Original, Converted, ReadableView, Metadata]`

So the standardization happens here in the other direction: the picking moves to
`AbstractPDPProvider::fetchImportableFlowDocument()`, one implementation used by both providers,
instead of one written for SuperPDP and none for Esalink.

When none of the three can be read, the flow is **postponed**, not failed: nothing was stored, it
stays pending at the Access Point, the batch carries on, and the user gets the message that already
exists for it (`SetTheAccessPointConversionFormat`, "CII or Factur-X, not UBL").

Driven through the actual `syncFlows()` of each provider against a fake Access Point that holds a
**different document per docType**, with a batch of two flows: the flow under test, then a Factur-X
flow the module reads, to show whether the batch carried on.

The UBL document is the real one from #718 (`SFR.9A0040013195.ubl.xml`, a SuperPDP delivery).

| scenario | `main` (55d35ce) | this branch |
|---|---|---|
| `Converted` = UBL, `Original` = Factur-X | `Error … on null`, **request dies** | Converted refused → **Original imported**, batch carries on |
| `Converted` = UBL, `Original` = UBL | `Error … on null`, **request dies** | all three tried → flow **postponed** with the action, batch carries on |
| `Converted` = Factur-X (nominal) | imported | imported, **one single API call**, unchanged |

Same on **18.0.10**, **23.0.4** and **24.0.0**, and on **both providers** — 12 runs.

The API calls the bench observed, on the first scenario:

```
AP call : T…  docType=Converted   -> 200 (114510 bytes)     # the UBL, refused
AP call : T…  docType=Original    -> 200 (108492 bytes)     # the Factur-X, imported
AP call : Z…  docType=Converted   -> 200 (108485 bytes)     # the flow behind it, imported too
res : 1 synced=2
```

and on the nominal one, the branch issues exactly the same single `Converted` call as `main` — the
fallback costs nothing when the first shape is readable.

PHPUnit of the module, file by file (never `AllTests.php`), on 18.0.10 and 23.0.4: **20 OK / 1 KO**,
the same count as `main` — the one failure is `EInvoicingSamplesTest`, the regression of #749 which
has its own pull request (#750).

## On UBL itself

This does not make the module read UBL, and a customer whose access point can only hand over UBL is
still stuck — the right answer for them today is to set the conversion format to CII or Factur-X.

If you want the module to read UBL, I am happy to write it. The import is table driven, so a
`UBLProtocol extends CIIProtocol` overriding the namespaces, the two XPath maps (80 header fields, 41
line fields) and the 7 multi-node sub-parsers inherits the whole import engine as it stands. The real
work is that the mappings are not 1:1 — `cac:TaxTotal/cac:TaxSubtotal` against `ram:ApplicableTradeTax`,
`cbc:ChargeIndicator` against two separate CII paths, party identity split across three containers,
plain dates instead of `udt:DateTimeString`, and a credit note being a different root element
(`<CreditNote>`) rather than a type code on the same one. Say the word and I will open an issue to
scope it properly rather than guess at it in a comment.


daGrumpf-bxp:
Two additions to the recipe above, both run after the pull request was opened.

**1. The `nominal` scenario, on the three cores instead of one.** The table said "same on 18.0.10,
23.0.4 and 24.0.0" for all three scenarios; the nominal one had only been measured on 23.0.4. Now run
on 18.0.10 and 24.0.0 as well, both providers, `main` and this branch: identical everywhere — a single
`Converted` call, invoice imported. The fallback costs nothing when the first shape is readable.

**2. A real end-to-end cycle on the sandbox**, which the bench above does not cover: it fakes the
Access Point, so it proves what the module does with an answer, not that the platform answers that way.

A new invoice is issued by one instance, routed by the Access Point, received by another, and the
lifecycle statuses come back — everything on this branch, on both sides:

```
--- emit on dolidev:
  RESULT REF=TRI-FA-2609-0604 FLOW=i_431179 code=15
--- reception on dd23:
  sync 1 "syncedFlows":2
  dd23 supplier invoice 8458 ref_supplier=TRI-FA-2609-0604 ht=200.00 ttc=240.00
    line 40225 vat=20.0000 ht=100.00 date_start=2026-06-01 date_end=2026-06-30
    line 40226 vat=20.0000 ht=100.00 date_start=2026-05-15 date_end=2026-07-31
--- statuses back on dolidev:
  lifecycle_msg 200 flow=ie_1262089
  lifecycle_msg 201 flow=ie_1262090
===== ref=TRI-FA-2609-0604 flow=i_431179 received=yes =====
```

Received on the very first synchronization, with the right lines, periods and amounts.

**Worth reporting: the first attempt at that cycle failed, and not because of this branch.** On this
branch alone — that is, on `main` as it stands — the Access Point **refused the flow**:

```
--- flow i_431169 at the Access Point:
  acknowledgement = {"status":"Error"}
```

The document it refused carries, in its header VAT breakdown:

```xml
<ram:CalculatedAmount>40.00</ram:CalculatedAmount>
<ram:RateApplicablePercent>0.00</ram:RateApplicablePercent>
```

which is #749 — the regression #709 introduced, fixed by #750. Adding that one commit on top of this
branch is the only difference between the refused cycle and the green one above.

So the cycle also says something about #750 that its own bench could not: **a real platform rejects
what `main` emits today**, and accepts it again with the fix.

PR #753, opened 2026-09-03
Qual: Use dol_strto{lower,upper}

mdeweerd:


eldy:
Because there is a high level of PR, to avoid conflicts i will no merge PR on code syntax during this period, until the code is more stabilized (in few month  ;-)

mdeweerd:
OK, but I am rather DRY!

PR #756, opened 2026-09-03
Qual: Use dol_basename

mdeweerd:


mdeweerd:
Issue: dol_basename is not a real replacement for php's basename.

einvoicing/call_card.php
```diff
@@ -91,7 +92,7 @@
 $action = GETPOST('action', 'aZ09');
 $confirm = GETPOST('confirm', 'alpha');
 $cancel = GETPOST('cancel', 'aZ09');
-$contextpage = GETPOST('contextpage', 'aZ') ? GETPOST('contextpage', 'aZ') : str_replace('_', '', basename(dirname(__FILE__)).basename(__FILE__, '.php')); // To manage different context of search
+$contextpage = GETPOST('contextpage', 'aZ') ? GETPOST('contextpage', 'aZ') : str_replace('_', '', dol_basename(dirname(__FILE__)).dol_basename(__FILE__, '.php')); // To manage different context of search
```
eldy: The dol_basename is used at 1 place in dolibarr code.
Also it seems bugged (ddoes not take into account windows \ for example).
If we have it at only one place, it probably means that we don't need it. This looks to be a legacy hack.

mdeweerd:
The best approach seems to be:
- Replace `basename($path)` with `pathinfo(__FILE__, PATHINFO_BASENAME);`
- Replace `basename($path, '.php')` with `pathinfo(__FILE__, PATHINFO_FILENAME);` when $path is the current file name.

In dolibarr, fix `dol_basename()` with `pathinfo(__FILE__, PATHINFO_BASENAME);`  .

eldy:
> * pathinfo

I suggest to first fix 

> `dol_basename()` with `pathinfo(__FILE__, PATHINFO_BASENAME);`

And why replacing basename with pathinfo(__FILE__, PATHINFO_BASENAME);   ?
It seems they are completely équivalent for basename function with same troubles with cyrillic.

mdeweerd:
Ok, I am going to drop this based on official doc and https://stackoverflow.com/questions/45268499/php-basename-and-pathinfo-with-multibytes-utf-8-file-names .

dol_basename could implement the regex and return the expected part:
https://www.php.net/manual/en/function.pathinfo.php#107461

PR #764, opened 2026-09-04
FIX #761 [einvoicing] A received document is judged on the invoice it carries, not on its envelope

### LINKED ISSUE #761: [einvoicing] A received hybrid PDF carrying UBL is read as Factur-X: the whole synchronization aborts, and the consistency check dies on a fatal
### Bug

The module decides which syntax a received document is written in by looking at the **outside** of the file, and never checks what the file actually **carries**.

`ProtocolManager::detectProtocolFromContent()` answers `FACTURX` for anything starting with `%PDF-`. What that PDF holds is a separate question: `ZugferdDocumentPdfReaderExt::resolveInvoiceDocumentContent()` picks the embedded invoice **by file name**, and its accepted list is

```php
public const ATTACHMENT_FILENAMES = [
    'ZUGFeRD-invoice.xml', 'zugferd-invoice.xml', 'factur-x.xml', 'xrechnung.xml'
];
```

`xrechnung.xml` is in that list, and an XRechnung is commonly **UBL** — a syntax this module recognizes but has no reader for (`$ublIsOk = 0`). So a perfectly ordinary German hybrid invoice is accepted as Factur-X and its UBL payload is handed to the CII reader.

This has two consequences, both reproduced on `main` (4079015b).

#### 1. One such invoice stops the whole incoming synchronization

`fetchImportableFlowDocument()` — the picker added by #751 — accepts the document without noticing:

```
protocol_name : 'FACTURX'
protocol      : FacturXProtocol
doc_type      : 'Converted'
attempts      : array()          <- no shape was rejected, nothing was noticed
payload really carried = 'UBL'
```

The CII reader then finds none of the elements it expects and returns a business error that names nothing actionable:

```
createSupplierInvoiceFromSource() =>
  array('res' => -1, 'message' => 'Unfounded dolibarr corresponding Invoice code for document type code: NA')
```

That `-1` carries no `postponeflow`, so `syncFlows()` does `$error++` and then

```php
if ($error > 0) {
    $results_messages[] = "Aborting synchronization due to errors.";
    break;
}
```

Every flow queued behind that one stops coming in. This is the failure mode #751 was written to remove, and it survives because the picker never opens the container. (Related, but not the same cause as #718 and #737: here a `Converted` document *is* delivered, and it *is* in a shape the module believes it can read.)

#### 2. The consistency check dies on a PHP fatal

With `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` on, validating such a received supplier invoice:

```
PHP Fatal error: Uncaught TypeError: get_class(): Argument #1 ($object) must be of type object,
null given in einvoicing/class/document.class.php:1298
#0 einvoicing/class/providers/AbstractPDPProvider.class.php(1128): Document::cleanXmlData()
#1 einvoicing/class/utils/SupplierInvoiceHelper.class.php(367): AbstractPDPProvider->fetchFlowXml()
#2 einvoicing/class/utils/SupplierInvoiceHelper.class.php(77): SupplierInvoiceHelper::getXmlData()
#3 einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php(290):
     SupplierInvoiceHelper::checkDolInvoiceAndEInvoiceConsistency()
#4 htdocs/core/class/interfaces.class.php(202): InterfaceEInvoicingTriggers->runTrigger()
#5 htdocs/core/class/commonobject.class.php(6354): CommonObject->call_trigger()
#6 htdocs/fourn/class/fournisseur.facture.class.php(1937): FactureFournisseur->validate()
```

`fetchFlowXml()` guards on `getProtocolFromContent()`, but that guard is computed on the PDF, while `extractXmlFromFileContent()` then returns the UBL. `cleanXmlData()` detects `UBL`, `getProtocol('UBL')` returns `null`, and `get_class(null)` is a fatal.

Worth noting on its own: `fetchFlowXml()` asks for the **`Original`** document, hardcoded, while the import reads the **`Converted`** one. A flow imported from a CII `Converted` can perfectly well have a UBL `Original`, so the two paths do not see the same document.

### Module

einvoicing

### Dolibarr Version

Reproduced on 22.0.5. The code involved is module code with no core API dependency, so every supported version (18 → 24) is concerned.

### Environment PHP

8.4

1. Have the access point return, for a received `SupplierInvoice` flow, a hybrid PDF whose embedded attachment is named `xrechnung.xml` and contains a UBL invoice (`urn:oasis:names:specification:ubl:schema:xsd:Invoice-2`). This is what a German issuer sending a hybrid XRechnung produces.
2. Run the incoming synchronization → `Aborting synchronization due to errors.`, with `Unfounded dolibarr corresponding Invoice code for document type code: NA` as the only explanation. Flows behind it are not imported.
3. With `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` set, validate a received supplier invoice whose flow holds such a document and whose `einvoicing_document.xml_data` is still empty → PHP fatal above.

**Expected:** a document whose payload cannot be read is treated exactly like one whose container cannot be read — the other shapes of the flow are tried, and if none is readable the flow is postponed with the "set the conversion format on your access point" action, so the batch carries on. And no code path dereferences a protocol that was never built.

**Control:** the same paths, fed a real Factur-X (CII inside the PDF), run to completion and produce a proper business comparison error — so the failures above come from the payload syntax, not from the container being a PDF.

daGrumpf-bxp:
Fixes #761.

`detectProtocolFromContent()` answers on the outside of a file: anything starting with `%PDF-` is Factur-X, whatever it holds. What it holds is a separate question, because `ZugferdDocumentPdfReaderExt` picks the embedded invoice **by file name** — and `xrechnung.xml` is one of the names it accepts, an XRechnung being commonly UBL, a syntax this module recognizes but has no reader for.

Nobody asked that second question, so an ordinary German hybrid invoice was accepted as Factur-X and its UBL handed to the CII reader. Two consequences:

- **Reception stops.** The CII reader finds none of the elements it expects and answers `Unfounded dolibarr corresponding Invoice code for document type code: NA`. That failure carries no `postponeflow`, so `syncFlows()` counts an error and breaks: one such invoice stops the whole incoming synchronization and every flow queued behind it. This is the failure mode `fetchImportableFlowDocument()` was built to remove, and it survived because the picker never opened the container.
- **The consistency check dies.** `fetchFlowXml()` guards on `getProtocolFromContent()`, but that guard is computed on the PDF while `extractXmlFromFileContent()` then returns the UBL. `cleanXmlData()` detects `UBL`, `getProtocol()` returns `null`, and `get_class(null)` kills the request — inside a trigger, so it takes the validation of the supplier invoice with it.

## The change

Both paths now ask the same question before accepting a document: not "what does this file look like" but "is there an invoice in it this module can read". One helper, `readableInvoicePayload()`, used by the picker and by `fetchFlowXml()`. A document that fails it is skipped like any other unreadable shape, so the other shapes of the flow are still tried and the flow ends up postponed with the action to do, instead of stalling the batch.

The document is deliberately **not** re-typed on what is found inside: a Factur-X PDF is still imported as Factur-X, whose reader does its own extraction. Only reachability is checked. `AbstractProtocol::extractXmlFromFileContent()` returns its input unchanged, so this costs nothing on a document that is already an XML.

Not addressed here, and worth a separate look: `fetchFlowXml()` asks for the **`Original`** document, hardcoded, while the import reads the **`Converted`** one — so the two paths do not necessarily see the same document.

## Measured, before and after

Same bench, same specimen, same harness — a hybrid PDF whose embedded `xrechnung.xml` holds a UBL invoice, which is what a German issuer sending a hybrid XRechnung produces. Only the network answer is stubbed. `before` is this branch's parent.

```
BEFORE  A. accepted as FacturXProtocol, attempts: (none - nothing was noticed)
           import res=-1 'Unfounded dolibarr corresponding Invoice code for document type code: NA'
           postponeflow=NULL  -> syncFlows() breaks the batch
        B. TypeError: get_class(): Argument #1 ($object) must be of type object, null given

AFTER   A. refused, attempts: Converted/Original/ReadableView: FACTURX but what it carries
           is UBL, which is not supported  -> postponeflow=1, the batch carries on
        B. returned null (nothing readable, no fatal)
```

Same on Dolibarr **18, 20, 22, 23 and 24**.

**Control — a real Factur-X must be untouched.** Same harness fed the module's own `INVTEST-dol_ok.facturx.pdf`: accepted, imported (`res=7499`), and `fetchFlowXml()` returns its 13017 bytes of XML. The fix rejects the unreadable payload and leaves the readable one strictly alone.

**Side benefit.** A plain PDF with no embedded invoice — a very ordinary `ReadableView` — used to be accepted as Factur-X too and blew up later in the import. It is now refused with `its container holds no invoice - No PDF attachment found`, and the flow is postponed instead of aborting the batch.

## Where such a document comes from — and where it does not

Measured against a live access point, so the origin of the case is not left to guess: converting a
UBL invoice to Factur-X there produces a PDF whose single attachment is `factur-x.xml` and whose
content is a `CrossIndustryInvoice`. A container that still carries UBL is therefore **not** what a
conversion produces; it is what an issuer's own hybrid document looks like, and that one reaches the
module untouched as the `Original` — the shape read whenever `Converted` is unusable, or first with
`EINVOICING_PREFER_ORIGINAL`.

Related to #718, without settling it: the failure this fixes carries no `postponeflow`, so it stops
the batch exactly the way that issue describes. The **side benefit** above is the part likely to be
met in the wild first — a flow whose `Converted` and `Original` are both in a syntax with no reader
falls through to the `ReadableView`, an ordinary PDF, which main still accepts as Factur-X and then
fails on with no postponement.

## Non-regression

- **PHPUnit**, file by file, on every supported core (18 → 24): `OK=23 KO=0` on each.
- **QA generation matrix**, 7 invoice types × CII and FACTURX = 14 generations: all produced and validated. Expected — this patch touches reception only.
- **Full lifecycle E2E**, every supported version, both directions: **14/14 cycles complete**, each recording `200, 201, 205, 211, 212` — emitted, routed, imported, approved, paid, cashed in. Run against this branch's head on a tree checked to be unchanged before and after.
- Dolibarr 17 shows 2 errors, but the module declares `need_dolibarr_version = array(18, -3)` and those same 2 errors are there with this branch's file reverted to its parent on the same bench.

### DIFF
--- a/einvoicing/class/providers/AbstractPDPProvider.class.php
+++ b/einvoicing/class/providers/AbstractPDPProvider.class.php
@@ -526,6 +526,15 @@ abstract class AbstractPDPProvider
 				continue;
 			}
 
+			// What was recognized above is the syntax of the file, not the syntax of the invoice it
+			// carries. Open the document before accepting it, or a syntax with no reader here is
+			// handed to the reader of another one.
+			$payloadReason = '';
+			if ($this->readableInvoicePayload($content, $protocol, $protocolManager, $payloadReason) === null) {
+				$result['attempts'][] = $docType . ": " . $protocolName . " but " . $payloadReason;
+				continue;
+			}
+
 			if ($docType != 'Converted') {
 				dol_syslog(__METHOD__ . " No usable 'Converted' document for flowId " . $flowId . " (" . implode(' | ', $result['attempts']) . "), reading the '" . $docType . "' one instead", LOG_WARNING, 0, "_einvoicing");
 			}
@@ -538,6 +547,57 @@ abstract class AbstractPDPProvider
 		}
 
 		return $result;
+	}
+
+	/**
+	 * The invoice a document carries, once its container has been opened, when this module can read it.
+	 *
+	 * detectProtocolFromContent() answers on the outside of a file: a PDF is 'FACTURX' on the strength
+	 * of its '%PDF-' header alone, whatever it holds. What it holds is a separate question, because the
+	 * embedded invoice is picked by file name and not by content - 'xrechnung.xml' is one of the names
+	 * accepted, and an XRechnung is commonly UBL, a syntax that has no reader here.
+	 *
+	 * Reading the outside only is how such a payload reaches the reader of another syntax. The CII
+	 * reader then finds none of the elements it expects and reports a business error that names nothing
+	 * the user can act on, which is counted as a failure and stops the whole synchronization batch; and
+	 * on the consistency check, the same document reaches code that dereferences a protocol which was
+	 * never built.
+	 *
+	 * The document is not re-typed on what is found inside: a Factur-X PDF is imported as Factur-X,
+	 * whose reader does the extraction itself. Only the question "can this be read at all" is answered.
+	 *
+	 * @param	string				$content			Raw document, as the access point returned it
+	 * @param	AbstractProtocol	$protocol			Protocol recognized on the container
+	 * @param	ProtocolManager		$protocolManager	Protocol factory used to recognize the payload
+	 * @param	string				$reason				Filled with why the payload was rejected
+	 * @return	?string									The invoice inside, or null when it cannot be read
+	 */
+	protected function readableInvoicePayload($content, $protocol, $protocolManager, &$reason)
+	{
+		$reason = '';
+
+		try {
+			// AbstractProtocol::extractXmlFromFileContent() returns the content unchanged, so a
+			// document that is already an XML is its own payload and this costs nothing there.
+			$payload = (string) $protocol->extractXmlFromFileContent($content);
+		} catch (Throwable $e) {
+			// A PDF with no embedded invoice at all raises ZugferdNoPdfAttachmentFoundException here.
+			$reason = "its container holds no invoice - " . $e->getMessage();
+			return null;
+		}
+
+		$payloadName = $protocolManager->detectProtocolFromContent($payload);
+		if (empty($payloadName)) {
+			$reason = "what it carries is in an unrecognized syntax";
+			return null;
+		}
+
+		if (empty($protocolManager->getProtocol($payloadName))) {
+			$reason = "what it carries is " . $payloadName . ", which is not supported";
+			return null;
+		}
+
+		return $payload;
 	}
 
 	/**
@@ -1122,7 +1182,16 @@ abstract class AbstractPDPProvider
 		if ($resProtocol['success']) {
 			$protocol = $resProtocol['protocol_object'];
 
-			$xmlData = $protocol->extractXmlFromFileContent($receivedFileContent);
+			// getProtocolFromContent() succeeded on the container. The invoice inside may still be in a
+			// syntax with no reader here, and everything downstream - cleanXmlData() first - assumes a
+			// protocol was found for what it is handed. Return nothing rather than that.
+			$payloadReason = '';
+			$xmlData = $this->readableInvoicePayload($receivedFileContent, $protocol, new ProtocolManager($this->db), $payloadReason);
+
+			if ($xmlData === null) {
+				dol_syslog(__METHOD__ . " Flow " . $flowId . " holds no invoice this module can read: " . $payloadReason, LOG_WARNING, 0, "_einvoicing");
+				return null;
+			}
 
 			if ($cleanXml) {
 				$xmlData = Document::cleanXmlData($xmlData);


einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -540,6 +549,57 @@ protected function fetchImportableFlowDocument($flowId, $protocolManager)
...
+	 * @return	?string									The invoice inside, or null when it cannot be read
+	 */
+	protected function readableInvoicePayload($content, $protocol, $protocolManager, &$reason)
+	{
+		$reason = '';
+
+		try {
+			// AbstractProtocol::extractXmlFromFileContent() returns the content unchanged, so a
+			// document that is already an XML is its own payload and this costs nothing there.
+			$payload = (string) $protocol->extractXmlFromFileContent($content);
```
eldy, line 582: This means that if the converted format was not forced to CII on AP setup, we may receive for "Converted", a PDF document that contain a UBL inside ?

daGrumpf-bxp: Not from a conversion — measured on the access point's own converter: `POST /v1.beta/invoices/convert?from=ubl&to=factur-x` returns a PDF whose single attachment is `factur-x.xml` and whose content is a `CrossIndustryInvoice`. A Factur-X *built by conversion* does not carry UBL.

If the conversion is not set to CII/Factur-X, what arrives as `Converted` is a plain UBL **XML**, not a PDF — that is hgy29's case in #718, already handled by #751 (flow postponed, batch carries on).

The PDF-carrying-UBL is the **issuer's own hybrid** document — a German XRechnung, embedded attachment named `xrechnung.xml` — and it reaches us as the `Original`: the shape read whenever `Converted` is unusable, or first with `EINVOICING_PREFER_ORIGINAL`.

Worth knowing for the setup question: we cannot ask for a syntax. The AFNOR Flow API `GET /v1/flows/{flowId}` takes `docType` only (`Metadata|Original|Converted|ReadableView`), no format parameter, and the account setting that decides it is not exposed by the API either — it is set in the access point's own UI.

So the guard is still needed, just not on the path your question names: not "the Converted was not forced to CII", but "the container we opened is not the one the conversion produced".

eldy: Can you confirm me that if format is correctly set to CII on AP (the recommended value), even if sender has sent a Fatur-x with UBL inside, we will never call the part of code managed by the zugferd lib ?  (we must avoid this absolutely, as zugferd hangs on a lot of environment due to Dolibarr version conflict, autolib conflict with external module, PHP of php version not supported, ...)
I understand that if AP is set to CII, there is a converted value that is ok, and the readable file will be PDF (and we should never need to parse the reabable file, it should be only savedasis), but i want to be sure.

daGrumpf-bxp: Measured on the live access point rather than reasoned about — I sent a real hybrid (a PDF whose embedded invoice is UBL) from one sandbox account to the other, with the recipient account set to CII, then set to Factur-X, and read every `docType` back. Probe used: `get_declared_classes()` filtered on `horstoeko\zugferd`, with a negative control proving it does catch the library when it is used. (An `spl_autoload_register` spy is useless here — Composer re-registers itself in front of it.)

**Your question, on the import path: yes, confirmed.** Account set to CII, flow `i_443316`:

```
Converted    : 18159 bytes, CII (CrossIndustryInvoice)   <- picked
Original     : 41554 bytes, PDF, sha256 identical to what the sender deposited, UBL inside
ReadableView : 37779 bytes, PDF built by the AP, CII inside

picker  : doc_type='Converted' protocol=CIIProtocol      syncFlow: res=1, invoice created
zugferd : not one class loaded — FacturXProtocol.class.php is not even included
```

Same on `main` and on this branch. So with the recommended setting the reception path never goes near the library.

**But the setting does not cover two other cases, both measured on the same specimen.**

**1. `fetchFlowXml()` asks for the `Original`, hardcoded.** So the consistency check gets the issuer's hybrid PDF whatever the account is set to, and loads zugferd there — 3 classes, on both trees. On `main` that call dies (`get_class(): Argument #1 must be of type object, null given`); with this branch it returns `null`. This is the caveat I flagged at the end of the PR body, and it is the one place your rule is broken today by design.

**2. `Converted` is not always converted.** Account set to Factur-X, same specimen, flow `i_443401`:

```
Converted    : 41546 bytes, PDF, sha256 IDENTICAL to the deposited file, UBL inside
Original     : same file
ReadableView : 37660 bytes, PDF built by the AP, CII inside
```

The access point converts only when the incoming syntax **differs** from the target. A flow that arrives already Factur-X is passed through byte for byte, whatever it carries. And it labels the flow from the container exactly the way `detectProtocolFromContent()` does: the same file deposited with `flowSyntax: UBL` declared came back as `flowSyntax: "Factur-X"` and was passed through all the same (`i_443417`).

Control, to be sure the platform does convert: the same UBL deposited **bare**, account still on Factur-X (`i_443409`) → `Converted` is a real Factur-X built by the AP, carrying CII. So it is the "already Factur-X" case that is never opened.

What that costs, on that Factur-X-configured account, same flow, Dolibarr 23.0.4:

```
main       picker takes Converted as FACTURX -> import res=-1
           'Unfounded dolibarr corresponding Invoice code for document type code: NA'
           = issue #761 on the ordinary path, with no EINVOICING_PREFER_ORIGINAL involved

this PR    Converted: FACTURX but what it carries is UBL, which is not supported
           Original : same
           -> falls through to ReadableView (built by the AP, CII inside), import res=1, invoice created
```

So: CII on the access point is the right recommendation and it does keep the library out of the import path — but it is a setting that lives outside Dolibarr, a customer can legitimately have set Factur-X (Etrema did, cf. #742), and in that case the container reaches the reader unopened. That is what the guard is for.

On the readable view: on a CII account it is downloaded and stored as an attachment, never parsed — the run above proves it. It is parsed only when neither `Converted` nor `Original` is readable, which is the loop #751 introduced, not something added here; in the Factur-X case above that is exactly what rescues the document.

PR #774, opened 2026-09-04
FIX: 500 error on first synchronization (epoch date sent to API)

lelex86:
## Summary

- Problem: The first synchronization (manual or cron) failed with a 500 error. getLastSyncDate() returned 0 when no sync history existed, causing syncFlows() to call the PDP API with updatedAfter: 1970-01-01T00:00:00.000Z. Most providers reject this date.

- Fix in AbstractPDPProvider: When no history exists in the database, getLastSyncDate() now calls getFirstUseDate(), which returns a meaningful start date through a 3-level cascade: ① date the PDP provider was configured (tms of EINVOICING_PDP in llx_const) → ② date the module was activated (tms of MAIN_MODULE_EINVOICING) → ③ French e-invoicing legal mandate date (2026-09-01). Each step is logged to _einvoicing.

- Fix in document_list.php: Two corrections to the manual sync form: (1) the date picker is now pre-filled via getLastSyncDate() on first sync so the field is never empty, (2) the validation used === '' which did not catch the integer 0 returned by GETPOSTINT — corrected to !$syncfromdate.

### DIFF
--- a/einvoicing/class/providers/AbstractPDPProvider.class.php
+++ b/einvoicing/class/providers/AbstractPDPProvider.class.php
@@ -796,7 +796,8 @@ abstract class AbstractPDPProvider
 	/**
 	 * Get the last synchronization date with the PDP provider.
 	 * Retrieves the timestamp of the most recent successful flow synchronization
-	 * for this provider. If no sync has occurred yet, returns 0.
+	 * for this provider. If no sync has occurred yet, falls back to the module
+	 * commissioning date (see getFirstUseDate()).
 	 * Optionally applies a margin in hours to the returned timestamp.
 	 *
 	 * @param 	int 		$marginHours 	Optional time margin in hours to go back from the current date of the last synchronization
@@ -837,8 +838,9 @@ abstract class AbstractPDPProvider
 			dol_syslog(__METHOD__ . " SQL warning: Failed to get last sync date: we try to sync all flows from today", LOG_WARNING);
 		}
 
+		// No sync history yet — use the commissioning date so we never call the API with epoch (1970).
 		if ($LastSyncDate === null) {
-			$LastSyncDate = 0;
+			$LastSyncDate = $this->getFirstUseDate();
 		}
 
 		// Apply margin in hours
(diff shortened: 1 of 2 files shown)


daGrumpf-bxp:
Tested this branch against `main` on twin instances (Dolibarr 23.0.4, same DB copy with `llx_einvoicing_document` emptied = real "first sync" state). The 3-level cascade does work exactly as documented, and the date picker is correctly pre-filled. But two measured findings, one of them blocking:

1. **`tms` is the last write, not the creation date.** `dolibarr_set_const()` with an *unchanged* value — one click on Save in the setup page — moves it: `EINVOICING_PDP` tms went from `2026-09-03 13:53:27` to `2026-09-04 09:46:05` in my run. Same for `MAIN_MODULE_EINVOICING` on every module re-activation.
2. **So the fix can silently skip flows.** With that bumped tms, `getLastSyncDate()` returned today, `flows/search` answered `0 results`, while the same account holds flows since `2026-06-21T23:00:53Z`. As `updatedAfter` only moves forward, those flows are never fetched again. Today's epoch call at least returns everything; here the operator gets a clean "synchronization successful, 0 new flows" and loses the backlog with no error at all.
3. **Regression in `document_list.php`.** Since `getLastSyncDate()` no longer returns 0, the guard at line 947 now fires on a first sync: picking *today* is refused with *"The synchronization start date cannot be after the date of the last synchronized flow"* — while no flow has ever been synchronized. Measured: `main` builds the confirm URL, this branch shows the error instead. Legitimate case: module installed months ago, operator only wants yesterday onwards.

Suggestion: an explicit floor (a config option, or `now - N days`) rather than a `tms` that drifts, and leave `getLastSyncDate()` returning 0 so the guard above stays off — the `!$syncfromdate` validation fix is right and can stay on its own.

Side note on the premise: could you tell which PDP returned the 500, and its response body? On SuperPDP `updatedAfter=1970-01-01T00:00:00.000Z` answers `HTTP 200` (measured), and both providers already handle a non-200 gracefully (`res=0` + message), so a rejected date does not produce a Dolibarr 500 — the root cause may be elsewhere.

Minor: `foreach` over a one-element `$constNames`; `isset($provider)` is redundant inside `if ($provider)`; `$db->jdate()` would be more idiomatic than `strtotime()` on a DB datetime; and level 3 is unreachable in practice (`MAIN_MODULE_EINVOICING` always exists when the module is enabled) — a hardcoded FR-specific date in a class shared by all PDPs.

eldy:
I agree with using tms. Dangerous (i can be updated, automatically, without user consent, at any time).
Why not just using 
`min(01/01/2026,  tms of the record in llx_const) `
?

We just need an old date that is not 0 the first time.

lelex86:
Thanks for the feedback. The client's initial report pointed us in the wrong direction — after getting access to the actual server logs, we identified the real root cause (a missing require_once for company.lib.php in cron context, causing a getCountry() fatal error).

The fix is in #780. This PR can be closed.

PR #793, opened 2026-09-04
FIX #791 Deleting the default product reference

### LINKED ISSUE #791: [einvoicing] Deleting the default product reference (was https://github.com/Dolibarr/dolibarr/issues/40103#issue-5348946604)
### Bug

In a supplier’s card, it is not possible to delete the default product reference used by einvoicing module. But it can be edited.

Module 1.0.3 -> 1.2.0

### Module

einvoicing

### Dolibarr Version

23,24

### Environment PHP

_No response_

### Environment Database

_No response_

Link to french forum : https://www.dolibarr.fr/forum/t/module-officiel-e-invoicing-ex-pdpconnectfr-facture-electronique-update-v-1-1-0/51552/228?u=philazerty

### Attached files

_No response_

emheyarssi:
If the `$routingProductId` is blank, the default product should also be updated to blank.

### DIFF
--- a/einvoicing/class/einvoicing.class.php
+++ b/einvoicing/class/einvoicing.class.php
@@ -2333,6 +2333,9 @@ class EInvoicing
 			$resprints .= '<td>' . $form->textwithpicto($langs->trans("DefaultProductEBilling"), $langs->trans("DefaultProductEBillingHelp")) . '</td>';
 			$resprints .= '<td'.(empty($parameters['colspanvalue']) ? '' : ' colspan="'.(((int) $parameters['colspanvalue']) - 1).'"').'>';
 			if ($mode == 'edit') {
+				if (preg_match('/^idprod/', $product_id)) {
+					$product_id = (int) str_replace('idprod_', '', $product_id);
+				}
 				$resprints .= $this->selectVendorProduct($form, $object->id, $product_id, 'routing_product_id');
 			} else {
 				if ($product_id != '' && $product_id != '-1') {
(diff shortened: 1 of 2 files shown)


daGrumpf-bxp:
Sorry @emheyarssi — I had not seen this PR and I opened #797 on the same issue. My apologies for the duplicate work.

Since both are on the table, here is what #797 does differently, so you can judge which one to keep:

* it also treats **`-1`** as a removal. The combo of `Form::select_produits_fournisseurs_list()` posts `-1` when its empty entry is picked (and `''` only in the ajax variant, `PRODUIT_USE_SEARCH_TO_SELECT`), so keeping `$routingProductId !== '-1'` leaves the plain combo — the default setup — unable to remove the product;
* about the `(int) str_replace('idprod_', '', $product_id)` before the combo: the number after `idprod_` is a **product** id, while `$selected` is compared to `idprodfournprice`, the id of a **supplier price** row, so the option would not be matched (and could match another product's price id). The real cause is on the core side: the `preg_match('/^idprod_/')` branch of that comparison only appeared in Dolibarr 22, so on 18 to 21 the combo shows its empty entry although a default product is set. #797 marks the option itself when the core did not;
* that second point matters for the removal too: with the combo showing nothing, **any plain save of the vendor card would drop the setting**. #797 carries what the field displays in a hidden `routing_product_id_shown` and only removes the routing when the user emptied a field that did show a value — which also covers a default product that is not among the products the combo lists (`PRODUIT_LIMIT_SIZE`);
* one detail of this PR worth checking either way: with an empty value and no routing yet, `addRouting()` returns `-1`, so the save reports `FailedToSaveRoutingID` although nothing was wrong.

#797 comes with `test/phpunit/DefaultProductRoutingResetTest.php` (7 tests) and was checked on Dolibarr 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4 and 24.0.1, over real HTTP as well as in PHPUnit.

Happy to close mine and move whatever is useful into yours if you prefer — your call.

daGrumpf-bxp:
Superseded by #797, merged: it covers both values the field posts when emptied (`-1` from the combo, `''` from the ajax variant, this one still skipping `-1`) and the `idprod_` display on cores below 22 — thanks for spotting the bug.

eldy:
Replaced in favor of #797

PR #796, opened 2026-09-04
QUAL #794 [einvoicing] An import made again keeps the line of the flow, the number of the draft, and the statuses already received

### LINKED ISSUE #794: Question on import doc
@daGrumpf-bxp About a feature you introduce (to solve pb of pad thirdparty import).

I import a supplier invoice.
Image

Then i try the button "Import doc again".
Now i have 
Image

Everything looks as expected.

But the question is why not reusing the same draft ref for ref-creating the new draft invoice ? Also as it is the same flow, why dropping the line in history to create a new one. Why not modifying the old one with a mention "(REF) Doc reimported by id XXX" on field ref and  adding another one ?

This should avoid headach to understand what happenwhen reading the flow list.

daGrumpf-bxp:
Fixes #794.

Answers #794, by doing what it asks: after an import made again, the flow keeps the line it had in
the list and the draft keeps the number it had. A third commit fixes what looking into it turned
up - the statuses already received stayed behind.

### 1. The flow keeps its line

The import writes a record of its own, as it does for any flow it reads. That record is now moved
onto the line the re-import was asked from, and the one the import wrote is removed instead. The
line keeps what makes it that line - its id, its creation date, its author - and takes everything
the access point said about the flow, the invoice it has just been booked on included.

Nothing ever forced one line per flow: `flow_id` carries a plain index, a synchronization skips what
it has already imported with `flow_id IN (...)`, and a vendor lifecycle status is attached to the
invoice by the reference the vendor gave it, never by the flow record. Keeping both lines was
possible too - one of them dead, detached from an invoice that no longer exists, next to the line
that describes the same flow for real.

### 2. The draft keeps its number

The module does not choose it: `FactureFournisseur::create()` inserts `'(PROV)'` and then rewrites
the row with `'(PROV'.$this->id.')'`, so a reference set by the caller beforehand is ignored (same
code from 18.0.10 to 24.0). And the draft has to be a new row - the import recognizes an invoice it
has already imported by (`ref_supplier`, `socid`) and hands it back untouched instead of resolving
the vendor again, which is exactly what a re-import must not do.

The number of the draft that has just been deleted is free, though, and no code of the core reads an
id back out of a temporary reference: it tests the `PROV` prefix, and skips such references when it
looks for the last number used. So it is given back to the invoice the import creates, with the
directory its documents are stored under and the index of those files - the two `ecm_files` updates
and the directory rename the core itself makes in `FactureFournisseur::validate()`. Only a temporary
reference is moved this way, and only onto a draft. The rename is best effort: if it fails, the
invoice keeps the number the core gave it and the import stands.

### 3. The statuses already received follow the invoice

Found while measuring the two above, and visible in the second screenshot of #794: the
`SupplierInvoiceLC` line of the invoice stays behind. Deleting the draft detaches every e-invoicing
record it has - the received invoice and the statuses that came after it - and the lifecycle messages
keep pointing at the row that has just disappeared.

Measured on 23.0.4, on a document received from the sandbox access point: after an import made
again, the LC line was linked to nothing and the status 202 it carried was still attached to the
deleted draft, so the new draft came up with no status at all. A status a vendor sends is about the
invoice the vendor issued, not about the row Dolibarr booked it on, so those records - listed before
the draft is deleted, since a detached record no longer says which invoice it came from - now follow
onto the invoice the import creates.

### Measured

A document really received through the sandbox access point (emitted by another instance, routed by
the platform, imported by a synchronization), then "Import doc again" on its flow:

| core | flow line | draft | LC line | status 202 |
|---|---|---|---|---|
| 23.0.4 | 6223 kept, now on invoice 9078 | 9077 deleted, 9078 created and named `(PROV9077)` | re-attached to 9078 | moved to 9078 |
| 18.0.10 | 6120 kept, now on invoice 8846 | 8845 deleted, 8846 created and named `(PROV8845)` | re-attached to 8846 | moved to 8846 |

Files follow the number in both cases (`fournisseur/facture/8/7/(PROV9077)` on 23.0.4,
`fournisseur/facture/6/4/(PROV8845)` on 18.0.10), with `llx_ecm_files` pointing at the directory
they are actually in. Read back from a second process, so nothing is left in an open transaction.
Running the action twice in a row on the same flow keeps the same line and the same number
throughout. Module test suite: 32/32 on 18.0.10.

### DIFF
--- a/einvoicing/class/document.class.php
+++ b/einvoicing/class/document.class.php
@@ -579,6 +579,7 @@ class Document extends CommonObject
 		// import that follows creates another one, which is given that same reference back.
 		$previousInvoiceId = (int) $this->fk_element_id;
 		$previousInvoiceRef = '';
+		$previousInvoiceRecords = array();
 		if ($previousInvoiceId > 0) {
 			$previousInvoice = new FactureFournisseur($this->db);
 			if ($previousInvoice->fetch($previousInvoiceId) > 0) {
@@ -586,6 +587,10 @@ class Document extends CommonObject
 					return array('res' => -1, 'message' => $langs->trans('EInvoiceReimportInvoiceIsNotADraft', $previousInvoice->ref));
 				}
 				$previousInvoiceRef = (string) $previousInvoice->ref;
+				// The other flows of that invoice - the lifecycle statuses its vendor has already sent -
+				// are listed before it goes: deleting it detaches every record it has at once, and a
+				// detached record no longer says which invoice it came from.
+				$previousInvoiceRecords = $this->fetchOtherRecordsOfInvoice($previousInvoiceId);
 				if ($previousInvoice->delete($user) <= 0) {
 					$reason = $previousInvoice->error ? $previousInvoice->error : implode(', ', (array) $previousInvoice->errors);
 					return array('res' => -1, 'message' => $langs->trans('EInvoiceReimportFailedToDeleteTheDraft', $previousInvoice->ref).' '.$reason);
@@ -618,6 +623,12 @@ class Document extends CommonObject
 			if ($newInvoice->fetch($newInvoiceId) > 0 && $this->reuseDraftRef($newInvoice, $previousInvoiceRef) > 0) {
 				$importedRecord->tracking_idref = $previousInvoiceRef;
 			}
+		}
+
+		// The statuses the vendor has already sent for this invoice describe the document, not the
+		// local row it was booked on, so they follow it onto the invoice the import has just created.
+		if ($newInvoiceId > 0 && $previousInvoiceId > 0) {
+			$this->moveHistoryToInvoice($previousInvoiceId, $newInvoiceId, $previousInvoiceRecords, (string) $importedRecord->tracking_idref);
 		}
 
 		// Move that record onto this line, so the flow keeps the line it already had in the list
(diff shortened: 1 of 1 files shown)


einvoicing/class/document.class.php
```diff
@@ -592,30 +603,311 @@ public function reimport(User $user)
...
+
+		$moved = false;
+		if (file_exists($dirsource)) {
+			if (file_exists($dirdest)) {
+				// Left over by something else: merging two directories is not this method's business.
+				dol_syslog(__METHOD__.' cannot reuse reference '.$wantedref.': '.$dirdest.' already exists', LOG_WARNING);
+
+				return 0;
+			}
+			if (!@rename($dirsource, $dirdest)) {
```
eldy, line 852: If you se dol_move_dir function, it seems don't need the rest of code (move in ecm_files, ...)

daGrumpf-bxp: Tried it: `dol_move_dir()` renames the directory first, then only reindexes the files whose *name* carries the old reference — so the imported XML (named after `ref_supplier`) is moved on disk but keeps its old `filepath`, and the one that is renamed gets a **second** `ecm_files` row instead of an update. Measured on 18.0.10, 20.0.4, 23.0.4 and 24.0, same on all four: 3 rows afterwards, 2 of them pointing at files that are no longer there.

And `FactureFournisseur::validate()` does this very rename without `dol_move_dir()` — two `ecm_files` updates, `@rename()`, then the loop over the prefixed names; this method is that code, on purpose.

Happy to switch if you would rather have `dol_move_dir()` fixed first.

Issue #799, opened 2026-09-04
Failed to send CDAR file to PDP. Status code: 400. Message: B2BInt invoices only support payment received event


mdeweerd:
### Bug

II synchronised invoices.  These are from OVH, and are paid later with direct debit.

I validated, selected "approve".
<img width="164" height="94" alt="Image" src="https://github.com/user-attachments/assets/0954c8b8-ce0f-4a44-8db2-79d4a8040898" />

Then got this error:
<img width="445" height="31" alt="Image" src="https://github.com/user-attachments/assets/4bacbfe0-683c-45f9-9c64-f28b59af14d4" />

Maybe 2 issues:
- Can not "Approve";
- Can/should not "Pay" - so "Paiement transmis" should not show ?  (Direct debit later).

### Module

einvoicing

### Steps to reproduce the behavior and expected behavior

Try to Approuve OVH supplier invoice under direct debit.

daGrumpf-bxp:
Thanks. There are two separate things in this report, and the first one needs your help before I can conclude.

**Where the message comes from.** `B2BInt invoices only support payment received event` is not our wording: the module relays the body of the platform's HTTP 400 verbatim. Our own half is `Failed to send CDAR file to PDP. Status code: 400. Message: ...`, and that half is byte-identical in `SuperPDPProvider` and `EsalinkPDPProvider` (and any provider an external module may ship as a copy) — so the message as quoted does not even tell us which platform refused. That part is fixed on my side: the message will name the provider.

**What the platform is saying.** The lifecycle statuses 200-213 belong to the domestic B2B e-invoicing circuit. A flow the platform classifies as *B2B International* falls under e-reporting instead: no lifecycle exchange between platforms, and the only event expected there is the cash-in (212 "Encaissee" = payment received), reported by the seller. So refusing a 205 "Approved" sent by the buyer is consistent — **if** the flow really is B2BInt. That is what I cannot tell from here: OVH is French, so if your company is French too I would expect a domestic B2B flow, and then the classification itself is the problem (most likely how the buyer is identified in the invoice), not your click.

**What I need to settle it:**

1. **Which access point (PDP/PA)** you are on — the value of `EINVOICING_PDP`. From #798, opened the same evening, I read SuperPDP, but please confirm: as explained above, this message cannot say it.
[...]

PR #805, opened 2026-09-05
NEW [einvoicing] Export the selected flows or API calls for support

daGrumpf-bxp:
## The problem

Instructing a refusal of the platform means, today, walking the user through four screens — the flow
(stored XML, syntax, profile), the API call it came from (endpoint, request body, response body), the
setup constants, the versions — and having them turn an option on, because the two most useful
columns are only written when `EINVOICING_DEBUG_MODE` is set:

- `llx_einvoicing_call.response`
- `llx_einvoicing_document.response_for_debug`

On #799 that back and forth has cost several days and the deciding information is still missing. The
idea of a mass action for it is @mdeweerd's, in that same issue.

## What this adds

One entry, **Export for support**, in the mass action combo of `document_list.php` and
`call_list.php`. Both pages already carry the whole mass action plumbing (`$massaction`, `$toselect`,
`$diroutputmassaction`, `actions_massactions.inc.php`, `$arrayofmassactions`), so nothing is wired
here — only the entry, its confirmation, and the call.

Tick the rows, pick the entry, confirm, and the browser is handed one archive:

```
manifest.json                     the context: core, module build (einvoicingModuleStamp()), PHP,
                                  provider, protocol, entity, date, counts, whether debug mode was
                                  on, and every EINVOICING_* constant
flows/<flow_id>/flow.json         the row, bodies left out
flows/<flow_id>/document.xml      document_body, or xml_data when there is no document_body
flows/<flow_id>/metadata.json     response_for_debug, when it was written
calls/<call_id>/call.json         the row, request_body and response included
```

`debug_mode` is in the manifest on purpose: without it a reader cannot tell *a call that answered
nothing* from *a call whose answer was never stored*.

## What never leaves the instance

- Everything written goes through `AbstractPDPProvider::redactSensitiveData()` — the very method the
  call log is written with, made `public static` for this second caller. The log is already written
  redacted, so this is defence in depth: rows written *before* that redaction existed still hold
  what the platform answered, tokens included, and an export is exactly what would take them out.
- A constant whose name carries `SECRET`, `TOKEN`, `REFRESH`, `PASSWORD`, `PASSWD`, `CLIENT_ID`,
  `API_KEY`, `CREDENTIAL` or `PRIVATE` is reported as `[SET, NOT EXPORTED]` / `[NOT SET]`, never
  quoted. That covers the setup credentials and the OAuth tokens the module stores as
  `<PREFIX>_PROD_TOKEN` / `_REFRESH` on cores below 23.

The invoice XML **is** in the archive — that is the point, it is the document to diagnose — so it
carries the names, addresses, e-mail addresses and phone numbers of the parties. The confirmation
screen says so in one sentence before anything is built.

## Two decisions worth reviewing

**The archive is handed over straight away, not left in the "generated documents" box.** That box
downloads through `document.php` with the modulepart `massfilesarea_einvoicing`, and
`dol_check_secure_access_document()` grants that modulepart on `hasRight(<module>, 'lire')` up to
Dolibarr 21 — it moved to `'read'` in 22. The module declares `read`, like every module builder
module, so the link is refused on the lower half of the supported range. Measured on the bench, with
a plain user carrying only `einvoicing → read` (an admin passes everywhere and hides this):

| core | 18 | 19 | 20 | 21 | 22 | 23 | 24 |
| `massfilesarea_einvoicing` download | refused | refused | refused | refused | served | served | served |

Streaming answers the same on all seven. The file is built under `$diroutputmassaction` as any mass
action would, and removed once sent, which also keeps archives full of invoice XML from piling up in
the temporary area.

**The confirmation is the non-ajax `formconfirm()`,** like the core's own `predelete`: the ajax
variant jumps to a GET url and would drop `toselect[]`.

Two guards: `ZipArchive` is checked with `class_exists()` and answers a plain message rather than
letting `dol_compress_dir()` report no handler; and a selection wider than
`EINVOICING_SUPPORT_EXPORT_MAX` (50 by default — an e-invoice XML weighs 100 to 200 kB) is refused
with a message naming the bound, rather than met with a timeout or a memory limit.

The build lives in `class/utils/SupportExport.class.php`, not in the pages: page code cannot be
exercised by PHPUnit, and the redaction is exactly what has to be covered by a test.

## What was measured

Bench: Dolibarr **18.0.10** and **24.0.1**, PHP 8.4, module deployed by symlink, real HTTPS sessions.

**Ran.**

- `SupportExportTest.php`, 6 tests / 62 assertions, green on 18 and 24. It plants an
  `access_token` + a `client_secret` in a call log row written *without* redaction — the shape of a
  row predating it — exports, reopens the archive and asserts neither string comes out while
  `[REDACTED]` does; it checks the manifest, the secret constants, a flow with and without a
  document, the bound, and a row that cannot be reached.
- The whole PHPUnit suite of the module, **34 files**, green on 18 and 24.
- The mass action itself over HTTP, on both lists, for an **admin** and for a **plain user carrying
  only `einvoicing → read`**: 80 checks each, all green, on 18 and 24 — the entry is in the combo,
  the confirmation appears and says what leaves, the answer is `application/zip` with a
  `Content-Disposition` naming `einvoicing-support-<entity>-<date>.zip`, the archive reopens, its
  manifest counts exactly the selection, there is one row file per selected record, and no
  `access_token` / `client_secret` / `refresh_token` in clear anywhere in it.
- Cases covered: 1 flow with its XML, 1 flow with no document at all (no `document.xml`, no failure),
  3 flows mixed, 1 API call with its response, 2 API calls — each with `EINVOICING_DEBUG_MODE` **off**
  and **on** (the manifest follows).
- 61 rows selected → no archive, and the message naming the bound and the constant.
- A user without `einvoicing → read`: the two lists are refused, the entry is absent, and posting
  `action=supportexport` straight at the page answers **403**.
- `dolics` (phpcs 3.x + the ruleset of the repository) clean on the seven files, and `phan 5.5.2`
  with the configuration and baseline of the repository clean as well.

`MoreSpecificElementTypePlugin` wants the documented return type of five methods to repeat the
union of the literal value types phan infers for one run, or to call an array filled in a loop
non-empty. Those five carry a `@phan-suppress` with the reason; the one real finding it raised,
an unprotected variable in the `IN (...)` of `reachableIds()`, is fixed with `$db->sanitize()`
(the ids are cast to int before, so this is belt and braces).

**Read, not run.** The `massfilesarea` table above is measured, but the reading of
`dol_check_secure_access_document()` that explains it (`$lire` up to 21, `$read` from 22) is a
reading of the core sources of the seven versions. Cores 19 to 23 were only used for that
measurement; the mass action itself was exercised on 18 and 24.

## Not fixed by this

#799 is the use case this comes from — a refusal `B2BInt invoices only support payment received
event`. This export is what lets such a refusal be instructed; it does not correct it. The issue
stays open.

No ChangeLog entry, per `.agents/AGENTS.md`. Language keys in `en_US` only.

### DIFF
--- a/einvoicing/call_list.php
+++ b/einvoicing/call_list.php
@@ -258,6 +258,29 @@ if (empty($reshook)) {
 
 	// You can add more action here
 	// if ($action == 'xxx' && $permissiontoxxx) ...
+
+	// Mass action of the module: pack the selection into one archive a maintainer can read without
+	// walking the user through four screens (issue #799). The archive is built by a class of its
+	// own, because page code cannot be exercised by PHPUnit and the redaction it performs has to be
+	// covered by a test; the page only reads the selection and hands the file over.
+	if (empty($error) && ($massaction == 'supportexport' || ($action == 'supportexport' && $confirm == 'yes')) && $permissiontoread) {
+		dol_include_once('einvoicing/class/utils/SupportExport.class.php');
+
+		$supportexport = new SupportExport($db);
+		$supportarchive = $supportexport->build(SupportExport::TYPE_CALL, $toselect, $diroutputmassaction);
+
+		if ($supportarchive == '') {
+			setEventMessages($supportexport->error, null, 'errors');
+			$massaction = '';
+			$action = 'list';
+		} else {
+			// Nothing has been printed yet at this point of the page, so the archive can be sent
+			// as the answer to this very request.
+			SupportExport::deliver($supportarchive);
+			$db->close();
+			exit;
+		}
+	}
 }
 
 
@@ -521,10 +544,13 @@ $arrayofmassactions = array(
 	//'builddoc'=>img_picto('', 'pdf', 'class="pictofixedwidth"').$langs->trans("PDFMerge"),
 	//'presend'=>img_picto('', 'email', 'class="pictofixedwidth"').$langs->trans("SendByMail"),
 );
+if (!empty($permissiontoread)) {
+	$arrayofmassactions['presupportexport'] = img_picto('', 'download', 'class="pictofixedwidth"').$langs->trans("EInvoicingSupportExport");
+}
 if (!empty($permissiontodelete)) {
 	$arrayofmassactions['predelete'] = img_picto('', 'delete', 'class="pictofixedwidth"').$langs->trans("Delete");
 }
-if (GETPOSTINT('nomassaction') || in_array($massaction, array('presend', 'predelete'))) {
+if (GETPOSTINT('nomassaction') || in_array($massaction, array('presend', 'predelete', 'presupportexport'))) {
 	$arrayofmassactions = array();
 }
 $massactionbutton = $form->selectMassAction('', $arrayofmassactions);
@@ -562,6 +588,14 @@ $modelmail = "call";
 $objecttmp = new Call($db);
 $trackid = 'xxxx'.$object->id;
 include DOL_DOCUMENT_ROOT.'/core/tpl/massactions_pre.tpl.php';
+
+// Confirmation of the support export, in the shape the core uses for its own mass actions: no ajax
+// and no form tag of its own, so that the buttons submit the list form and the selection survives.
+// The ajax variant jumps to a GET url instead, which would drop toselect[].
+if ($massaction == 'presupportexport') {
+	$formquestion = array('text' => $langs->trans("EInvoicingSupportExportPrivacy"));
+	print $form->formconfirm($_SERVER["PHP_SELF"], $langs->trans("EInvoicingSupportExport"), $langs->trans("EInvoicingSupportExportQuestion", count($toselect)), "supportexport", $formquestion, '', 0, 250, 500, 1);
+}
 
 if ($search_all) {
 	$setupstring = '';
--- a/einvoicing/class/providers/AbstractPDPProvider.class.php
+++ b/einvoicing/class/providers/AbstractPDPProvider.class.php
@@ -922,10 +922,14 @@ abstract class AbstractPDPProvider
 	 * persisted in the API call log, whatever its shape: PHP array, application/x-www-form-urlencoded
 	 * string (OAuth token requests), JSON string, or plain text (left untouched in that last case).
 	 *
+	 * Public because the redaction has a second caller outside the write path: SupportExport
+	 * replays it on everything it puts in a support archive, so that rows written before this
+	 * method existed do not carry their tokens out of the instance.
+	 *
 	 * @param  array<mixed>|string|null $value Value to redact
 	 * @return array<mixed>|string|null Redacted value, same shape as the input
 	 */
-	private static function redactSensitiveData($value)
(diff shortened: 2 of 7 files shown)


eldy:
As this is a newfeature of UX, can you provide a screenshot of what appaears ?

daGrumpf-bxp:
Here they are — Dolibarr **24.0.1**, this branch, English session, an ordinary user list.

**1. The entry in the mass action combo of the flow list.** It appears where every mass action of the
page appears, once at least one row is ticked (3 flows here).

Mass action combo of the flow list, showing Export for support

**2. The confirmation.** One sentence says what leaves the instance before anything is built, and the
question carries the count of selected records. The default answer is **No**; the archive is only
built on **Yes**.

Confirmation screen of Export for support on the flow list

**3. The same entry on the API call list** (*Calls & Sync logs*). Its confirmation is the same screen,
with the number of selected calls.

Mass action combo of the API call list, showing Export for support

**4. The guard on the size of the selection** — 60 ticked, bound at 50. It names the bound and the
constant rather than letting the build time out.

Message when more records are selected than the bound allows

Nothing else shows: answering **Yes** hands the browser the archive itself
(`Content-Disposition: attachment; filename=einvoicing-support-<entity>-<date>.zip`) and the page
stays where it is. The archive produced by screenshot 2 above:

```
manifest.json
flows/i_451192/document.xml
flows/i_451192/flow.json
flows/ie_1334985/flow.json
flows/ie_1334953/flow.json
```

and its manifest, as written — the two credentials of the setup are named but not quoted:

```json
{
    "export_type": "flow",
    "generated_at": "2026-09-05 11:26:06 GMT",
    "dolibarr_version": "24.0.1",
    "einvoicing_version": "1.2.0 (8b87481)",
    "php_version": "8.4.25",
    "entity": 1,
    "provider": "SUPERPDP",
    "protocol": "CII",
    "debug_mode": false,
    "requested_count": 3,
    "exported_count": 3,
    "skipped_count": 0,
    "records": ["ie_1334953", "i_451192", "ie_1334985"],
    "constants": {
        "EINVOICING_AUTO_SEND_ON_GENERATION": "0",
        "EINVOICING_EINVOICE_IN_REAL_TIME": "1",
        "EINVOICING_LIVE": "0",
        "EINVOICING_PDP": "SUPERPDP",
        "EINVOICING_PROTOCOL": "CII",
        "EINVOICING_SUPERPDP_CLIENT_ID": "[SET, NOT EXPORTED]",
        "EINVOICING_SUPERPDP_CLIENT_SECRET": "[SET, NOT EXPORTED]",
        "EINVOICING_SUPERPDP_GRANT_TYPE": "client_credentials",
        "EINVOICING_SUPERPDP_ROUTING_ID": "000000002",
        "EINVOICING_THIRDPARTIES_AUTO_GENERATION": "1",
        "EINVOICING_THIRDPARTIES_COMPLETE_INFO": "1"
    }
}
```

(the constant list is cut here for the comment — the real file holds all the `EINVOICING_*` of the
instance.)

PR #809, opened 2026-09-05
FIX einvoicing: the prepaid amount follows what the core counts as paid

daGrumpf-bxp:
The module rebuilt by hand what an invoice has already been paid, and its rule was narrower than
the core's. An over-payment converted into a discount and applied to another invoice was invisible
to the document: **the e-invoice asked for more money than Dolibarr said was due.**

The core has one definition, identical from 18 to 24 in `core/class/commoninvoice.class.php`:

```php
public function getRemainToPay($multicurrency = 0)
{
    $alreadypaid  = $this->getSommePaiement($multicurrency);
    $alreadypaid += $this->getSumDepositsUsed($multicurrency);
    $alreadypaid += $this->getSumCreditNotesUsed($multicurrency);
```

`DiscountAbsolute::getSumCreditNotesUsed()` does not filter on the description: it joins on the
**type of the source invoice** (STANDARD, CREDIT_NOTE, SITUATION). It therefore counts the discount
born from an over-payment, which the core writes with the description `(EXCESS RECEIVED)`
(`compta/facture/card.php`, `api_invoices.class.php`) and applies through
`DiscountAbsolute::link_to_invoice(0, $id)` — setting `fk_facture` and creating **no line**, so the
line-walking path of the module did not see it either.

`buildinvoicelines.inc.php` filtered on `description = '(CREDIT_NOTE)'` instead.

Measured on a bench, invoice of 240.00 incl. tax carrying a 30.00 `(EXCESS RECEIVED)` discount:

| | core `getRemainToPay` | BT-113 | BT-115 |
|---|---|---|---|
| before, on 18 and on 24 | 210.00 | **0.00** | **240.00** |
| after, on 18 and on 24 | 210.00 | 30.00 | 210.00 |

Control with a `(CREDIT_NOTE)` discount of the same amount: 30.00 / 210.00 before and after — so
the description filter was the cause, and nothing else.

No double counting: a row of `societe_remise_except` carries `fk_facture_line` **or** `fk_facture`,
never both (`link_to_invoice()` refuses both parameters together). A deposit consumed as an invoice
line, which the module already handles, is therefore not counted twice; the test asserts it.

The referenced document (BT-25) now takes its type from the source invoice — 380, 381 or 386 —
instead of always announcing 381, which told the recipient that an invoice they may have paid was
a credit note.

Tested: PHPUnit 34/34 on each of the seven cores.

---

This PR stands alone on `main`. It is one of eleven opened together, each on its own subject; every
pair of them, and the whole set in either order, merge onto `main` with no conflict, and the fully
merged tree passes 44/44 on the seven cores. They can be reviewed and merged in any order, or
individually.

### DIFF
--- a/einvoicing/lib/buildinvoicelines.inc.php
+++ b/einvoicing/lib/buildinvoicelines.inc.php
@@ -838,42 +838,73 @@ if (getDolGlobalString($roundTotalConstName) == '1') {		// Same comparison as up
 	$grand_total_ttc = (float) price2num($grand_total_ht + $grand_total_tva, 2);
 }
 
-// already used credit note amount
-$usedcreditnoteamount = 0;
-$usedcreditnote = array();
-$sql = "SELECT re.rowid, re.amount_ht, re.amount_tva, re.amount_ttc,";
-$sql .= " re.description, re.fk_facture_source";
-$sql .= " FROM ".MAIN_DB_PREFIX."societe_remise_except as re";
-$sql .= " WHERE fk_facture = ".((int) $object->id) ." AND description = '(CREDIT_NOTE)'";
-$resql = $db->query($sql);
-if ($resql) {
-	while ($obj = $db->fetch_object($resql)) {
-		$usedcreditnoteamount += abs($obj->amount_ttc);
+// Documents this invoice refers to because a discount applied on it comes from them (BT-25/BT-26, BG-3).
+// Same rows as the core sums used below (attached by fk_facture, joined on the type of the source invoice,
+// never filtered on description), so references and prepaid amount always speak of the same discounts.
+// Customer invoices only: a purchase invoice uses fk_invoice_supplier and would match a foreign rowid here.
+if ($object->element == 'facture' || $object->element == 'invoice') {
+	// Type of the source invoice (BT-X-...), not necessarily a credit note: an '(EXCESS RECEIVED)' is born
+	// from a commercial invoice (380), announcing it as 381 would call an avoir an invoice already paid.
+	$refDocTypeByInvoiceType = array(
+		Facture::TYPE_STANDARD => '380',		// commercial invoice, the source of an excess received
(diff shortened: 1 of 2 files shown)


einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -833,42 +833,113 @@
...
-} else {
-	dol_syslog("Error " . $db->error() . " when looking for credit note linked to invoice to calculate prepaid amount for invoice " . $object->id, LOG_WARNING);
 }
 
-// Amount already received for this invoice: direct payments + used credit notes.
-// getSommePaiement() returns the sum of payments; on Dolibarr <= 22 it also stores that same
-// value into $object->sumpayed, so adding both double-counted the payment (#372: a fully paid
-// invoice reported TotalPrepaidAmount = 2x the amount and a negative DuePayableAmount).
-$getAlreadyPaid = $object->getSommePaiement();
+// Amount already received for this invoice (BT-113 TotalPrepaidAmount), which is what decides how
```
eldy: I added this rule in agents.md
---

## Comments

- Block and inline comments must be written in English.
- Comments must be concise and clear (never more that 5 lines, never more than he number of lines code added or modified).
- Block comments can reach 120 characters 

because having too much comment at each PR fix accumlate too much comment and make the code very difficult to read.

daGrumpf-bxp: Rule applied, branch force-pushed (b417fadc).

The five blocks of the fix went from 54 comment lines to 16, none longer than 5 lines and none wider than 120 characters; the longest one (34 lines, with its ASCII table) is now 5. In the test file the header `\remarks` went from 30 lines to 4 and the prose parts of the docblocks were cut the same way. Only the code and the comments moved, the fix itself is unchanged.

What was cut out is the reasoning, which belongs in the PR description rather than in the file. Same rule from now on in my other PRs.

PR #817, opened 2026-09-05
QUAL einvoicing: format the CDAR dates with the core date helpers

daGrumpf-bxp:
`CdarHandler` formatted its dates with the raw PHP functions while the rest of the module goes
through the core — `dol_print_date($ts, 'dayrfc')` in the protocols, `dol_now()` for the clock. Two
conventions for the same kind of value, in one module.

Converted: the two stamps built from `date_creation` in `generateCdarFile()`, the
`FormattedIssueDateTime` built from `$object->date`, and `getCurrentDateTime()`/`getCurrentDate()`.

This is a refactoring — a CDAR is a document sent to the platform, so the value produced must stay
identical byte for byte for a given timestamp. It does, and the tests prove it rather than assert it.

The one thing that is **not** a pure refactoring is the timezone, and it is worth stating. The
default `$tzoutput = 'auto'` of `dol_print_date()` is not `tzserver`. On 18 as on 24 the core does:

```php
if ($tzoutput === 'auto') {
    $tzoutput = (empty($conf) ? 'tzserver' : (isset($conf->tzuserinputkey) ? $conf->tzuserinputkey : 'tzserver'));
}
```

and `$conf->tzuserinputkey` is overwritten by the constant `MAIN_TZUSERINPUTKEY`, which can be
`tzuserrel`. On such an instance, leaving the default would make the CDAR follow the timezone of
whoever happens to be logged in — a shift that can change the day and, on 1 January, the year. So
`'tzserver'` is written explicitly, with the reason in a comment. `'tzserver'` is what reproduces
`date()` exactly: it formats through a `DateTime` placed in `date_default_timezone_get()`, which is
what Dolibarr sets from `MAIN_SERVER_TZ`.

Only three arguments of `dol_print_date()` are used: `$decorate` appears in 23 and 24 only.

`formatDate()` and `formatDateTime()` are deliberately **left alone**. They pass their input
through unchanged when it is not 8 or 14 digits, and `dol_stringtotime()` cannot reproduce that —
it strips non-digits and pads, so `''`, `'unknown'` or a partial stamp could not come back as they
went in. Changing an output was not the point of this PR. (Neither has any caller in the module
today.)

Tested: PHPUnit 34/34 on each of the seven cores (18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23, 24).
The test compares old and new forms over eight remarkable timestamps — 1 January at 00:00 and
00:30, 31 December at 23:30, both European DST switches, the epoch — across five timezones, and
pins the `MAIN_TZUSERINPUTKEY` behaviour that motivates the explicit argument.

---

This PR stands alone on `main`. It is one of eleven opened together, each on its own subject; every
pair of them, and the whole set in either order, merge onto `main` with no conflict, and the fully
merged tree passes 44/44 on the seven cores. They can be reviewed and merged in any order, or
individually.

### DIFF
--- a/einvoicing/class/utils/CdarHandler.class.php
+++ b/einvoicing/class/utils/CdarHandler.class.php
@@ -282,7 +282,10 @@ class CdarHandler
 
 		// Id format: {SupplierRef}_{StatusCode}_{CreationDate}#{DocType}_{CreationDate} as defined in documentation
 		// TODO: map DOC_INVOICE with $object type
-		$ID = ($statusCode == 212 ? $object->ref : $object->ref_supplier) . '_' . $statusCode . '_' . date('YmdHis', $object->date_creation) . '#' . CdarHandler::DOC_INVOICE . '_' . date('Ymd', $object->date_creation);
+		// 'tzserver' and not the 'auto' default: 'auto' resolves to $conf->tzuserinputkey, which the
+		// MAIN_TZUSERINPUTKEY constant can switch to 'tzuserrel', and the id would then follow the
+		// timezone of whoever triggers the send instead of the server one.
+		$ID = ($statusCode == 212 ? $object->ref : $object->ref_supplier) . '_' . $statusCode . '_' . dol_print_date((int) $object->date_creation, '%Y%m%d%H%M%S', 'tzserver') . '#' . CdarHandler::DOC_INVOICE . '_' . dol_print_date((int) $object->date_creation, '%Y%m%d', 'tzserver');
 
 		// We use same as ID for Name as its not required to be different
 		$Name = $ID;
@@ -502,7 +505,7 @@ class CdarHandler
 					// MDT-97, mandatory in the CTC-FR profile: it says what the lifecycle message is about
 					'ReferenceTypeCode' => CdarHandler::REFERENCE_TYPE_EINVOICE,
 					// Every XP Z12-012 reference example dates the referenced invoice with a plain date
-					'FormattedIssueDateTime' => date('Ymd', $object->date),
+					'FormattedIssueDateTime' => dol_print_date((int) $object->date, '%Y%m%d', 'tzserver'),
 					'ProcessConditionCode' => $statusCode,
 					'ProcessCondition' => $ProcessCondition,
 
@@ -752,6 +755,11 @@ class CdarHandler
 	/**
 	 * formatDateTime
 	 *
+	 * Kept on plain substr() on purpose: this is a pass-through formatter, anything that is not
+	 * exactly 14 digits comes back untouched. dol_stringtotime() strips every non-digit and pads
+	 * the result with '000000', so it cannot give that pass-through back, and it would also
+	 * normalize a 14-digit string that is not a real date (a DST gap hour, or '99999999999999').
+	 *
 	 * @param  string $dateTimeStr datetime
 	 *
 	 * @return string
@@ -767,6 +775,9 @@ class CdarHandler
 
 	/**
 	 * formatDate
+	 *
+	 * Same reason as formatDateTime(): a pass-through formatter the core date helpers cannot
+	 * reproduce, so it stays on substr().
 	 *
 	 * @param  string $dateStr date
 	 *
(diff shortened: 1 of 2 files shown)


einvoicing/class/utils/CdarHandler.class.php
```diff
@@ -500,7 +503,7 @@ public function generateCdarFile($object, $statusCode, $reasonCode = '', $paymen
 					'StatusCode' => $StatusCodeCdar,
 					'TypeCode' => CdarHandler::DOC_INVOICE, // TODO: map DOC_INVOICE with $object type
 					// Every XP Z12-012 reference example dates the referenced invoice with a plain date
-					'FormattedIssueDateTime' => date('Ymd', $object->date),
+					'FormattedIssueDateTime' => dol_print_date((int) $object->date, '%Y%m%d', 'tzserver'),
```
eldy, line 508: Date inside filesmust be surely in UTC or GMT format ?
If yes, dol_print_date should use 'gmt' as last parameter.

einvoicing/class/utils/CdarHandler.class.php
```diff
@@ -780,7 +791,10 @@ public static function formatDate($dateStr)
 	 */
 	public static function getCurrentDateTime()
 	{
-		return date('YmdHis');
+		// dol_now('gmt') is time(), and 'tzserver' renders it in the PHP timezone Dolibarr sets
+		// from MAIN_SERVER_TZ, so this is what date('YmdHis') used to return. See generateCdarFile()
```
eldy: But do we really need to use a local date ?
If date is for a file transfer, expected date is surelly a UTC date?

einvoicing/class/utils/CdarHandler.class.php
daGrumpf-bxp: Checked before answering, and on this one the measurement points the other way: `gmt` would move the date by a day.

`$object->date` is not a UTC instant. `Facture::fetch()` loads it with `$this->db->jdate($obj->df)`, and `DoliDB::jdate($string, $gm = 'tzserver')` defaults to the **server** timezone on all seven supported cores, 18 to 24 (the TODO just above it says it should have been GMT, but it is not). The timestamp therefore holds the stored calendar value read in the server timezone, and printing it back as GMT shifts it.

Stored `2026-01-01 00:00:00`, read with `jdate()`, printed both ways:

| MAIN_SERVER_TZ | 'tzserver' | 'gmt' |
|---|---|---|
| UTC | 20260101 | 20260101 |
| Europe/Paris | 20260101 | **20251231** |
| Indian/Reunion | 20260101 | **20251231** |
| Pacific/Noumea | 20260101 | **20251231** |

On any instance east of Greenwich, so every French one, a 1 January invoice would be dated 31 December in the CDAR.

And there is no UTC to express in this field to begin with. `FormattedIssueDateTime` is `format="102"`, and the XSD types `DateTimeString` as `xs:string` with a required `format` attribute, not as `xs:dateTime` which is the only XSD type carrying a Z. The module's own zugferd helper reads it back as `DateTime::createFromFormat("Ymd", ...)`: a calendar date, no timezone. That matches what platforms actually send us, 93 dates received across 796 logged API calls on a production instance, every one of them `format="102"`, and not a single Z or offset anywhere in a received document.

If the CDAR specification says something else for this field I will follow it, but as the code stands today `gmt` here would be a one day regression.

There is a better answer available for this particular field, though, and I would rather propose it than argue: it echoes the issue date of the **referenced** invoice, which for an incoming document we already hold verbatim as a `format="102"` string. Copying that string instead of recomputing it from the Dolibarr timestamp removes the timezone question entirely and guarantees the CDAR references the document with the date its sender wrote. That only leaves status 212, sent on one of our own invoices, where there is no received document to copy from. Say the word and I will do it in a follow-up.

einvoicing/class/utils/CdarHandler.class.php
daGrumpf-bxp: This one is a different kind of value and you are right. `getCurrentDateTime()` and `getCurrentDate()` come from `dol_now('gmt')`, so unlike the referenced invoice date above they really do hold an instant, and for a timestamp that travels between platforms UTC is the safer convention. I will switch these two to `gmt`.

One caveat worth putting on the record. `IssueDateTime` is `format="204"`, which carries **no offset**: the XSD types `DateTimeString` as `xs:string` plus a format code, and the reader parses it as a local wall clock, `DateTime::createFromFormat("YmdHis", ...)`. Writing UTC into it therefore does not *declare* UTC to the receiver, it only changes the value. The UN/EDIFACT 2379 code that carries the offset is 205, `CCYYMMDDHHMMZHHMM`, which the library does not implement. If the intent is that the receiver knows the stamp is UTC, that needs the format code too, and I would rather do that as its own change once the specification point is settled.

For information, I could not settle it by looking at what the platforms do: across 1358 logged API calls on two production instances, no AcknowledgementDocument was ever received, so there is no incoming CDAR to copy the convention from.

eldy (approved):

PR #827, opened 2026-09-05
FIX: [einvoicing] The module no longer shows a raw translation key

daGrumpf-bxp:
Twenty-seven translation keys the module asks for with a literal argument are defined by no
`en_US` file the page loads — neither the module's nor the core's — on at least one supported
core. Dolibarr then prints the key itself: `Document` as the title of a document card,
`CheckStatus` in the agenda event written after a status check, `Failed` as the status of an API
call, `until` next to the expiry of the access token.

## What is in here

- **15 keys added to `en_US`**, none of which any supported core defines: `Document`,
  `ShowDocument`, `DeleteDocument`, `ShowCall`, `DeleteCall`, `Failed`, `Skipped`,
  `ClassNotFound`, `ErrorNumberingModuleNotSetup`, `ErrorLoadingInvoice`, `FlowId`,
  `CheckStatus`, `MyCompany`, `Seller`, `Buyer`.
- **10 keys the core added along the way**, so they are missing on the older cores the module
  supports: `RefSupplierBill` is absent from 18, `AccessToken` and `Message` from 18 to 19,
  `Automatic`, `Input` and `Output` from 18 to 21, `InvoiceType` up to 22, and
  `ErrorThirdPartyNotFound`, `Success` and `VATExemptionCode` up to 23. They are added with the
  wording of the core, so nothing changes on 24.
- **2 keys live in a core lang file the page does not load**, on every supported core: `Until`
  in `agenda.lang` and `Options` in `accountancy.lang`. The setup page and the admin tabs now load
  those files, and the module carries no copy of the key. On a French instance the token line of
  the setup page reads `(jusqu'à …)` where `main` shows `(until …)`, the raw key it asks for.
- **6 call sites named a key that already exists**: `EInvStatus212Paid` is
  `EInvStatus212PaymentReceived`, the status label of this module; `bytes` and `Standard` are
  `Bytes` and `InvoiceStandard` in the core. And the agenda tab of a document asks for three
  keys of `agenda.lang` and `commercial.lang` without loading them, so it now does.

Only `en_US` is touched, the other languages are left to Transifex.

## Measurement

A script loads each instance through `Translate` and asks the loader, for the literal keys the
module passes to `trans()`, whether it knows them at all. On the seven supported cores:

| | 18.0.10 | 19.0.4 | 20.0.4 | 21.0.4 | 22.0.5 | 23.0.4 | 24.0.1 |
| main | 33 | 32 | 30 | 30 | 27 | 26 | 23 |
| this branch | 0 | 0 | 0 | 0 | 0 | 0 | 0 |

Each key added to `en_US` was also checked by hand against every `en_US` file of the seven
cores: the 15 exist in none of them, the 10 only from the version stated above.

The PHPUnit suite of the module, 39 files, passes on those same seven cores.

The check itself is not added to the CI here: it wants a Dolibarr instance, so its place is a
step of the `einvoicing_phpunit` job, and that file is being changed elsewhere. Happy to add it
once that lands, if you want the guard.

### DIFF
--- a/einvoicing/admin/setup.php
+++ b/einvoicing/admin/setup.php
@@ -78,7 +78,7 @@ require_once "../class/einvoicing.class.php";
 
 
 // Translations
-$langs->loadLangs(array("admin", "einvoicing@einvoicing"));
+$langs->loadLangs(array("admin", "agenda", "einvoicing@einvoicing"));
 
 // Initialize a technical object to manage hooks of page. Note that conf->hooks_modules contains an array of hook context
 /** @var HookManager $hookmanager */
--- a/einvoicing/admin/setup_devtools.php
+++ b/einvoicing/admin/setup_devtools.php
@@ -334,7 +334,7 @@ if (getDolGlobalString('EINVOICING_PDP')) {
 	print '<span class="width150 inline-block">'.$langs->trans("InvoiceType").'</span> ';
 	if ((float) DOL_VERSION >= 24.0) {
 		$typeofinvoice = array(
-			Facture::TYPE_STANDARD => array('label' => $langs->trans('Standard')),
+			Facture::TYPE_STANDARD => array('label' => $langs->trans('InvoiceStandard')),
 			Facture::TYPE_REPLACEMENT => array('label' => $langs->trans('ReplacementInvoice')),
 			Facture::TYPE_DEPOSIT => array('label' => $langs->trans('Deposit')),
 			Facture::TYPE_CREDIT_NOTE => array('label' => $langs->trans('CreditNote')),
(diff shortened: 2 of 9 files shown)


eldy:
The key Until is added but is already present into core lang files (for example agenda.lang) so should not be added.

daGrumpf-bxp:
Right: `Until` is in agenda.lang, and `Options` in accountancy.lang, on every core from 18 to 24.
Both keys are removed; the setup page now loads agenda.lang and the admin tabs accountancy.lang.
The other 25 were checked by hand against every en_US file of the seven cores: 15 exist in none
of them, 10 only from the version stated in the description, so they stay for the older cores.

PR #828, opened 2026-09-05
QUAL: CI - the same checks, three jobs less, and the 23.0.0 core back on a pull request

daGrumpf-bxp:
`44e0976e` dropped 23.0.0 from the pull request matrix to lighten the CI. That is the only lever
visible from the outside, and it costs coverage in the middle of the version range, where the traps
of this module live. This gives that core back and still costs less.

Measured over 300 runs in 10 hours (~45 pull request pushes a day): per job that declares MariaDB,
~20 s to start the container and ~25 s to build the instance, for **12 s of PHPUnit**. Setup is
~85 % of the cost, paid nine times on a push to main - so the lever is to pay it fewer times, and
not to start what a change cannot affect.

## What this does

- **What a change touches is answered in one place**: `.github/actions/changed` reads the files
  of the pull request once and says which modules they belong to - from the `F:` lines of
  `MAINTAINERS`, where the repository already declares them - and what kind of files they are,
  from `.github/changed.map`. It answers facts; which cores to run on, which check to skip, stays
  the business of the workflow that asks. Anything it cannot classify it calls "everything".
- **The conformance validation no longer builds an instance of its own**: it runs on the matrix
  leg of the stable core, which already has one standing, and it now also validates the documents
  of `test/samples/`. The `einvoicing_conformance` workflow goes, and with it the documented risk
  that its two paths filters drift from those of the phpunit job.
- **The readiness loop of the `package` job probed `/index.php` with `curl -f` on an instance it
  had just broken on purpose** (compat/ removed), so it slept its 30 seconds at every run. That
  step goes from 31 s to 1 s.
- **The matrix follows what the change touches**: seven cores on main, 18/23/24 on a pull
  request, and the stable core alone when nothing but translations or documentation of this module
  changed - 58 of the last 300 commits touching it are in that case. And the **PDF/A job no longer
  starts on a translation**: it builds from the committed fixtures, with no Dolibarr and no
  language file.
- **phpcs reads the files of the pull request**, the whole repository on main and whenever the
  ruleset changes; the setup no longer builds 13 PHP extensions that no check ever loads; the
  workflow gets the `concurrency` group the four others had; and **actionlint** reads the workflow
  files when a change touches them - it found `inputs.gh_event` in `phan.yml`, on a workflow that
  declares no input, fixed here.

## Measurements (real runs, ±10 % variance)

| | before | after |
|---|---|---|
| push on main | 10 jobs / 660 s | 9 jobs / **564 s**, plus one workflow removed |
| pull request on code | 5 jobs / 308 s, **2 cores** | 5 jobs / 313 s, **3 cores** |
| pull request on translations | 9 jobs / 438 s | 6 jobs / **224 s** |
| phpcs job | 54 s | **20 s** |

Wall clock does not move: the 23 leg goes to 116 s, what the conformance job it replaces cost.

Tried out on three test pull requests on a fork - one on code (3 cores, phpcs in list mode, which
did catch a violation), one on a language file (1 core, no PDF/A job), one push on main (7 cores),
all green. The classifier and the two policies reading it are covered by a harness feeding them
file lists, 9 cases each - one being the module the hardcoded list in `ci-pr-autoassign.yml` never
matched, since it tests `dolisecu/*` for a directory named `dolisecure/`.

**Built on top of #825, and it carries its commit.** A pull request opened against this repository can only take one of its own branches as a base, so the dependency cannot be declared to GitHub: the one commit of #825 is the first of this branch and shows in the diff below until #825 is merged. Its step is rewritten here to use the scope this pull request computes, rather than fetching the base ref on every pull request.


daGrumpf-bxp:
Two things measured and dropped along the way, so nobody redoes them: lowering the `--health-interval` of the MariaDB service from 10s to 2s gains nothing - the runner has its own backoff wait loop that swallows the finer healthcheck - and fetching the core through the codeload tarball or an `actions/cache` entry does not pay against the 7.7 s of a shallow `git fetch`.

PR #834, opened 2026-09-06
Fix XSS: escape received e-invoice fields shown in the sync results panel

daGrumpf-bxp:
### Problem

The synchronization results panel (`document_list.php`) prints a run's technical messages and details as raw HTML. Several of those messages are built by concatenating fields read verbatim from a **received** e-invoice: document number (BT-1), referenced document ids (BT-25), line id, document type code (BT-3), and the product/thirdparty errors that echo them.

`getXPathValue()` returns the decoded node text, so a value delivered XML-escaped inside a perfectly valid document (`&lt;img src=x onerror=...&gt;`) turns back into live markup when the panel renders it. Under the e-invoicing mandate the sender of an incoming invoice cannot be chosen, so a counterparty could store markup that runs in the operator's browser session when a reception fails and the run is displayed.

It is a hard-to-reach stored XSS (it needs a reception that fails on one of these fields, and a strict access point may reject the payload at intake), but the module should not depend on the platform to sanitize its own output.

### Fix

Escape the untrusted values with `dol_escape_htmltag()` **at the point where they are interpolated** into the message, so the intended panel layout (line breaks, the "skipped" span, the action buttons) is preserved while the data is neutralized:

- `CIIProtocol`: document number, referenced document ids, line id and document type code in the reception error messages.
- `SupplierInvoiceHelper::refLookupErrorMessage()`: escapes its `ref` and `context` once, covering every caller.
- `CommonProtocol`: the thirdparty and product creation/update error messages.
- `SuperPDPProvider` / `EsalinkPDPProvider`: the flow id and exception text pushed into the per-run details.

Clean references (letters, digits, dashes) pass through unchanged, so the panel and the existing `refLookupErrorMessage` unit test are unaffected. `dolics` clean, `php -l` clean on all five files.

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -887,7 +887,7 @@ class CIIProtocol extends AbstractProtocol
 		}
 
 		if ($supplierInvoiceId > 0) {
-			$return_messages[] = 'Supplier Invoice with reference ' . $parsedHeader['documentno'] . ' already exists';
+			$return_messages[] = 'Supplier Invoice with reference ' . dol_escape_htmltag((string) ($parsedHeader['documentno'] ?? '')) . ' already exists';
 
 			$supplierInvoice = new FactureFournisseur($db);
 			$supplierInvoice->fetch($supplierInvoiceId);
@@ -948,7 +948,7 @@ class CIIProtocol extends AbstractProtocol
 					return [
 						'res' => -1,
 						'postponeflow' => 1,
-						'message' => 'Document ' . $refDoc . ', required by received document ' . $parsedHeader['documentno'] . ', was not found in Dolibarr',
+						'message' => 'Document ' . dol_escape_htmltag((string) $refDoc) . ', required by received document ' . dol_escape_htmltag((string) ($parsedHeader['documentno'] ?? '')) . ', was not found in Dolibarr',
 						'actioncode' => 'LINKED_INVOICE_NOT_FOUND',
 						'actionurl' => 'none',
 						'actiondata' => array('supplierref' => $refDoc, 'linkedref' => ($parsedHeader['documentno'] ?? ''), 'socid' => (int) $socId),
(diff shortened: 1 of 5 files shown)


daGrumpf-bxp:
Tested on a local multi-version bench (Dolibarr 18.0.10 and 24.0.1), reception path.

**Fix proven at the sink (controlled, both directions).** A received CII carrying `` XML-escaped in BT-3 (`TypeCode`) is fed to `CIIProtocol::createSupplierInvoiceFromSource()`; the returned message is what the sync panel renders.

| version | on `main` | on this branch |
|---|---|---|
| 18.0.10 | `... : ` (raw) | `... : &lt;img src=x onerror=...&gt;` (escaped) |
| 24.0.1 | `... : ` (raw) | `... : &lt;img src=x onerror=...&gt;` (escaped) |

**No regression.** A clean received invoice still imports (`res > 0`) on both versions, and its normal messages are unchanged (clean references pass through `dol_escape_htmltag()` untouched).

**Full module PHPUnit suite green on both versions** (28 test files each, incl. `CIIProtocolTest`, `SupplierInvoiceHelperTest`, `ReceivedInvoiceLinesTest`, `CIITextEscapingTest`). `dolics` and `php -l` clean on the five changed files.

daGrumpf-bxp:
Re-ran the reception recette across the 7 supported versions (v18 → v24), fix branch deployed to each.

**Fix at the sink** — a received CII with `` XML-escaped in BT-3 fed to `createSupplierInvoiceFromSource()`, checking the returned message:

| v18.0.10 | v19.0.4 | v20.0.4 | v21.0.4 | v22.0.5 | v23.0.4 | v24.0.1 |
|---|---|---|---|---|---|---|
| escaped | escaped | escaped | escaped | escaped | escaped | escaped |

**No regression** — a clean received invoice imports (`res > 0`) on all 7 versions.

**Full module PHPUnit suite** (28 files) on all 7 versions: no failures, no errors. Only pre-existing version-dependent skips in `MultidirOutputCompatTest`, unrelated to this change. `dolics` and `php -l` clean on the five files.

PR #835, opened 2026-09-06
QUAL: [einvoicing] Remove the compat code for v17 every supported Dolibarr already ships

daGrumpf-bxp:
The module requires Dolibarr 18 or later, and three of its backports polyfill symbols that
**18.0.0 already ships** (checked in the 18.0.0 core, not only in 18.0.10):

- `compat/files.lib.php` carried `dolChmod()`, in `functions.lib.php` since 18.0.0. Its loader
  in `CIIProtocol.class.php` was gated on `DOL_VERSION < 18`, a test never true here.
- `GETPOSTDATE()` in `compat/functions.lib.php`: in the core since 18.0.0, so the
  `function_exists()` guard always found the core version and the copy never ran.
- `einvoicingDolGetButtonActionDropdown()` in `lib/einvoicing.lib.php`: `dolGetButtonAction()`
  builds the dropdown from an array `$url` since 18.0.0, and the two call sites in
  `actions_einvoicing.class.php` only reached the polyfill below 18.

The file, the two functions, the loader and the two `< 18` branches are removed; the reload test
of the shims drops the entries that went with them. **`dolPrintHTML()` stays**: it appeared in a
18.0.x patch release only, so 18.0.0 still needs that copy. The other shims (`getDolGlobalFloat`,
`GETPOSTFLOAT`, `dolPrintHTMLForAttribute`, `getMultidirTemp`) cover 19 to 21 and are untouched.

## Tested

- PHPUnit, 28 files on Dolibarr 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1 and
  develop (25.0.0-alpha): 224/224.
- A probe on 18.0.10 and develop: `dolChmod`, `GETPOSTDATE` and `dolGetButtonAction` are served by
  the core file, `GETPOSTDATE('x', 'getpost', 'gmt')` reads the selector fields, and the dropdown
  built from the array the hooks pass has its toggle, its enabled item and hides the refused one.
- On 18.0.10 over HTTP: the invoice card shows the e-invoice dropdown, the flow list loads with
  and without a date filter, and a CII is generated on a validated invoice (the writer calls the
  core `dolChmod()` on it).
- phpcs and phan clean.

## Also in here

The phpcs workflow hands phpcs the file list of the pull request, deleted files included, and phpcs
stops on a path that does not exist. The workflow now keeps only the files present in the checkout,
otherwise no pull request that deletes a PHP file can be green.


atm-maxime:
Are you sure it's the right thing to do ? I think some are working on making the module compatible with 17.0 version but it's not declared yet in the module descriptor. If you remove these, the 17.0 compatibility is lost.
i.e. #617

daGrumpf-bxp:
Fair question, but v17 is already off the table: @eldy restored `need_dolibarr_version = array(18, -3)` straight on main in 6fb1149 ("Restore initial minimum target. Too many pb with v17."), and said it plainly in https://github.com/Dolibarr/dolibarr-community-modules/pull/621#issuecomment-5407593741 ("we should just forgot the compatibility with v17", v18 being the oldest LTS still maintained) and again in review on #620 ("We can abandon maintenability for version < 18, it is an endless race").

And yes, I was one of those making v17 work — #619/#616/#621 are mine. I'm not sad to drop it: the descriptor says 18, so code branching on `< 18` is dead code that no CI covers and nobody can test, i.e. noise.

Note this PR removes only what **18.0.0 itself already ships** (`dolChmod`, `GETPOSTDATE`, the array form of `dolGetButtonAction`) — verified in the 18.0.0 core, not just 18.0.10. Everything covering 19-21 stays, and `dolPrintHTML` stays too since it only landed in an 18.0.x patch.

If the board reverses the decision, #621 and the rest of the v17 work are still in git and easy to bring back; until then main is a v18+ module.

eldy:
The board has validated that the "official version"  (understand "supported") is v18+

However, i think we can make a difference between "officialy supported"  and  "completely removed".

We can discard and ignore any problem for v17 and not spend time on it anymore (no CI need, no test and no review need). 
Also if there is a risky change or too annoying trouble with v17, we are free to remove code related to v17.

But for the moment, we can keep code as it is (so if some users that started to use itin v17 can continue). Developers that need v17 compat will just use it at their own risk and will port breaking things themself but we can keep what was already done.

So 
- i discard this PR (we keep what work in v17 working) 
- i validate the PR #621 that are bugs on v17 only.
- But i keep the flag minimal version in the module descriptor to v18 and we must continue to discourage any user to use another version than v18+

PR #838, opened 2026-09-06
QUAL: [einvoicing] A pull request runs on the two ends of the supported range, in cores and in PHP

daGrumpf-bxp:
A pull request now runs the suite on the two ends of the supported range, Dolibarr 18.0.0 and 24.0.0, instead of 18/23/24: the floor the module declares and the latest release, where the version traps sit. The FNFE validation (EN 16931 and CTC-FR schematrons), the translations-only run and the package job move from 23.0.0 to 24.0.0.
The PHP of each core now walks the range `install/check.php` of that release accepts, one version per core: 7.4 on 18.0.0 up to 8.5 on 24.0.0. Before this, 18.0.0 ran on PHP 8.2, above the 8.1 its own check.php warns about.
So a pull request meets both the oldest PHP the vendored libraries take and the newest PHP the core's own CI runs, with no job added; main covers every PHP from 7.4 to 8.5 once, on its seven cores.
Checked locally before pushing: `php -l` of the 93 non-vendor PHP files of the module, its tests and the CI scripts is clean under both PHP 7.4.33 and 8.5.10; actionlint reports nothing beyond pre-existing shellcheck notes.
25 is not in the matrix yet: `develop` is 25.0.0-alpha and ships no demo dump, so the instance action cannot build it without driving the 24 to 25 migration; it joins once a 25.0.0 release ships its dump.

### DIFF
--- a/.github/workflows/module_einvoicing_phpunit.yml
+++ b/.github/workflows/module_einvoicing_phpunit.yml
@@ -47,8 +47,9 @@ env:
 jobs:
   # Which cores the suite runs on. What lands on main is run on all seven supported cores: the
   # version specific traps of this module live in the middle of the range. A pull request answers on
-  # the floor, the stable and the newest - and on the stable alone when it changes no code at all,
-  # since a translation or a document cannot behave differently from one core to the next.
+  # the two ends, the floor the module declares and the latest release - and on the latest alone
+  # when it changes no code at all, since a translation or a document cannot behave differently
+  # from one core to the next. The next major joins the matrix once a release ships its demo dump.
   cores:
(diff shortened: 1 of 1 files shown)


daGrumpf-bxp:
Correction: the second and third paragraphs of the description above, about the PHP version of each core, describe a commit that is not part of this merge. Only the change of cores landed here: a pull request runs on 18.0.0 and 24.0.0, and the validation, the translations-only run and the package job moved to 24.0.0. Every core still runs on PHP 8.2 or 8.3 as before.

The PHP part is #839, on its own.

PR #840, opened 2026-09-06
QUAL: [einvoicing] The module declares the PHP version its libraries need

daGrumpf-bxp:
The module declared `phpmin = array(7, 0)`, below what it can actually run on.

The bundled horstoeko/zugferd requires PHP >= 7.3 in its `composer.json`, as do its
`horstoeko/stringmanagement` and `horstoeko/mimedb` dependencies, and the vendor tree was resolved
with `platform.php = 7.3.0` (see `einvoicing/composer.disabled.json`). The generated
`einvoicing/vendor/composer/platform_check.php` therefore refuses anything below PHP 7.3.0, with a
fatal `E_USER_ERROR` rather than a message — and `FacturXProtocol.class.php` requires the autoloader
at file scope, so merely including it is enough.

**Choosing CII does not keep that file out of the way.** It governs what we *emit*. Reception picks
the protocol from what *arrived*: `detectProtocolFromContent()` answers `FACTURX` for anything
starting with `%PDF-`, and `getProtocol()` is then called by the flow picker of
`AbstractPDPProvider`, by `Document::cleanXmlData()`, by the consistency check of
`SupplierInvoiceHelper` and by `product_mapping.php`. None of them reads `EINVOICING_PROTOCOL`, and
incoming synchronisation is on unless it is turned off. A supplier that sends Factur-X — the format
most French senders use — loads the library on a site set to CII.

Measured on the bench for #764/#785 (04-05/09/2026), with a `horstoeko\zugferd` probe on
`get_declared_classes()`: with the Approved Platform account set to CII, not one class is loaded and
`FacturXProtocol.class.php` is never even included; with the account set to Factur-X, the document
handed back is a PDF and **three classes are loaded**, on `main` and on the branch alike. That
account setting lives at the platform, not in Dolibarr — hgy29 reported exactly such an account on
 #718. So "the library is never loaded with the recommended setup" holds only for one particular
platform-side configuration, not for the module.

Nothing reachable needs more than 7.3. The only PHP 8.x files under `vendor/` are the optional
`horstoeko\zugferd\codelistsenum` enums, autoloaded PSR-4 and referenced neither by the module nor by
zugferd itself, two Rector configs under `build/` that are not autoloaded, and
`symfony/polyfill-mbstring/bootstrap80.php`, required only when `PHP_VERSION_ID >= 80000`. The PHP
7.4+ and 8.x functions the reached libraries call (`str_contains`, `get_debug_type`, `mb_str_split`,
`array_is_list`) all come from the bundled symfony polyfills.

Measured, not assumed: `php -l` over the vendor tree under PHP 7.3.33 and 7.4.33 errors on exactly
those 18 unreachable files and on no other, and is clean on the 92 non-vendor PHP files of the
module — which, for the record, also parse under 7.0, 7.1 and 7.2. Parsing is not the point; the
`platform_check.php` fatal is.

So a site below 7.3 now gets the module's own prerequisite warning instead of a fatal from the
autoloader, the first time a supplier sends a Factur-X.

### DIFF
--- a/einvoicing/core/modules/modEInvoicing.class.php
+++ b/einvoicing/core/modules/modEInvoicing.class.php
@@ -157,7 +157,7 @@ class modEInvoicing extends DolibarrModules
 		$this->langfiles = array("einvoicing@einvoicing");
 
 		// Prerequisites
-		$this->phpmin = array(7, 0); // Minimum version of PHP required by module
+		$this->phpmin = array(7, 3); // Minimum version of PHP required by module: the bundled horstoeko/zugferd needs 7.3
 		// $this->phpmax = array(8, 0); // Maximum version of PHP required by module
 		$this->need_dolibarr_version = array(18, -3); // Minimum version of Dolibarr required by module
 		// $this->max_dolibarr_version = array(19, -3); // Maximum version of Dolibarr required by module


einvoicing/core/modules/modEInvoicing.class.php
```diff
@@ -157,7 +157,7 @@ public function __construct($db)
 		$this->langfiles = array("einvoicing@einvoicing");
 
 		// Prerequisites
-		$this->phpmin = array(7, 0); // Minimum version of PHP required by module
```
eldy, line 160: The minimal version of module is php is
7.0
Php 7.3 is required only if you choose the facturx format, that is not the default. And when you select the facturx format (not recommended), we should have a guard to allow selection only if we are on php 7.3 
If using the correct recommended default setup (cii format), there is no reason to restrict to php7.3+ since the lib horstoeko shoul NEVER be loaded with the default recommended setup.

daGrumpf-bxp: Agreed, and reverted: `phpmin` stays `array(7, 0)`. The library is behind Factur-X only, and the
recommended CII setup never loads it, so it has no business setting the prerequisite of the module.

The branch now carries the guard you asked for instead. The entry was already greyed with "PHP 7.3+",
but that alone stops a browser and nothing else: `FormSetupItem::setValueFromPost()` saves what the
form posted without checking it against the list it offered (its own `// TODO add value check`). So
the setup page now also says *why* the format is out of reach, next to the selector, and refuses the
value server side through `setValueFromPostCallBack()` — the extension point the core has for that,
present since v18. The rest of the form still saves.

Threshold in one place, `ProtocolManager::FACTURX_PHP_MIN`, read by the greying, the sentence and
the refusal.

Verified on the seven supported cores (18.0.10 to 24.0.1, PHP 7.4 to 8.5): with the real threshold
Factur-X stays selectable and savable everywhere — no change for anyone; with the threshold raised
above the PHP of each instance, the entry is greyed with its reason on all seven and a forced
`EINVOICING_PROTOCOL=FACTURX` leaves the constant on `CII`. Screenshots in the comments.

daGrumpf-bxp:
**Obsolete — please disregard.** This comment showed a setup-page guard I had put on this branch
after conceding that the module could stay on PHP 7.0. That concession was wrong (see the review
thread): choosing CII does not keep `horstoeko/zugferd` out of the way, because reception picks the
protocol from what arrived, not from `EINVOICING_PROTOCOL`. The branch is back to its single change,
`phpmin = array(7, 3)`, and the guard is not part of it.

Screenshots removed to avoid describing code this pull request no longer contains.

einvoicing/core/modules/modEInvoicing.class.php
daGrumpf-bxp: Correcting my previous answer: I conceded too fast, and the branch is back to `phpmin = array(7, 3)`.

The premise does not hold. `EINVOICING_PROTOCOL` governs what we **emit**. Reception picks the
protocol from what **arrived**: `detectProtocolFromContent()` answers `FACTURX` for anything starting
with `%PDF-`, and `ProtocolManager::getProtocol()` is then called by the flow picker of
`AbstractPDPProvider`, by `Document::cleanXmlData()`, by the consistency check of
`SupplierInvoiceHelper` and by `product_mapping.php`. None of them reads `EINVOICING_PROTOCOL`, and
incoming synchronisation is on unless it is turned off. `FacturXProtocol.class.php` requires
`vendor/autoload.php` at file scope, so including it is enough to hit
`vendor/composer/platform_check.php`, whose `E_USER_ERROR` no `try/catch` can hold.

Measured, not argued — the bench work behind #764/#785 (04-05/09), with a probe on
`get_declared_classes()` filtered on `horstoeko\zugferd`:

| Approved Platform account | what comes back | zugferd classes loaded |
|---|---|---|
| CII | CII, 18 159 B | **0** — `FacturXProtocol.class.php` not even included |
| Factur-X | the sender's PDF, byte for byte (pass-through) | **3** |
| Factur-X, sender in UBL | a Factur-X built by the platform | **3** |

That setting lives at the platform, not in Dolibarr, and hgy29 described exactly such an account on
 #718 ("my SuperPDP account was configured with Factur-X as preferred conversion format"). So a site
set to CII, doing nothing unusual, loads the library the first time a supplier sends a Factur-X —
and on PHP 7.2 that is a fatal, not the prerequisite message.

Factur-X is the format most French senders use. A module that cannot read one cannot claim to run on
the PHP where it cannot read one.

Two consequences, both yours to arbitrate:

- if 7.3 goes in, the `'PHP 7.3+'` greying of the Factur-X entry in `ProtocolManager` becomes dead
  code — happy to remove it in a separate PR, or to leave it as a belt.
- 7.0 has been end of life since **10 January 2019**, and nothing in this repository has ever
  verified it: the PHPUnit matrix starts at PHP 7.4 (the floor of Dolibarr 18) and phan runs
  `--minimum-target-php-version 7.2`. Keeping 7.0 in the declaration would mean adding a check that
  actually proves it.

daGrumpf-bxp:
@eldy — a genuine question before this gets arbitrated, because it decides the answer more than the
code does: are there usage figures on the board side?

PHP 7.0 has been end of life since **10 January 2019**. `7.0.33` was its last release, so any site
still on it has had no security patch for seven and a half years. And the core has already moved
past it: `install/check.php` errors below 7.0 on Dolibarr 18 to 22, but below **7.1** on 23 and 24.
So the only installs a `phpmin` of 7.0 can serve here are 18-to-22 sites running an unpatched
interpreter — and they would have to receive Factur-X invoices through an Approved Platform, which
is not something a site of that vintage is usually doing.

If the telemetry, Dolistore, or the support side says those sites exist and use this module, then
keeping 7.0 means adding a check that actually proves it: nothing in this repository ever has. The
PHPUnit matrix starts at PHP 7.4, the floor of Dolibarr 18, and phan runs
`--minimum-target-php-version 7.2`. I measured that the 92 non-vendor PHP files of the module do
parse under 7.0, 7.1 and 7.2 — so a syntax-only job would go green and prove nothing, since the wall
is `vendor/composer/platform_check.php` at runtime, not the parser.

If they do not exist, 7.3 costs nobody anything and buys the module a clear prerequisite message
instead of a fatal error from the autoloader.

eldy:
> @eldy — a genuine question before this gets arbitrated, because it decides the answer more than the
> code does: are there usage figures on the board side?
> 
> PHP 7.0 has been end of life since **10 January 2019**. `7.0.33` was its last release, so any site
> still on it has had no security patch for seven and a half years. And the core has already moved
> past it: `install/check.php` errors below 7.0 on Dolibarr 18 to 22, but below **7.1** on 23 and 24.
> So the only installs a `phpmin` of 7.0 can serve here are 18-to-22 sites running an unpatched
> interpreter — and they would have to receive Factur-X invoices through an Approved Platform, which
> is not something a site of that vintage is usually doing.
> 
> If the telemetry, Dolistore, or the support side says those sites exist and use this module, then
> keeping 7.0 means adding a check that actually proves it: nothing in this repository ever has. The
> PHPUnit matrix starts at PHP 7.4, the floor of Dolibarr 18, and phan runs
> `--minimum-target-php-version 7.2`. I measured that the 92 non-vendor PHP files of the module do
> parse under 7.0, 7.1 and 7.2 — so a syntax-only job would go green and prove nothing, since the wall
> is `vendor/composer/platform_check.php` at runtime, not the parser.
> 
> If they do not exist, 7.3 costs nobody anything and buys the module a clear prerequisite message
> instead of a fatal error from the autoloader.
> 

18 is lts. There is a very high nulber of user on it, like we have on 19, 20, like we have so on php 7.0, 7.1, etc.

Also upgrading an erp is an operation with a high cost for enterprises. Dolibarr adn is to never decide and force a user to upgrade if he don t want. This must as long as possible remain the choice of end user.

Also there is no reason that a user get a fatal error on hostroeko lib, whatever is its version of dolibarr and php since the default recommanded setup is cii that never use this lib, and enabling thebmode that need it is restricted and possible only if php is compatible.

daGrumpf-bxp:
Taken from the other end, without touching `phpmin`: **#866**.

Since the argument here was that the recommended setup is CII, I measured what a CII site does when a supplier sends a Factur-X. Dolibarr 18.0.10, receiver on `EINVOICING_PROTOCOL=CII`, real invoices through an access point, five PHP built side by side: on 7.0, 7.1 and 7.2 the incoming synchronization dies with the Composer fatal (`exit 255` in CLI, a 107-byte HTTP 500 in the browser), because reception picks the protocol from the document that arrived, not from the setting.

#866 does not raise the requirement. `ProtocolManager::getProtocol()` answers `null` before the include when the libraries cannot run on this PHP, the flow is kept pending as it already is for an unreadable syntax, and the user is told the two ways out: set the conversion format of the access point account to CII, or upgrade PHP. A site that upgrades nothing keeps working, and its received invoices are no longer lost behind a white page.

Issue #783, opened 2026-09-04
[einvoicing] A line discount without BT-137 is converted using the wrong base


TDS-France:
### Bug

Split out of #772, where it was suggested this deserves its own issue.

A supplier line that carries a discount is imported with a total that is too low, so the invoice no longer matches the document.

Dolibarr stores line discounts as a percentage, so the import has to turn the discount amount into one, and for that it needs the price *before* the discount. EN 16931 has an optional field for it, BT-137, and my documents (OVH through SuperPDP) do not send it. The fallback then uses BT-131, which is the amount *after* the discount:

```php
$base = $allowances[0]['basisAmount'] ?? $lineTotalAmount;
```

The percentage is computed against the wrong number, and the line comes out short:

| discount | BT-131 announced | imported | diff |
|---|---|---|---|
| 5.00 | 95.00 | 94.74 | -0.26 |
| 10.00 | 90.00 | 88.89 | -1.11 |

The amount before discount is already computed a few lines below, as `priceWithoutDiscount`. Using it as the fallback would leave documents that do send BT-137 untouched.

Not covered by #776: that one repairs a line whose rebuild lands on zero or on the wrong sign. Here the rebuild is plausible, only short, so it keeps its current behaviour, with a warning in the log.

Two limits worth stating: the figures above come from replaying the function's arithmetic, not from a real import, since the only document I have is the gift line of #772. And this is my first time in this code, so I may be missing a reason for the current base.

### Module

einvoicing

### Dolibarr Version

23.0.2, module 1.1.0

### Environment PHP

8.1.34

### Environment Database

MariaDB 10.6.23

### Steps to reproduce the behavior and expected behavior

Import a supplier document with a line that carries a discount amount, no BT-137, and a non-zero unit price.

Expected: the line total matches the announced BT-131.
Observed: it comes out short, and the invoice total no longer matches the document.

hansemschnokeloch:
Same issue I think with factur-x xml
```
    <ram:IncludedSupplyChainTradeLineItem>
      <ram:AssociatedDocumentLineDocument>
        <ram:LineID>4</ram:LineID>
      </ram:AssociatedDocumentLineDocument>
      <ram:SpecifiedTradeProduct>
        <ram:Name>6C - Colissimo Domicile Sign. F</ram:Name>
        <ram:Description>Charge 6C</ram:Description>
        <ram:ApplicableProductCharacteristic>
          <ram:Description>BLOC</ram:Description>
[...]

PR #845, opened 2026-09-07
Fix line discount dropped when BasisAmount is zero

Kazimir42:
## What

When an incoming line discount has `<BasisAmount>0.00</BasisAmount>` but a real `ActualAmount`, the discount was silently dropped and the line got imported at full price.

`?? ` only falls back on `null`, so a BasisAmount of `0` was kept as the base, failed the `if (!$base)` guard, and `resolveLineDiscountPercent()` returned `false`.

Some pivots (e.g. Esalink) emit `BasisAmount = 0` next to a valid discount amount, so we now treat `0` as "not stated" and fall back to the amount before discount.

## Example

Line with gross 4766, 50% discount, net 2383:

```xml
    <ram:CalculationPercent>50.00</ram:CalculationPercent>
    <ram:BasisAmount>0.00</ram:BasisAmount>
    <ram:ActualAmount>2383.00</ram:ActualAmount>
<ram:LineTotalAmount>2383.00</ram:LineTotalAmount>
````

- Before: discount ignored, line imported at 4766.00
- After: discount applied, line imported at 2383.00

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -3430,9 +3430,9 @@ class CIIProtocol extends AbstractProtocol
 		// price of the Dolibarr line.
 		$priceWithoutDiscount = (float) $lineTotalAmount - $totalChargeAmount + $totalDiscountAmount;
 
-		// Base used for percent calculation — BT-137 when the document states it, the amount before
-		// discount otherwise (issue #783).
-		$base = $allowances[0]['basisAmount'] ?? $priceWithoutDiscount;
+		// Base for the percent — BT-137 if given, amount before discount otherwise (issue #783).
+		// A BasisAmount of 0 (some pivots emit it) counts as not given: ?? would keep the 0 and drop the discount.
+		$base = !empty($allowances[0]['basisAmount']) ? $allowances[0]['basisAmount'] : $priceWithoutDiscount;
 
 		if (!$base) {
 			return false;

### CODE REVIEW

**Comment by daGrumpf-bxp** [comment]
Verified on the bench: the reading is right and the fix does what it says.

`toFloat()` answers `null` for an absent `ram:BasisAmount` and `0.0` for `<ram:BasisAmount>0.00</ram:BasisAmount>`, so `??` only ever caught the first. A stated zero went through as the base, `if (!$base)` then sent the function back `false`, and the line landed on the invoice at its gross. `!empty()` catches both, and a real base — or a negative one — goes on as before. A zero base is meaningless next to a real BT-136 anyway: no percentage can be taken off nothing.

**Measured.** A line of 5 × 100.05 with a 10 % allowance of 50.03, `BasisAmount` 0.00 and `LineTotalAmount` 450.22, imported through `createSupplierInvoiceFromSource()` on Dolibarr 23.0.4 and read back from a second process:

| | invoice, VAT included |
|---|---|
| announced by the document | **540.26** |
| `main` | 600.30 — discount dropped, line at 500.25 |
| this branch | **540.26** |

Module suite file by file on **18.0.10, 19, 20, 21, 22, 23.0.4, 24.0.0 and 25**: 29 files OK on each, unchanged.

**One thing missing: a test.** `ReceivedInvoiceLinesTest.php` already pins the two neighbouring shapes of that base (`testADocumentStatingItsBaseIsUnchanged`, `testTheFallbackRebuildsTheBaseTheDocumentDidNotState`), and without a third the `??` comes back the next time someone tidies that line. This one sits next to them, passes on your branch and fails on `main` with *"Failed asserting that false is not false"* — measured on 18.0.10 and 24.0.0, feel free to take it as is:

```php
	/**
	 * A base stated as zero is a base the document did not state. Some issuers write BT-137 = 0.00 next
	 * to a real allowance amount, and the null coalescing kept that zero: the guard below it sent the
	 * discount back false and the line was imported at its gross.
	 *
	 * @return	void
	 */
	public function testABaseStatedAsZeroFallsBackTheSameWay()
	{
		global $db;

		$protocol = new CIIProtocol($db);

		$zeroBase = $protocol->parseInvoiceLines($this->discountedLine(5.0, 100.05, 50.03, 450.22, 0.00));
		$this->assertSame(0.0, $zeroBase[0]['lineAllowances'][0]['basisAmount'], 'BT-137 is read, and it is zero');

		$discount = $this->call('resolveLineDiscountPercent', array($zeroBase[0]['lineAllowances'], $zeroBase[0]['lineTotalAmount']));

		$this->assertNotFalse($discount, 'the discount is resolved and not dropped');
		$this->assertEqualsWithDelta(500.25, $discount['base'], 0.011, 'the amount before the allowance, as when BT-137 is absent');
		$this->assertEqualsWithDelta(10.001, $discount['percent'], 0.0001);
		$this->assertEqualsWithDelta(450.22, $this->importedLine($zeroBase[0])['rebuilt'], 0.011, 'and the line totals BT-131');
	}
```

Small thing while you are there: the replaced comment carried the `(issue #783)` reference, which is where that fallback comes from — worth keeping on the line.

No overlap with #848, which touches `resolveLineAmounts()` a thousand lines above; the two merge in either order.

PR #854, opened 2026-09-07
FIX: [einvoicing] The scheduled flow sync no longer reports "Unknown error"

nicolas-eoxia:
## What

A scheduled `EInvoicingDocumentSync` run that stops on a business error files this on the job card:

```
Synchronisation interrompue sur le flux # 54 / 100 (Flux i_XXXXXX)
Nombre total de flux ignorés (existants ou déjà traités): 2 - Total des nouveaux flux synchronisés: 51
Erreur inconnue
```

"Erreur inconnue" is not an error the module raised: it is the scheduler's fallback. `CronJob::run_jobs()` builds the job message from `$object->error` and `$object->errors`, never from `$object->output`, and when both are empty it substitutes `ErrorUnknown`.

`Document::cronSyncFlows()` fills `$this->output` with `$sync_result['messages']` — the recap — and drops `$sync_result['details']`, which is where the reason lives. So the one line that says *why* the run stopped is nowhere on the card:

```
ERROR_SYNCFLOW - Failed to synchronize flow i_XXXXXX: Failed to create supplier invoice
from E-invoice document for flowId: i_XXXXXX. Document "06XXXXXXXX" linked to line 000001
was not found in Dolibarr. [...]
```

It survives only in `llx_einvoicing_call.processing_result` and in the `_einvoicing` syslog, neither of which the job card links to. On an instance with syslog off, the operator is told a sync failed and given no way to find out why.

## How

- `syncFlows()` collects the messages that actually incremented `$error` into a new `errors` key of its return, next to the existing `details` (the whole transcript, skipped flows included). Both real providers, same two spots each.
- `cronSyncFlows()` hands that array to `$this->errors`, falling back on `details` for a third-party provider that only fills the older key.
- The no-provider case is set on `error` instead of `output`, so it reads once instead of twice — the scheduler concatenates the two.

`output` is untouched: the recap still shows as before, the cause is now appended instead of `Erreur inconnue`.

## Test

Replaying the scheduler's message-building block on a `Document` prepared the way `cronSyncFlows()` leaves it:

```
----- before -----
Synchronisation interrompue sur le flux # 54 / 100 (Flux i_XXXXXX)
Nombre total de flux ignorés: 2
Unknown error

----- after -----
Synchronisation interrompue sur le flux # 54 / 100 (Flux i_XXXXXX)
Nombre total de flux ignorés: 2
ERROR_SYNCFLOW - Failed to synchronize flow i_XXXXXX: Failed to create supplier invoice from
E-invoice document for flowId: i_XXXXXX. Document "06XXXXXXXX" linked to line 000001 was not
found in Dolibarr.
```

Found on a Dolibarr 22 instance.

### DIFF
--- a/einvoicing/class/document.class.php
+++ b/einvoicing/class/document.class.php
@@ -1601,6 +1601,11 @@ class Document extends CommonObject
 
 			if ($sync_result['res'] <= 0) {
 				$error++;
+				// The scheduler builds what it shows from $this->error and $this->errors, never from
+				// $this->output. Leaving them empty is what turns a precise business error into the bare
+				// "Unknown error" the job card ends up displaying, so hand it the messages that aborted the run.
+				// A third-party provider may only fill 'details', so fall back on it rather than on nothing.
+				$this->errors = $sync_result['errors'] ?? ($sync_result['details'] ?? array());
 				$errortype = 'errors';
 				if (!empty($sync_result['actions'])) {
 					$errortype = 'warnings';
@@ -1615,7 +1620,8 @@ class Document extends CommonObject
 			}
 		} else {
 			$error++;
-			$this->output = $langs->trans("NoPDPProviderConfigured");
+			// Set on error only, not on output: the scheduler concatenates the two and would show it twice.
+			$this->error = $langs->trans("NoPDPProviderConfigured");
 		}
 
 		$this->output = trim($this->output);
--- a/einvoicing/class/document.class.php
+++ b/einvoicing/class/document.class.php
@@ -1602,21 +1602,36 @@ class Document extends CommonObject
 			if ($sync_result['res'] <= 0) {
 				$error++;
 				// The scheduler builds what it shows from $this->error and $this->errors, never from
-				// $this->output. Leaving them empty is what turns a precise business error into the bare
-				// "Unknown error" the job card ends up displaying, so hand it the messages that aborted the run.
-				// A third-party provider may only fill 'details', so fall back on it rather than on nothing.
-				$this->errors = $sync_result['errors'] ?? ($sync_result['details'] ?? array());
-				$errortype = 'errors';
+				// $this->output. Leaving both empty is what turns a precise cause into the bare
+				// "Unknown error" the job card ends up displaying.
 				if (!empty($sync_result['actions'])) {
-					$errortype = 'warnings';
-					$this->output .= '<br>' . $langs->trans("EINVOICING_JOB_MANUAL_ACTION_REQUIRED") . '<br>';
+					// A business error already carries a message written for a human, and the technical
+					// line with it, inside the tooltip its picto opens on the card. Handing the transcript
+					// to the scheduler as well would print that same line a second time and in clear,
+					// under the very action the operator is being asked to carry out. What is missing here
+					// is a cause, not a copy of one: give the sentence that closes the list, and stop
+					// writing it above them.
+					$this->output .= '<br>';
 					foreach ($sync_result['actions'] as $action) {
 						$this->output .= "---<br>";
 						$this->output .= $action['businessmessage'] . '<br>';
 					}
-					$this->output = rtrim($this->output, '<br>');
+					// Not rtrim($output, '<br>'): rtrim strips characters, not a string, so it eats into
+					// the closing tags of the business message and leaves broken markup behind.
+					$this->output = preg_replace('/<br>$/', '', $this->output);
+					$this->error = $langs->trans("EINVOICING_JOB_MANUAL_ACTION_REQUIRED");
+				} else {
+					// Everything else has no message written for a human, so the transcript is what the
+					// operator gets. A third-party provider may only fill the older 'details' key, so fall
+					// back on it rather than on nothing.
+					$this->errors = $sync_result['errors'] ?? ($sync_result['details'] ?? array());
+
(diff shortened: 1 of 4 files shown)


daGrumpf-bxp:
Confirmed the premise against the core: `CronJob::run_jobs()` builds `$errmsg` from `$object->error` and `$object->errors` only, never from `$object->output`, and falls back to `ErrorUnknown` when both are empty. Same block, unchanged, from 18.0.10 to 24. Moving the cause into `$this->errors` is the right lever, and setting `NoPDPProviderConfigured` on `error` instead of `output` does avoid the duplicate, since the core concatenates the two.

One remark on coverage, one detail.

### The early returns of `syncFlows()` still land on "Unknown error"

Both providers leave the function well before the flow loop on the two most banal failures, and those returns carry neither `details` nor `errors`:

- `SuperPDPProvider.class.php:1896` / `EsalinkPDPProvider.class.php:937` — the access point answers something other than 200 (unreachable PA, expired token);
- `SuperPDPProvider.class.php:1938` / `EsalinkPDPProvider.class.php:984` — the query listing the already-processed flows fails;
- plus `EsalinkPDPProvider.class.php:906`, the pre-count when `limit == 0`.

They all `return array('res' => 0, 'messages' => $results_messages);`. Since `??` only tests for `null` and the key is absent, `$sync_result['errors'] ?? ($sync_result['details'] ?? array())` yields an empty array, `$this->error` stays empty, and the job card is back to `ErrorUnknown`. Replaying the core block on the three return shapes a failed run can produce, with this branch applied:

```
----- per-flow error (the case of the description)
Synchronisation interrompue sur le flux # 54 / 100 (Flux i_458872)
ERROR_SYNCFLOW - Failed to synchronize flow i_458872: Document "0641495329" not found
----- HTTP != 200
Failed to retrieve flows for synchronization. (HTTP 500)
Unknown error
----- DB query failed
Failed to retrieve from database the list of flows already processed. ...
Unknown error
```

Those three spots already fill `$this->errors[]` on the provider object, right next to `$results_messages[]`, and nothing ever reads it back. Adding `'errors' => $this->errors` to the returns would close the gap where the message is already written.

### The business-error case now prints the technical message twice

For `THIRDPARTY_NOT_FOUND` and `PRODUCT_NOT_FOUND` the `ERROR_SYNCFLOW` line also goes into `$results_messages`, so it now reaches `errors` as well. But it is already on the card, inside the tooltip `textwithpicto()` appends to the business message (`SuperPDPProvider.class.php:2017` and `2044`, `EsalinkPDPProvider.class.php:1075` and `1102`). The operator gets the manual action, then the same raw technical line underneath. Still better than `ErrorUnknown`, but on the one path where the module has a message written for a human, the raw one now competes with it.

Nothing else: the loop breaks on the first `$error++`, so `errors` holds at most one entry and cannot push the recap out of `MAXIMUM_LENGTH_FOR_LASTOUTPUT_FIELD`.

(Unrelated to this change, spotted while reading: `$errortype` in `cronSyncFlows()` is computed and never used, the only line reading it is commented out.)

nicolas-eoxia:
Both points taken, fixed in 4944378. Thanks for replaying the core block on the other return shapes — the gap was real and I had only looked at the loop.

### The early returns

Returning the cause from the three spots, in both providers. One correction to the suggested lever, though: `'errors' => $this->errors` would carry more than this run.

`$this->errors` is never reset — not in `syncFlows()`, not anywhere in the provider — and it is appended to from about fifteen places, including `getAccessToken()`. A token that fails once and is obtained on the retry leaves `FailedToRetrieveAccessToken` in there, and the run would then report a failure that did not happen. So the messages go into the run-local array the previous commit already introduced, next to `$results_messages`, and `$this->errors[]` is left as it was.

That surfaced a second duplicate on the same path: those returns put their one message in `messages` as well, so with a cause now filled the card printed it twice, once from `output` and once from `$errmsg`. The message is removed from `output` when the cause already carries it:

```
----- HTTP != 200
Failed to retrieve flows for synchronization. (HTTP 500)
```

### The business-error case

You are right, and the transcript is the wrong thing to hand over there. The business message is the one written for a human and it already carries the technical line in the tooltip. The scheduler only needs a non-empty cause, so it gets the sentence that closes the list, `EINVOICING_JOB_MANUAL_ACTION_REQUIRED`, which is no longer written above the actions:

```
----- business error (THIRDPARTY_NOT_FOUND)
Sync aborted on flow # 12 / 100
---
Cannot find the thirdparty of the imported invoice (Name: ACME)   [i]
Manual action is required. The job has been interrupted.
```

The technical line stays one click away, where you put it, and appears once.

### Two things met on the way

- `rtrim($this->output, '<br>')` strips *characters*, not the string: on a business message ending in `</div></div><br>` it removes the `<br>`, then the `>` of the `</div>`, and leaves broken markup that swallows what follows. Replaced with a `preg_replace` anchored on the end.
- `$errortype` is gone.

`$this->error` is set instead of `$this->errors[]` for the business case, which the core handles: `errors` stays empty, `in_array()` on it is false, and `$errmsg` takes `$object->error` as is — `<br>` included, where the array would have been imploded on `, `.

The five shapes a failed run can now produce were replayed through the core block, including the no-provider one.

(Unrelated to the review: the sample data in this description and in #855 has been anonymized since your comment — the reference quoted in it is a real subscriber number.)

PR #855, opened 2026-09-07
FIX: [einvoicing] A telecom invoice no longer stops the whole synchronization

nicolas-eoxia:
## What

A scheduled sync stopped on a received telecom invoice with:

```
Document "06XXXXXXXX" linked to line 000001 was not found in Dolibarr. [...]
you must manually create the invoice using the supplier invoice reference "06XXXXXXXX".
```

That reference is not an invoice. It is the mobile line the invoice bills. The document says so:

```xml
<ram:AdditionalReferencedDocument>
  <ram:IssuerAssignedID>06XXXXXXXX</ram:IssuerAssignedID>
  <ram:TypeCode>130</ram:TypeCode>
  <ram:ReferenceTypeCode>AWV</ram:ReferenceTypeCode>
</ram:AdditionalReferencedDocument>
```

`TypeCode 130` with a `ReferenceTypeCode` is BT-128, *Invoiced object identifier*: what the line bills — a phone number, a meter, a subscription. Never a document this Dolibarr could hold.

`createSupplierInvoiceLinesFromSource()` treated every `AdditionalReferencedDocument` of a line as a supplier invoice to find. It read `$lineRefDocType` and the parser filled `referenceTypeCode`, but neither was ever consulted.

Two consequences:

- **No telecom or utility invoice could be imported at all** — they all carry one identifier per line.
- **From the scheduler, the first one aborted the entire batch.** The flows behind it never synchronized, and since the cause never goes away, every following run stopped at the same place. Following the message and creating an invoice named after a phone number would not have helped: the next monthly invoice carries the same identifier, and each line of a fleet would ask for one more.

## How

`TypeCode 130` alone cannot carry the decision — **this module emits its own deposit lines with it**, building an `AdditionalReferencedDocument` with `TypeCode` 130. Keying on it would break the round trip of our own invoices.

The `ReferenceTypeCode` can: it qualifies the identifier as a billed object, and the deposit references emitted here have none. That single distinction is extracted into `lineRefDocIsInvoicedObjectIdentifier()`, tested through reflection the way the neighbouring decision helpers are.

## Test

Two cases added to `ReceivedInvoiceLinesTest`: the telecom shape is recognized as a billed object, and the deposit shape — same `TypeCode 130`, no `ReferenceTypeCode` — still goes through the lookup. The test number comes from the range the ARCEP reserves for fiction, so no test data points at a real subscriber.

Checked against a real flow from a French telecom issuer, imported inside a rolled-back transaction:

```
before -> res: -1   Document "06XXXXXXXX" linked to line 000001 was not found in Dolibarr.

after  -> res: <id>  Supplier Invoice created or updated
          ref_supplier=<the issuer's own invoice number>  HT=29.16  TTC=34.99  lines=3
```

The invoice imports under its real number instead of being refused, with the amounts the document announces.

Dolibarr 22, Factur-X CIUS.

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -1223,6 +1223,25 @@ class CIIProtocol extends AbstractProtocol
 
 
 	/**
+	 * Tell whether a document referenced by a line is the identifier of the object being billed
+	 * rather than a document this Dolibarr could hold.
+	 *
+	 * A ReferenceTypeCode (BT-128-1) qualifies the scheme of an "Invoiced object identifier"
+	 * (BT-128): what the line bills - a phone number, a meter, a subscription - and never a document.
+	 * Telecom and utility issuers put one on every line, so looking those up as supplier invoices
+	 * fails each of their invoices, and from the scheduler it stops the whole synchronization on the
+	 * first one. The deposit references this module itself emits carry no ReferenceTypeCode, which is
+	 * what keeps them out of this test and on the lookup path.
+	 *
+	 * @param 	array 	$refDoc		One entry of the line's additionalRefDocs, as parseInvoiceLines() returns it
+	 * @return 	bool				True when the reference designates the billed object, not a document
+	 */
+	protected function lineRefDocIsInvoicedObjectIdentifier(array $refDoc): bool
+	{
+		return !empty($refDoc['referenceTypeCode']);
+	}
+
+	/**
 	 * Add lines to a supplier invoice from e-invoice parsed lines
 	 *
 	 * @param 	FactureFournisseur 	$supplierInvoice						The supplier invoice to add lines on
@@ -1252,6 +1271,10 @@ class CIIProtocol extends AbstractProtocol
 					$lineRefDocId = $refDoc['IssuerAssignedID'] ?? null;
 					$lineRefDocType = $refDoc['typeCode'] ?? null;
 					$lineRefDocDate = $refDoc['issueDate'] ?? null;
+
+					if ($this->lineRefDocIsInvoicedObjectIdentifier($refDoc)) {
+						continue;
+					}
 
 					$linkedObjectId = SupplierInvoiceHelper::findIdByRef($lineRefDocId, (int) $parsedLine['supplierId']);
 					if ($linkedObjectId < 0) {
--- a/einvoicing/test/phpunit/ReceivedInvoiceLinesTest.php
+++ b/einvoicing/test/phpunit/ReceivedInvoiceLinesTest.php
@@ -1344,4 +1344,95 @@ class ReceivedInvoiceLinesTest extends CommonClassTest
 			array('indicator' => 'false', 'actualAmount' => 5.00, 'reason' => 'Commercial gesture', 'rateApplicablePercent' => 20.0),
 		)));
 	}
+
+	/**
+	 * Call CIIProtocol::lineRefDocIsInvoicedObjectIdentifier() through reflection: pure decision
+	 * logic, no DB access and no side effect.
+	 *
+	 * @param	CIIProtocol		$protocol	Protocol instance
+	 * @param	array			$refDoc		One entry of the line's additionalRefDocs
+	 * @return	bool						What the caller would decide
+	 */
+	private function callLineRefDocIsInvoicedObjectIdentifier(CIIProtocol $protocol, array $refDoc)
+	{
+		$method = new ReflectionMethod(CIIProtocol::class, 'lineRefDocIsInvoicedObjectIdentifier');
+		$method->setAccessible(true);
+
+		return $method->invoke($protocol, $refDoc);
+	}
+
+	/**
+	 * A telecom issuer writes the billed line on each invoice line as an Invoiced object identifier
+	 * (BT-128): TypeCode 130 and, what tells it apart, a ReferenceTypeCode qualifying its scheme.
+	 * That identifier is a phone number, not a document, and it used to be looked up as a supplier
+	 * invoice - failing every such invoice, and stopping the scheduled synchronization on the first.
+	 *
+	 * The number below is taken from the 0639 98 00 00 - 0639 98 99 99 range the ARCEP reserves for
+	 * fiction, so that no test data points at a real subscriber.
+	 *
+	 * @return	void
+	 */
+	public function testAnInvoicedObjectIdentifierIsNotLookedUpAsADocument()
+	{
+		global $db;
+
+		$protocol = new CIIProtocol($db);
+
+		$xml = $this->documentWithLine('
+      <ram:AssociatedDocumentLineDocument><ram:LineID>000001</ram:LineID></ram:AssociatedDocumentLineDocument>
+      <ram:SpecifiedTradeProduct><ram:Name>Forfait 250Go 5G SIM Seule</ram:Name></ram:SpecifiedTradeProduct>
+      <ram:SpecifiedLineTradeSettlement>
+        <ram:AdditionalReferencedDocument>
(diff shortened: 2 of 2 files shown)


einvoicing/class/protocols/CIIProtocol.class.php
```diff
@@ -1222,6 +1222,25 @@ protected function doCreateSupplierInvoiceFromSource($file, $readableViewFile, $
...
+	 * fails each of their invoices, and from the scheduler it stops the whole synchronization on the
+	 * first one. The deposit references this module itself emits carry no ReferenceTypeCode, which is
+	 * what keeps them out of this test and on the lookup path.
+	 *
+	 * @param 	array 	$refDoc		One entry of the line's additionalRefDocs, as parseInvoiceLines() returns it
+	 * @return 	bool				True when the reference designates the billed object, not a document
+	 */
+	protected function lineRefDocIsInvoicedObjectIdentifier(array $refDoc): bool
+	{
+		return !empty($refDoc['referenceTypeCode']);
```
eldy, line 1241: No test on $refDoc['typeCode'] = or != 130 , referenceTypeCode alone is enough to say if it is a ref of document or other ?
Also if $refDoc['referenceTypeCode'] = "AFL", it seems it is a ref of document, not object ?

nicolas-eoxia: You are right on both counts, and the second one is a real false positive.

**On AFL** — UNTDID 1153 `AFL` is "Associated invoices: a number that identifies
associated invoices". So a line carrying it does point at documents, and my check
would silently skip it. And it is not an isolated case: 1153 has 806 codes, most of
them documentary — `IV` (Invoice document identifier), `ACE` (Related document
number), `ON` (order), `CT` (contract), `DQ` (delivery note)... So
`!empty($refDoc['referenceTypeCode'])` is not a valid discriminator; it only happens
to work on the telecom case I had at hand.

**On TypeCode 130** — you are right that this is the conformant test. At line level
EN 16931 defines only BT-128 (Invoiced object identifier), with `TypeCode` fixed to
130 and `ReferenceTypeCode` (BT-128-1) optional. What made me avoid it is that this
module emits its own deposit references at line level with `TypeCode 130` as well
(`CIIProtocol::buildLineItem()`, line 2698) — which is itself non-conformant, it puts
an invoice reference in the BT-128 slot. Keying the read side on 130 would stop
linking deposits on every invoice this module has already issued. My
`testADepositReferenceIsStillLookedUp` locks that non-conformance in instead of
flagging it, which is the wrong thing for a test to do.

**What I propose instead** — stop deciding from the code list at all, and use BT-128-1
only to decide whether a *missing* reference is fatal:

- reference found in Dolibarr → it is a document, link it (deposits, and a genuine
  `AFL`, keep working);
- not found **and** BT-128-1 present → invoiced object identifier, skip silently (the
  telecom case);
- not found and no BT-128-1 → hard error, current behaviour unchanged.

That fixes the original bug (one telecom invoice aborting the whole scheduled
synchronization), no longer claims that `referenceTypeCode` means "object", and does
not depend on reading 1153 semantics.

Separately, the deposit emission using `TypeCode 130` at line level looks worth fixing
on its own — I would rather open a follow-up issue for it than widen this PR.

If you prefer the fully conformant route instead (test 130 on read + fix the deposit
emission, with a compatibility fallback for invoices already issued), say so and I
will do that.

daGrumpf-bxp:
Heads-up: **#897** fixes the same loop (BT-128 at line level), opened yesterday, with a broader rule — it moves the loop into a `resolveLineLinkedDocuments()` and steps over *any* reference matching no supplier invoice, reporting it in the import messages. Each merges cleanly on `main`, but one after the other they conflict in `CIIProtocol.class.php` (`git merge-tree`, both orders), and neither thread mentions the other.

Bench, Dolibarr 18→25, both branches merged on current main: PHPUnit green on both. Two differences measured on Dolibarr 22, with a real deposit invoice and the import rolled back — the second one corrects an earlier version of this comment, which claimed neither PR touched the deposit path:

- a reference carrying no `ReferenceTypeCode` (a contract number) still aborts the whole synchronization here, where #897 imports it;
- and skipping *before* the lookup costs the deposit deduction: a line pointing at an existing deposit gets no discount at all as soon as the reference is qualified — `AWV`, and `AFL` too. That is the false positive raised in review above, and on a deposit it does not merely fail to link, it drops money from the imported invoice.

Which is why the rule you describe in your answer — look up first, and use BT-128-1 only to decide whether a *missing* reference is fatal — is the form that holds. Measured after the lookup it changes nothing on the deposit path, and it still keeps a telecom invoice from adding one import message per line. The suggestion of carrying this over as a `continue` above #897's `resolveLineLinkedDocuments()` was the pre-lookup form, and I withdraw it.

daGrumpf-bxp:
Opened #911 with a fix for this defect after #897 was closed by its author, since it is still on `main`.

It does not carry this rule as it stands. Measured on a bench (Dolibarr 22, real import rolled back, a deposit invoice recorded beforehand): deciding from the presence of a `ReferenceTypeCode` **before** the lookup costs the deposit deduction — a deposit referenced with `AWV`, and with `AFL` too, gets no discount at all. That is the false positive raised in the review above, and on a deposit it does not merely fail to link, it drops money from the imported invoice.

So #911 looks the reference up first, and uses BT-128-1 only to decide whether a *missing* reference is reported loudly or stepped over silently — which is the rule you described yourself in your answer of 09/09. The reading that BT-128-1 is worth reading at all is yours, and #911 says so.

nicolas-eoxia:
Closing this in favour of #911.

@daGrumpf-bxp's bench found a defect in this PR that neither the review nor I had caught, and I have confirmed it against the diff here: the `continue` sits **before** `SupplierInvoiceHelper::findIdByRef()`, so `$is_deposit_line` and `$fk_remise` are never set. A line pointing at an existing deposit therefore loses its discount as soon as the reference carries any `ReferenceTypeCode` — `AWV`, and `AFL` too. On a deposit that is not a missing link, it is a wrong amount in the imported invoice. That alone disqualifies the code as it stands.

The underlying reason is the one @eldy raised: `!empty($refDoc['referenceTypeCode'])` is not a valid discriminator. UNTDID 1153 has 806 codes and most of them are documentary — `IV`, `ACE`, `ON`, `CT`, and the `AFL` he pointed at.

I proposed the corrected rule in my answer of 09/09 — look the reference up first, and use BT-128-1 only to decide whether a *missing* reference is fatal — but never pushed it here. #911 implements it, and goes further: when the document declares an amount already paid (BT-113), a missing reference postpones the flow instead of failing it, which handles the missing-deposit case this PR did not address at all. It is the better fix, and it is already reviewed against a real import.

Rather than push a duplicate implementation onto the same file and conflict with it, I am closing this. Nothing here is lost that #911 does not cover.

The one thing that would have been dropped is the follow-up I promised in the review: this module emits its own deposit references at line level with `TypeCode 130`, in the BT-128 slot, which is what makes `130` unusable as a discriminator on the reading side. Opened as #912.

Thanks @eldy for the review, and @daGrumpf-bxp for benching it rather than taking it at face value — the deposit regression would have shipped otherwise.

daGrumpf-bxp:
Thanks for the diff check and for closing it that way — and for #912, which is the part I would not have thought to open.

One thing from #897 has been carried into #911 since: the guard on an entry whose `IssuerAssignedID` is empty. `findIdByRef()` answers `0` for an empty reference exactly as for one it cannot find, so on a document declaring an amount already paid the flow was postponed for ever — nothing will ever make an empty reference resolvable. Measured on the branch, and credited there.

So between the three of us the reading is now: look the reference up first (yours), do not abort the batch on one that is not ours (#897's), skip an entry that identifies nothing (#897's too), and postpone rather than import short when BT-113 says money is at stake.

Issue #449, opened 2026-07-24
[Einvoicing] - Sync blocked — "Original" document not found in Esalink (documentType)


biedseb1:
During a synchronization, the system attempted to retrieve the original document. However, this original document did not exist in Esalink (also absent on the Esalink platform itself, not just on the integration side). Only the "Converted" version of the document was available.

Result: the synchronization remained blocked and could not proceed, with no way to bypass it.

Expected behavior:
The synchronization should be able to continue (either by automatically falling back to the converted version, or by raising a clear error) when the original document is unavailable, rather than remaining blocked indefinitely.

Workaround applied:

Temporarily changed the docType to "Converted".
Re-ran the synchronization → success.
Reverted the docType back to "Original" once the sync passed.
Final status: resolved.

Impact:

Synchronization blocked with no way to bypass it unless the workaround is applied manually.
Requires manual intervention on the docType, which is not a sustainable/automatable solution.

### Module

--Please select a module--

### Dolibarr Version

22.0.4

### Steps to reproduce the behavior and expected behavior

Steps to reproduce:

Trigger a synchronization on a document whose original is missing from Esalink (only the converted version is available).
Observe that the sync gets stuck while trying to retrieve the original document.
Check on the Esalink platform: the original is indeed absent, only the converted version exists.

PR #858, opened 2026-09-07
FIX #857: EsalinkPDPProvider does not record an incoming supplier invoice lifecycle status

### LINKED ISSUE #857: eInvoicing: CDAR status code 212 failed with Hubtimize
### Bug

`SuperPDPProvider::syncFlow()` handles an **incoming** `SupplierInvoiceLC` flow (a vendor reporting
a lifecycle status - "Cashed in" / 212 above all - on an invoice this Dolibarr instance never sent a
flow for) by routing it to `processIncomingSupplierInvoiceStatus()`, which resolves the target
invoice by vendor reference instead of by `flowId`.

`EsalinkPDPProvider::syncFlow()` has no equivalent guard in its own `"SupplierInvoiceLC"` case. Every
incoming flow falls through to the code path built for flows **we** send
(`EInvoicing::fetchStatusMessages($flowId)`), which only resolves a `flowId` recorded in
`llx_einvoicing_lifecycle_msg` when the LC message was sent from Dolibarr - a row an incoming flow
never has. The lookup fails every time, and the flow is skipped with:

Flow <flowId> skipped: Failed to fetch status messages for flowId: <flowId>

Since the flow is still stored in `llx_einvoicing_document` regardless of the outcome (`syncFlow()`
unconditionally calls `$document->create($user)`), the pre-check of the next synchronization run
treats it as "already known" and discards it silently - it is never retried.

**Net effect on EsalinkPDP**: a vendor's payment/lifecycle status on a supplier invoice this instance
received is never recorded, and the failure is invisible on every synchronization after the first one.

### Module

einvoicing

### Dolibarr Version

23.0

### Environment PHP

php8.3

### Environment Database

_No response_

1. Configure the module with `EINVOICING_PDP=ESALINK`.
2. Have a vendor send an incoming `SupplierInvoiceLC` CDAR flow (e.g. status 212 "Cashed in") on a
   supplier invoice this Dolibarr received from that vendor.
3. Run the synchronization (cron or manual).
4. Observe `"Flow <flowId> skipped: Failed to fetch status messages for flowId: <flowId>"` in the
   processing result / call log, and no status recorded on the supplier invoice.

### Expected behavior

The vendor's status should be attached to the matching supplier invoice, the same way it already is
on SuperPDP.

### Evidence

- `class/providers/SuperPDPProvider.class.php`, `case "SupplierInvoiceLC":` — guards on
  `$document->flow_direction == 'In'` and calls `processIncomingSupplierInvoiceStatus()`.
- `class/providers/EsalinkPDPProvider.class.php`, `case "SupplierInvoiceLC":` — no such guard, always
  calls `fetchStatusMessages($flowId)`.
- `processIncomingSupplierInvoiceStatus()` and `findSupplierInvoiceByVendorReference()` are currently
  `private` methods of `SuperPDPProvider` only, so `EsalinkPDPProvider` cannot reuse them as they
  stand.

### Suggested fix

Move `processIncomingSupplierInvoiceStatus()` and `findSupplierInvoiceByVendorReference()` to
`AbstractPDPProvider` (neither uses anything SuperPDP-specific) and add the same
`flow_direction == 'In'` guard to `EsalinkPDPProvider::syncFlow()`'s `"SupplierInvoiceLC"` case, so
both providers share one implementation and cannot drift apart again. A patch implementing this is
ready - PR to follow.

### Workaround for an already-skipped flow

Delete the corresponding row in `llx_einvoicing_document` (`flow_id = '<flowId>'`) so the next
synchronization run no longer treats it as already known, then re-run the sync.

comaiteseb:
Fixes #857

## Problem

`SuperPDPProvider::syncFlow()` already guards its `"SupplierInvoiceLC"` case for an **incoming** flow
- a vendor reporting a lifecycle status (cash-in, 212, above all) on an invoice we never sent a flow
for - and routes it to `processIncomingSupplierInvoiceStatus()`, which resolves the target invoice by
vendor reference instead of by `flowId`.

`EsalinkPDPProvider::syncFlow()` has no equivalent guard. Every incoming `SupplierInvoiceLC` flow falls
through to `EInvoicing::fetchStatusMessages($flowId)`, the path built for flows **we** send, which only
resolves a `flowId` recorded in `llx_einvoicing_lifecycle_msg` when the LC message was sent from
Dolibarr - a row an incoming flow never has. The lookup fails every time:

Flow <flowId> skipped: Failed to fetch status messages for flowId: <flowId>

`syncFlow()` stores the flow in `llx_einvoicing_document` regardless of outcome, so the next
synchronization's pre-check treats it as already known and silently discards it - it is never retried,
and the vendor's status is never recorded on the supplier invoice.

## Fix

- Move `processIncomingSupplierInvoiceStatus()` and `findSupplierInvoiceByVendorReference()` from
  `SuperPDPProvider` to `AbstractPDPProvider`. Neither uses anything SuperPDP-specific (only
  `$this->callApi()`, already abstract on the base class, and generic Dolibarr/DB access), so both
  providers now share one implementation instead of being able to drift apart again the way they
  already have once.
- Add the same `$document->flow_direction == 'In'` guard to `EsalinkPDPProvider::syncFlow()`'s
  `"SupplierInvoiceLC"` case, calling the now-shared `processIncomingSupplierInvoiceStatus()`.

No behavior change on SuperPDP - it keeps calling the same code, now inherited instead of declared
locally.

## Note for a flow already skipped by the bug

A flow already recorded with the old failure is not retried on its own - deleting its row in
`llx_einvoicing_document` lets the next synchronization process it through the corrected path.


einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,176 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+	 */
+	protected function processIncomingSupplierInvoiceStatus($flowId, $document, $einvoicing)
+	{
+		global $db;
+
+		require_once DOL_DOCUMENT_ROOT . '/fourn/class/fournisseur.facture.class.php';
+		dol_include_once('einvoicing/class/utils/CdarHandler.class.php');
+
+		$flowResource = 'flows/' . $flowId . '?' . http_build_query(array('docType' => 'Original'));
+		$flowResponse = $this->callApi($flowResource, "GET", false, array('Accept' => 'application/octet-stream'));
```
daGrumpf-bxp, line 1095: `callApi()` here bypasses `fetchFlowData()` from this same class, and more importantly it has no `Converted` fallback. Some flows have no `Original` document on the access point, only the converted copy; `EsalinkPDPProvider::syncFlow()` already hit that in its `"CustomerInvoiceLC"` case and notes that the sync then stops on that flow *and on every flow behind it*. Both documents carry the same CDAR.

```suggestion
		$flowResponse = $this->fetchFlowData($flowId, 'Original');
		if ($flowResponse['status_code'] != 200) {
			// Some flows have no 'Original' on the access point, only the converted copy. Both carry
			// the same CDAR, so fall back rather than fail the flow (and every flow behind it).
			$flowResponse = $this->fetchFlowData($flowId, 'Converted');
		}
```

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,176 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		if ($flowResponse['status_code'] != 200) {
+			return array('res' => -1, 'message' => "Failed to retrieve flow details for flowId: " . $flowId);
```
daGrumpf-bxp: `res = -1` on a transient platform failure is worse than it looks: the caller does `break`, `syncFlow()` then runs `$document->create($user)` unconditionally, and the next run's pre-check discards the now known `flow_id`. So a 500 or a timeout on this single GET loses the vendor status for good, with no trace but one line in the processing result.

The docblock right above already promises the opposite behaviour (*"Never returns a negative result for a status it cannot attach ... The flow is stored either way"*). Either keep every outcome at `0`/`1` here, or let the caller `return` instead of `break` so the flow is not recorded and is retried on the next run - see my comment on `EsalinkPDPProvider.class.php`. Same remark for the parse failure below (line 1103).

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,176 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		$vendorLegalId = isset($refDoc['IssuerTradeParty']['GlobalID']) ? (string) $refDoc['IssuerTradeParty']['GlobalID'] : '';
+
+		$document->tracking_idref = $vendorReference;
+
+		if ($vendorReference === '') {
+			dol_syslog(__METHOD__ . " FlowId " . $flowId . " carries no IssuerAssignedID, nothing to attach the status to", LOG_WARNING);
+			return array('res' => 0, 'message' => "FlowId " . $flowId . " - Vendor lifecycle status with no invoice reference");
+		}
+
+		$supplierInvoiceId = $this->findSupplierInvoiceByVendorReference($vendorReference, $vendorLegalId);
```
daGrumpf-bxp, line 1126: Optional, but it turns a heuristic into an exact link. The flow metadata of an incoming lifecycle message carries `trackingId` = the `flowId` of the flow it comments on (verified on Hubtimize: the incoming 212 carries the `flowId` of the received supplier invoice flow). That flow is already stored in `llx_einvoicing_document` with `fk_element_type = 'invoice_supplier'` and `fk_element_id` set when the supplier invoice was created, so one lookup resolves the invoice with no reference matching at all:

```php
// $document->tracking_idref still holds the metadata trackingId on entry (overwritten line 1119)
SELECT fk_element_id FROM llx_einvoicing_document
 WHERE flow_id = <trackingId> AND fk_element_type = 'invoice_supplier' AND fk_element_id > 0
```

It cannot be guaranteed on every platform (on SuperPDP `trackingId` may hold an invoice ref instead), so `findSupplierInvoiceByVendorReference()` stays as the fallback. Worth it because `ref_supplier` is editable, not unique across vendors, and absent when the invoice was keyed in by hand rather than imported.

einvoicing/class/providers/EsalinkPDPProvider.class.php
```diff
@@ -1634,6 +1634,17 @@ public function syncFlow($flowId, $call_id = null)
...
 
+				// An incoming one is a status the VENDOR issues about one of its own invoices - "Cashed in"
+				// (212) above all, the answer to the payment we reported with a 211. We never sent it, so it
+				// has no row in einvoicing_lifecycle_msg and the flowId lookup below cannot resolve it.
+				if ($document->flow_direction == 'In') {
+					$resIncoming = $this->processIncomingSupplierInvoiceStatus($flowId, $document, $einvoicing);
+
+					$returnRes = $resIncoming['res'];
+					$returnMessage = $resIncoming['message'];
+					break;
```
daGrumpf-bxp, line 1645: A negative result is recorded all the same by `$document->create($user)` further down, and the next `syncFlows()` pre-check then skips this `flow_id` forever - so a transient failure of the CDAR download is not retried, it is lost. Returning early leaves the flow unrecorded, which is exactly what the `"CustomerInvoiceLC"` case above already does on its own error paths.

```suggestion
					if ($resIncoming['res'] < 0) {
						// Left unrecorded on purpose: a stored flow is treated as known and never retried.
						return $resIncoming;
					}

					$returnRes = $resIncoming['res'];
					$returnMessage = $resIncoming['message'];
					break;
```

The same applies to `SuperPDPProvider::syncFlow()`, whose case block this PR leaves unchanged.

daGrumpf-bxp (commented):
Thanks for picking this up. The move into `AbstractPDPProvider` is the right call, and `flow_direction` is safe to rely on here: `EsalinkPDPProvider::syncFlow()` already uses it in its `"SupplierInvoice"` case, and both providers fill it from the same `$flowData['flowDirection']`.

I checked the case against a real incoming `SupplierInvoiceLC` on a production Hubtimize account: `flowType: SupplierInvoiceLC`, `flowDirection: In`, `acknowledgement: Ok`, carrying a CDAR with `ProcessConditionCode 212`, an `IssuerAssignedID` and an `IssuerTradeParty/GlobalID` - which `findSupplierInvoiceByVendorReference()` does resolve. The diagnosis holds end to end, and the status was indeed lost by the current code.

Three changes I would make before merge (inline), all around the same weak spot: since `syncFlow()` calls `$document->create($user)` whatever the outcome, and `syncFlows()` discards an already known `flow_id`, **any failure returned by this helper condemns the flow permanently**.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Good catch. Fixed in c9783303: SuperPDPProvider::syncFlow() now returns instead of breaking when processIncomingSupplierInvoiceStatus() comes back negative, so the flow is left unrecorded and retried on the next run (matches the fix already applied on EsalinkPDPProvider via the suggestion). 
Updated the docblock on processIncomingSupplierInvoiceStatus() to state the actual contract: -1 is a transient failure that must not be stored, 0 is a resolved-but-unattachable outcome that is stored and not retried, 1 is attached.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Implemented in c9783303. Added findSupplierInvoiceByFlowId() on AbstractPDPProvider, doing the exact llx_einvoicing_document lookup on flow_id + fk_element_type = 'invoice_supplier' + fk_element_id > 0 you described. processIncomingSupplierInvoiceStatus() now captures the incoming flow's metadata trackingId before it gets overwritten by the CDAR's IssuerAssignedID, tries this exact match first, and falls back to findSupplierInvoiceByVendorReference() only when it misses — including in the previously-dead case where IssuerAssignedID is empty, which the exact match can now rescue.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,230 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+
+		$flowResponse = $this->fetchFlowData($flowId, 'Original');
+		if ($flowResponse['status_code'] != 200) {
+			// Some flows have no 'Original' on the access point, only the converted copy. Both carry
+			// the same CDAR, so fall back rather than fail the flow (and every flow behind it).
+			$flowResponse = $this->fetchFlowData($flowId, 'Converted');
+		}
+		if ($flowResponse['status_code'] != 200) {
+			return array('res' => -1, 'message' => "Failed to retrieve flow details for flowId: " . $flowId);
+		}
```
daGrumpf-bxp, line 1098: This is the transient one, and it is the one that has to come back. As written it makes `syncFlow()` return `-1` bare, and `syncFlows()` then counts an error and breaks out of its flow loop - the batch stops here and everything behind this flow waits for a run that will abort in the same place.

Marking it in the helper rather than in the two providers keeps your `if ($resIncoming['res'] < 0) { return $resIncoming; }` exactly as it is, and gets the two `SupplierInvoiceLC` cases right at once. Note the caller reads `$res['actioncode']` and `$res['action']` unconditionally to fill `$actions[]`, so both have to be there (see the existing postponed return, EsalinkPDPProvider.class.php:1362-1370).

```suggestion
		if ($flowResponse['status_code'] != 200) {
			// Nothing is stored for this flow, so it must come back: 'postponeflow' is what tells
			// syncFlows() to carry on with the flows behind it instead of aborting the whole batch.
			return array(
				'res' => -1,
				'postponeflow' => 1,
				'message' => "Failed to retrieve flow details for flowId: " . $flowId,
				'actioncode' => 'CANT_READ_INCOMING_LIFECYCLE_STATUS',
				'actionurl' => '',
				'action' => '',
				'businessmessage' => $langs->trans('CantReadTheStatusSentByTheVendor', $flowId)
			);
		}
```

That needs `global $db, $langs;` at the top of the method and one `en_US` key (wording yours), along the lines of `CantReadTheStatusSentByTheVendor=The status sent by the vendor in flow %s could not be read. Nothing was stored, the flow will be retried on the next synchronization.` - the caller's default message would otherwise say the *invoice* could not be imported, which is not what happened.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,230 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+
+		$cdarHandler = new CdarHandler($db);
+		$cdarDocument = $cdarHandler->readFromString($flowResponse['response']);
+		if (empty($cdarDocument) || empty($cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument'])) {
+			return array('res' => -1, 'message' => "FlowId: " . $flowId . " - Failed to parse CDAR document");
+		}
```
daGrumpf-bxp, line 1104: This one is not transient: a CDAR that does not parse, or that carries no `ReferenceReferencedDocument`, will not parse any better next week. With the early return in the providers it is never stored, so it is read again on every synchronization for ever - and once the suggestion above is in, postponed and reported to the user on every run too.

`0` is what your own docblock prescribes for it ("resolved but stored without attaching"), and it is the same treatment as the missing `IssuerAssignedID` a few lines below.

```suggestion
		if (empty($cdarDocument) || empty($cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument'])) {
			// Not transient, unlike the HTTP failure above: a CDAR that does not parse never will. Stored
			// so the synchronization does not read it again, and report it, on every run.
			dol_syslog(__METHOD__ . " FlowId " . $flowId . " carries no readable CDAR", LOG_WARNING);
			return array('res' => 0, 'message' => "FlowId: " . $flowId . " - Failed to parse CDAR document");
		}
```

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,230 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		if ($supplierInvoiceId <= 0) {
+			dol_syslog(__METHOD__ . " No supplier invoice found for vendor reference " . $vendorReference . " (vendor " . $vendorLegalId . "), flowId " . $flowId, LOG_WARNING);
+			return array('res' => 0, 'message' => "FlowId " . $flowId . " - No supplier invoice matching the vendor reference " . $vendorReference);
+		}
+
+		$supplierInvoice = new FactureFournisseur($this->db);
+		if ($supplierInvoice->fetch($supplierInvoiceId) <= 0) {
+			return array('res' => 0, 'message' => "FlowId " . $flowId . " - Failed to load supplier invoice id " . $supplierInvoiceId);
+		}
```
daGrumpf-bxp, line 1135: Minor, but the exact match now shadows the fallback it was meant to complete: when `findSupplierInvoiceByFlowId()` returns an id whose invoice cannot be read, the vendor reference is never tried, where before this PR it would have found the invoice.

The ordinary path is safe - the deletion trigger sets `fk_element_id = 0` and you filter on `> 0` - but a pointer can be left behind by a deletion outside the trigger (`isEInvoice()` false, an invoice moved to another entity, a hand-made delete), and that is precisely a case where the vendor reference still resolves.

```suggestion
		$supplierInvoice = new FactureFournisseur($this->db);
		if ($supplierInvoice->fetch($supplierInvoiceId) <= 0) {
			// The flow pointed at an invoice that cannot be read (deleted outside the trigger, moved to
			// another entity): the vendor reference may still resolve it, so try it before giving up.
			$supplierInvoiceId = ($vendorReference !== '' ? $this->findSupplierInvoiceByVendorReference($vendorReference, $vendorLegalId) : 0);
			if ($supplierInvoiceId <= 0 || $supplierInvoice->fetch($supplierInvoiceId) <= 0) {
				return array('res' => 0, 'message' => "FlowId " . $flowId . " - Failed to load supplier invoice id " . $supplierInvoiceId);
			}
		}
```

einvoicing/class/providers/EsalinkPDPProvider.class.php
```diff
@@ -1634,6 +1634,22 @@ public function syncFlow($flowId, $call_id = null)
...
+				// An incoming one is a status the VENDOR issues about one of its own invoices - "Cashed in"
+				// (212) above all, the answer to the payment we reported with a 211. We never sent it, so it
+				// has no row in einvoicing_lifecycle_msg and the flowId lookup below cannot resolve it.
+				if ($document->flow_direction == 'In') {
+					$resIncoming = $this->processIncomingSupplierInvoiceStatus($flowId, $document, $einvoicing);
+
+					if ($resIncoming['res'] < 0) {
+						// Left unrecorded on purpose: a stored flow is treated as known and never retried.
+						return $resIncoming;
+					}
```
daGrumpf-bxp: This block is right, and it stays as it is - provided the helper marks the failure. Today it does not, and the price of the early return is paid by the whole batch: `syncFlows()` falls past the `postponeflow` branch (line 1020), lands in the generic error path, `$error++`, and line 1127 prints `"Aborting synchronization due to errors."` and breaks the loop over the flows. Nothing behind this flow in the run is processed, and the next run stops in the same place since nothing was stored.

With `postponeflow` set in `processIncomingSupplierInvoiceStatus()` (suggestion on `AbstractPDPProvider.class.php`), the loop `continue`s, counts the flow in `TotalPostponedSync`, and tells the user it will be retried - which is what the comment here promises.

einvoicing/class/providers/SuperPDPProvider.class.php
```diff
@@ -2640,6 +2640,13 @@ public function syncFlow($flowId, $call_id = null)
 				if ($document->flow_direction == 'In') {
 					$resIncoming = $this->processIncomingSupplierInvoiceStatus($flowId, $document, $einvoicing);
 
+					// A negative result is a transient failure (platform call or CDAR parsing): return
+					// without storing the flow so this flowId is retried on the next sync run instead of
+					// being marked processed and losing the vendor status for good.
+					if ($resIncoming['res'] < 0) {
+						return $resIncoming;
+					}
```
daGrumpf-bxp: Same as on `EsalinkPDPProvider`: correct once the helper returns `postponeflow` with the negative result, otherwise this `return` aborts the whole synchronization at line 2085 instead of only this flow. Good that you applied it here too - the two `syncFlow()` implementations diverging is what produced #857 in the first place.

The comment above will need a word too, once the parse failure returns `0`: the negative result is then the platform call only.

daGrumpf-bxp (changes requested):
Thanks, that is a much better shape - the exact match on the flow, the `Converted` fallback and the docblock now saying what the return values mean. I re-read the whole path on `c9783303`; CI is green on the 8 checks.

One blocker, and it is the very change I asked for: **returning early is only half of the protocol**. `syncFlows()` treats a `res < 0` that does not carry `postponeflow` as a hard error - `$error++`, then `"Aborting synchronization due to errors."` and `break` (EsalinkPDPProvider.class.php:1127, SuperPDPProvider.class.php:2085). So the flow no longer condemns itself, it condemns **the whole batch**: every flow behind it in the run is left unprocessed, and since nothing was stored it comes back at the head of the next run and aborts that one too. The `postponeflow` key exists for exactly this ("marks a failure that stored nothing, so the batch may go on and the flow be retried later", syncFlow docblock line 1221 / 2207). Inline suggestions below, in the helper rather than in each provider, so the two `syncFlow()` cases keep the shape you gave them.

Second point, same area: of the two negative returns, only one is transient. A CDAR that does not parse never will, so `-1` there means a flow postponed on every run, for ever - your own new docblock says a negative result is reserved for a transient failure.

The rest checks out, including two things worth writing down since they are the kind that bite later:

- `fetchFlowData($flowId, 'Original')` is equivalent to the `callApi()` it replaces down to the `Accept: application/octet-stream` header and the empty `$callType` (so still not logged as an API call), and it is the same method the `SupplierInvoice` import already uses. No hidden change.
- an incoming LC flow is now stored with `fk_element_type = 'invoice_supplier'` and `fk_element_id > 0`, so it lands in the same table rows as the received invoice itself. It does not create a false duplicate: `isEInvoice()` and `getXmlData()` both filter on `flow_type = 'SupplierInvoice'` (SupplierInvoiceHelper.class.php:345 and :404), and the deletion trigger sets `fk_element_id = 0` rather than deleting, which your `fk_element_id > 0` filter already handles.

Nit, no change needed: `$flowTrackingId` is never `''`. Both providers set `tracking_idref` to `$flowData['trackingId'] ?? (getDolGlobalString('EINVOICING_PDP', 'REF').' '.$flowId)` (EsalinkPDPProvider.class.php:1264, SuperPDPProvider.class.php:2250), so a flow with no `trackingId` carries `"Esalink <flowId>"`. Harmless - it cannot match a `flow_id` - but the guard is dead code and the docblock describes a case that does not happen.

daGrumpf-bxp:
Follow-up on the same point, since you asked: **`"CustomerInvoiceLC"` has exactly the same defect, and it is already in `main`** - this PR does not introduce it, but it is the same convention, so it is worth settling once.

Every negative return of that case block happens *before* `$document->create($user)` (EsalinkPDPProvider.class.php:1795, the single store of the whole method), so none of them stores anything - and none of them carries `postponeflow`. Which means each one aborts the batch at line 1127 / 2085, and then keeps aborting it: `getLastSyncDate()` is `MAX(updatedat)` over the *stored* documents (AbstractPDPProvider.class.php:848-851), so a flow that is never stored never leaves the synchronization window. The stall is permanent, and everything behind that flow in the list stops with it.

The live returns, on `c978330` (the three at 1459-1465 / 2457-2463 are inside the commented-out block):

| Esalink | SuperPDP | Failure | Nature | Suggested |
|---|---|---|---|---|
| 1481 | 2481 | Neither `Original` nor `Converted` retrievable | transient | `-1` + `postponeflow` |
| 1497 | 2497 | CDAR does not parse | **permanent** | `0`, stored and skipped |
| 1509 | 2509 | `Facture::fetch()` returns `< 0` | transient (SQL error only - a ref that matches nothing returns `0`, and that case is already handled below by storing with `fk_element_id = 0`) | `-1` + `postponeflow` |
| 1514 | 2514 | Invoice found in another entity | **permanent** | `0`, same treatment as the `(NOTFOUND)` case right below |
| 1621 | 2621 | Any exception while processing the CDAR | transient (the inner block rolls back and rethrows, so nothing is committed when this is reached) | `-1` + `postponeflow` |

Two more things in the same block, both of which your new shared helper already gets right - which is the argument of this PR made twice over:

1. **A CDAR with no `ReferenceReferencedDocument` stalls the sync on a misleading message.** The guard at 1496 / 2496 only tests `isset($cdarDocument['AcknowledgementDocument'])`, then line 1504 / 2504 indexes `$cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument']['IssuerAssignedID']` unguarded: undefined-key warnings on PHP 8, `$issuerAssignedID` null, and `Facture::fetch(0, null)` returns `-1` because of its own `if (empty($rowid) && empty($ref) && empty($ref_ext)) return -1;` (facture.class.php:2152 on 18/19/20/21, :2239 on 24 and develop - unchanged across all of them). So the sync aborts for ever on `"Failed to fetch customer invoice using CDAR IssuerAssignedID/ref: "` with an empty ref, which points the reader at the wrong thing entirely. `processIncomingSupplierInvoiceStatus()` tests `empty($cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument'])` - the same guard here would turn a permanent stall into one skipped flow.

2. **`SuperPDPProvider` still has no `Converted` fallback** (2468-2481: a third inline copy of the `flows/{id}?docType=Original` call, with the `Accept: application/octet-stream` header written out by hand). Esalink got the fallback at 1473-1481, this PR is adding it a second time in the shared helper, and SuperPDP's customer path is left as the only one that fails the whole synchronization when the access point holds no `Original`. `fetchFlowData($flowId, 'Original')` then `'Converted'` is the same three lines you already wrote.

For completeness, the `create()` failure itself (1797 / 2797) is also a bare `-1` with nothing stored - genuinely transient, so `postponeflow` fits there too.

None of this is a blocker for #858 and I am not asking you to widen its scope: fixing `"SupplierInvoiceLC"` alone is a complete fix for #857. But the `postponeflow` convention should be decided once for both cases rather than twice, and per the rule set on #718 it is meant for exactly this - a failure that stored nothing. Happy to open the follow-up issue, or to take the `"CustomerInvoiceLC"` half myself once this one is merged, whichever you prefer.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,251 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+
+		$cdarHandler = new CdarHandler($db);
+		$cdarDocument = $cdarHandler->readFromString($flowResponse['response']);
+		if (empty($cdarDocument) || empty($cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument'])) {
+			// Not transient, unlike the HTTP failure above: a CDAR that does not parse never will. Stored
+			// so the synchronization does not read it again, and report it, on every run.
+			dol_syslog(__METHOD__ . " FlowId " . $flowId . " carries no readable CDAR", LOG_WARNING);
+			return array('res' => 0, 'message' => "FlowId: " . $flowId . " - Failed to parse CDAR document");
+		}
```
daGrumpf-bxp, line 1104: Both halves of the guard, in one place: catch what `readFromString()` actually does on a body it cannot read (it throws, it never returns empty), and test the lifecycle code instead of the array that is never empty. With this, the four non-XML bodies of the table stop the flow instead of the batch, and a well-formed document that is not a CDAR no longer reaches `insertOrUpdateExtLink()` with an empty status.

```suggestion
		$cdarHandler = new CdarHandler($db);
		try {
			$cdarDocument = $cdarHandler->readFromString($flowResponse['response']);
		} catch (Exception $e) {
			// readFromString() throws as soon as the body is not well-formed XML - an empty body, a JSON
			// error, an HTML page: what an access point answers when it has nothing to serve for a flow.
			dol_syslog(__METHOD__ . " FlowId " . $flowId . " - " . $e->getMessage(), LOG_WARNING);
			$cdarDocument = array();
		}
		// Every value of a parsed CDAR may be empty, but the array itself never is (parseReferencedDocument()
		// always returns its seven keys): test the code this function is here to record, not the array.
		if (empty($cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument']['ProcessConditionCode'])) {
			// Not transient, unlike the HTTP failure above: a CDAR that does not parse never will. Stored
			// so the synchronization does not read it again, and report it, on every run.
			dol_syslog(__METHOD__ . " FlowId " . $flowId . " carries no readable CDAR", LOG_WARNING);
			return array('res' => 0, 'message' => "FlowId: " . $flowId . " - Failed to parse CDAR document");
		}
```

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,251 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+			// Nothing is stored for this flow, so it must come back: 'postponeflow' is what tells
+			// syncFlows() to carry on with the flows behind it instead of aborting the whole batch.
+			return array(
+				'res' => -1,
+				'postponeflow' => 1,
+				'message' => "Failed to retrieve flow details for flowId: " . $flowId,
+				'actioncode' => 'CANT_READ_INCOMING_LIFECYCLE_STATUS',
+				'actionurl' => '',
+				'action' => '',
+				'businessmessage' => $langs->trans('CantReadTheStatusSentByTheVendor', $flowId)
```
daGrumpf-bxp: `action` is empty, and `syncFlows()` writes this entry into `$actions` unconditionally (EsalinkPDPProvider.class.php:1024-1032). `document_list.php:1080-1090` then prints the **Suggested actions** title and, under it, `$tmpactionstodo['action']` - nothing. The user gets an empty section. The other postponed case never hits it because it always fills `action` (`SetTheAccessPointConversionFormat`). Either give it a sentence of its own - even "nothing to do, the flow is retried on the next synchronization" is an answer - or one pointing at the access point, since a 404 on both `Original` and `Converted` is something to check over there.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,251 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		}
+
+		$supplierInvoice = new FactureFournisseur($this->db);
+		if ($supplierInvoice->fetch($supplierInvoiceId) <= 0) {
+			// The flow pointed at an invoice that cannot be read (deleted outside the trigger, moved to
+			// another entity): the vendor reference may still resolve it, so try it before giving up.
+			$supplierInvoiceId = ($vendorReference !== '' ? $this->findSupplierInvoiceByVendorReference($vendorReference, $vendorLegalId) : 0);
+			if ($supplierInvoiceId <= 0 || $supplierInvoice->fetch($supplierInvoiceId) <= 0) {
+				return array('res' => 0, 'message' => "FlowId " . $flowId . " - Failed to load supplier invoice id " . $supplierInvoiceId);
+			}
```
daGrumpf-bxp: The comment does not match the core: `FactureFournisseur::fetch()` filters on the entity only when it searches by ref - by rowid it does not (`fourn/class/fournisseur.facture.class.php`, `if ($id) { WHERE t.rowid = ... } else { WHERE t.entity IN (...) }`, identical on 18.0, 24.0 and develop, with the core's own comment `Don't use entity if you use rowid`). An invoice moved to another entity is fetched fine, so this fallback only covers a row deleted without the trigger - worth keeping, but say that.

Also, line 1172 now prints the id the fallback overwrote (`0` when there is no vendor reference), so the id the exact flow match had found is lost from the message. Keeping it in a separate variable makes that log usable again.

einvoicing/langs/fr_FR/einvoicing.lang
```diff
@@ -368,6 +368,7 @@ PeppolTestAnnuary=Annuaire Test Peppol
 CantFindThirdpartyFromTheImportedInvoice=Le tiers fournisseur de la facture importée (%s) est introuvable dans votre base de données. La création automatique des tiers étant désactivée (ce qui est conseillé), vous devez le créer manuellement. Consultez les actions recommandées…
 CantFindProductFromTheImportedInvoice=Un produit figurant sur la facture importée (%s) est introuvable dans votre base de données. La création automatique de produits est désactivée (ce qui est conseillé), vous disposez donc de plusieurs options pour résoudre ce problème manuellement. Consultez les actions recommandées…
 CantReadTheDocumentOfTheImportedInvoice=Impossible de lire la facture du flux %s : aucun des documents associés détenus par le point d'accès ne respecte la syntaxe requise. Le flux n'a pas été importé et sera réessayé lors de la prochaine synchronisation. Voir la ligne suivante : Action recommandée…
+CantReadTheStatusSentByTheVendor=Le statut envoyé par le fournisseur dans le flux %s n'a pas pu être lu. Rien n'a été enregistré, le flux sera réessayé lors de la prochaine synchronisation.
```
daGrumpf-bxp: `fr_FR` should not be in the PR: eldy asked for `en_US` only, since the other languages are overwritten by the Transifex sync (they come back through the `Trans` commits). The `en_US` line you added is all that is needed here.

daGrumpf-bxp (changes requested):
I re-read the whole path on `677b25c9` and this time I ran it: on the bench (Dolibarr 24.0.1), with the PR's own two class files, the two HTTP calls stubbed, and `syncFlow()` wrapped in exactly the `try/catch` that `syncFlows()` puts around it. CI is green on the 8 checks.

**The postpone works — the blocker of the last round is closed.** With the access point unable to serve the document, `fetchFlowData()` is tried on `Original` then `Converted`, `syncFlow()` returns `res=-1` with `postponeflow=1`, and nothing is written to `llx_einvoicing_document`:

```
[stub] fetchFlowData(docType=Original) -> 404
[stub] fetchFlowData(docType=Converted) -> 404
syncFlow res=-1 postponeflow=1 actioncode=CANT_READ_INCOMING_LIFECYCLE_STATUS
flow stored: no -> retried next run
```

**The other half of the commit does not.** `An unparseable CDAR is not transient (...): stored as res=0 instead of a negative result` is what the commit message promises, and that branch is unreachable, because `CdarHandler::readFromString()` does not return an empty value on a body it cannot read - it **throws** (`CdarHandler.class.php:176-178`), and it is the only way out. Same five bodies, an access point answering 200 to `docType=Original`:

| body of the 200 | what happens |
|---|---|
| empty | `Exception: Error parsing XML string`, uncaught, out of `syncFlow()` |
| `Flow document not available` | idem |
| `<html><body>502 Bad Gateway</body>` | idem |
| `{"error":"no document for this flow"}` | idem |
| well-formed XML that is not a CDAR | `res=0`, flow stored, never retried |

And an exception out of `syncFlow()` is the case the last round was about, one notch worse: `syncFlows()` catches it, `$error++`, then `Aborting synchronization due to errors.` and `break` (EsalinkPDPProvider.class.php:1122-1134, SuperPDPProvider.class.php:2064-2076 - read, not run). Nothing is stored, so the same flow is at the head of the next run and aborts it too: one such flow and the synchronization stops for good. Suggestion inline - it is the `try/catch` `SuperPDPProvider` already has around the same call in its `CustomerInvoiceLC` case (`SuperPDPProvider.class.php:2489`).

The last row of the table is the second half of the same weak spot: `empty($cdarDocument['AcknowledgementDocument']['ReferenceReferencedDocument'])` is **always false**, whatever the XML. `parseReferencedDocument()` returns a fixed array of seven keys (`CdarHandler.class.php:1031-1043`), so the array is never empty - only its values are:

```
readFromString('<ApplicationResponse .../>') :
  empty($cdarDocument)                    = false
  empty(ReferenceReferencedDocument)      = false
  ProcessConditionCode                    = ''
  IssuerAssignedID                        = ''
```

My probe stopped at `Vendor lifecycle status with no invoice reference` only because its trackingId matched no flow of mine. On the case this PR is built for - Hubtimize, where the incoming trackingId *is* one of our flow ids - `findSupplierInvoiceByFlowId()` resolves the invoice, and the code carries on to `insertOrUpdateExtLink()` and `storeStatusMessage()` with `cdar_lifecycle_code = ''`: an empty status written onto a real supplier invoice, from a document that is not a CDAR. The guard has to test the code it is about to record, not the shape of the array.

Three smaller points inline (the empty `Suggested actions` block, the `fr_FR` file, and a comment that does not match what the core does).

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		// an exact link to try below, before it is overwritten with the CDAR's IssuerAssignedID.
+		$flowTrackingId = (string) $document->tracking_idref;
+
+		$flowResponse = $this->fetchFlowData($flowId, 'Original');
+		if ($flowResponse['status_code'] != 200) {
+			// Some flows have no 'Original' on the access point, only the converted copy. Both carry
+			// the same CDAR, so fall back rather than fail the flow (and every flow behind it).
+			$flowResponse = $this->fetchFlowData($flowId, 'Converted');
+		}
+		if ($flowResponse['status_code'] != 200) {
```
daGrumpf-bxp, line 1096: The contract this return announces - nothing is stored, so the flow must come back - does not survive the synchronization cursor.

`getLastSyncDate()` is `MAX(updatedat)` over the stored `einvoicing_document` rows (line 848 of this file), `Document::cronSyncFlows()` passes no margin (`document.class.php:1592`), and `syncFlows()` sorts the batch by `updatedAt` ascending (`EsalinkPDPProvider.class.php:959`) before `continue`-ing past a postponed flow. So any flow *behind* the postponed one that does get stored pushes the cursor past it, and the next run's `updatedAfter` never lists it again.

Concretely: the access point 500s on flow A (`updatedAt` T1), flow B (T2 > T1) imports fine in the same batch. A is postponed, B is stored, the next run starts at T2, and the vendor's 212 is gone for good - the exact outcome this PR exists to prevent, only reported once instead of on every run. The manual sync of `document_list.php` can still be re-run with a hand-picked `syncfromdate`; the cron cannot.

Not introduced here - `CONVERSION_FORMAT_NOT_SUPPORTED` has the same hole - but this PR is the first to make an incoming *status* depend on it, so it is worth closing now. `getLastSyncDate()` already takes `$marginHours` and nobody ever passes one; a margin on the cron path is enough, since the flows a margin re-lists are discarded by the `alreadyProcessedFlowIds` pre-check (`EsalinkPDPProvider.class.php:1003`), which queries only the flowIds of the current listing.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+			return 0;
+		}
+
+		$candidates = array();
+		while ($obj = $db->fetch_object($resql)) {
+			$candidates[] = $obj;
+		}
+		$db->free($resql);
+
+		if (count($candidates) == 1) {
```
daGrumpf-bxp, line 1220: `count($candidates) == 1` returns before `$vendorLegalId` is ever looked at, so the identifier the CDAR carries is only ever a tie-break. The docblock says the reference is "only trusted alone when it matches exactly one invoice" - but one match *in this Dolibarr* is not one match *for this vendor*.

Vendor A reports a 212 on its invoice `2026-001`, an invoice this instance never received; you hold exactly one supplier invoice with `ref_supplier = '2026-001'`, from vendor B - generic numbering is the rule, not the exception. A's "Cashed in" is then written onto B's invoice: wrong `syncstatus` in `einvoicing_extlinks`, a wrong lifecycle row, and a STATUS event on the wrong third party. And it is precisely the case this path exists for: the exact flow match missed because the referenced invoice is unknown here.

Pre-existing on SuperPDP, but this PR puts it on Esalink too, so it is worth a check now rather than twice later. The three comparisons of the loop below want to be a small helper (`$vendorLegalId` against `siren`, `siret`, `tva_intra`) called from both places - see my other remarks on how those three compare - and this branch becomes `count($candidates) == 1 && ($vendorLegalId === '' || helper(...))`, falling through to `0` when the single candidate is somebody else's.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+			$db->commit();
+		} catch (Exception $e) {
+			$exceptionmessage = $e->getMessage();
+
+			$db->rollback();
+		}
+
+		if ($exceptionmessage) {
+			throw new Exception($exceptionmessage);
+		}
```
daGrumpf-bxp: This `try` / `catch (Exception)` / `rollback` cannot fire, and the two calls it wraps report a failure by return value only.

`EInvoicing::insertOrUpdateExtLink()` returns `-1` on a SQL error (`einvoicing.class.php:2902`), `EInvoicing::storeStatusMessage()` returns `-1` (`einvoicing.class.php:3383`); neither throws, and neither return value is read here. So an `INSERT INTO llx_einvoicing_lifecycle_msg` that fails - DB error, an `lc_status_message` longer than the column - falls straight through to `$db->commit()`, the method returns `res=1` "Vendor status 212 recorded on supplier invoice ...", `syncFlow()` books the flow as processed, and the status is lost while the run reports a success. That is the silent loss of #857 again, one layer down.

Reading the two results is also what makes the rollback reachable at all:

```suggestion
		$db->begin();

		// Neither write throws: a SQL failure comes back as -1, so the rollback is only reachable
		// if the two results are read.
		$resExtLink = $einvoicing->insertOrUpdateExtLink($supplierInvoice->id, $supplierInvoice->element, '', $document->cdar_lifecycle_code, '', $statusComment);

		$resStatusMessage = ($resExtLink > 0 ? $einvoicing->storeStatusMessage(
			$supplierInvoice->id,
			$supplierInvoice->element,
			$document->cdar_lifecycle_code,
			$statusComment,
			$document->flow_direction,
			$flowId,
			$document->ack_status,
			$document->ack_info,
			$document->submittedat,
			$document->cdar_reason_code
		) : -1);

		if ($resExtLink <= 0 || $resStatusMessage <= 0) {
			$db->rollback();
			dol_syslog(__METHOD__ . " FlowId " . $flowId . " - failed to record the vendor status: " . $db->lasterror(), LOG_ERR);
			return array('res' => 0, 'message' => "FlowId " . $flowId . " - Failed to record the vendor status on supplier invoice " . $supplierInvoice->ref);
		}

		$db->commit();
```

`res=0` is the conservative reading (the run says so and moves on); the postpone return is defensible too if you consider a DB error worth retrying. Either way, not a `res=1` that claims the status was recorded.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+				'businessmessage' => $langs->trans('CantReadTheStatusSentByTheVendor', $flowId)
+			);
+		}
+
+		$cdarHandler = new CdarHandler($db);
+		try {
+			$cdarDocument = $cdarHandler->readFromString($flowResponse['response']);
+		} catch (Exception $e) {
+			// readFromString() throws as soon as the body is not well-formed XML - an empty body, a JSON
+			// error, an HTML page: what an access point answers when it has nothing to serve for a flow.
```
daGrumpf-bxp: These two comments disagree about the same event. This one says an empty body, a JSON error or an HTML page is "what an access point answers when it has nothing to serve for a flow" - a state that changes on the next run - and the one four lines below concludes "not transient ... a CDAR that does not parse never will", which sends every one of them to `res=0`: stored, never read again.

Both are true, of different bodies. A malformed CDAR will not parse any better next week; a 200 with an empty body on a flow whose document is not materialised yet will. As written the second reading wins for all four cases, and an empty body burns the vendor's 212 for good with a single `LOG_WARNING` for a trace.

Either narrow this comment to the malformed-XML case it really covers, or split the two - `$flowResponse['response'] === ''` deserves the postpone path above, everything else stays at `res=0`.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		if (empty($candidates) || $vendorLegalId === '') {
+			return 0;
+		}
+
+		// Several invoices carry that number: only the one whose vendor is the issuer of the status.
+		$matches = array();
+		foreach ($candidates as $candidate) {
+			if ($vendorLegalId === (string) $candidate->siren
+				|| $vendorLegalId === (string) $candidate->siret
+				|| $vendorLegalId === (string) $candidate->tva_intra) {
```
daGrumpf-bxp: `tva_intra` is compared raw here, where the module normalises it everywhere else: `CommonProtocol.class.php:546` matches on `REPLACE(tva_intra, ' ', '')` against `removeAllSpaces()`. A third party keyed in as `FR 12 345678901` against a CDAR `GlobalID` of `FR12345678901` fails the tie-break, `count($matches) != 1`, and the status is dropped with nothing but a `LOG_WARNING`. `removeAllSpaces()` on both sides costs nothing.

Same for `siret`: a CDAR carrying a SIREN under scheme `0002` - which is what the Hubtimize 212 of #857 carries, `908544638` - never matches a third party whose SIREN field is empty and whose SIRET is filled, where comparing against `substr($candidate->siret, 0, 9)` would. Both belong in the shared helper suggested above, so the single-candidate check and this loop cannot answer differently.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		global $db, $langs;
+
+		require_once DOL_DOCUMENT_ROOT . '/fourn/class/fournisseur.facture.class.php';
+		dol_include_once('einvoicing/class/utils/CdarHandler.class.php');
+
+		$langs->load('einvoicing@einvoicing');
+
+		// On some platforms (verified on Hubtimize) an incoming flow's own trackingId is the flowId of
+		// the outgoing SupplierInvoiceLC flow it acknowledges, already booked in llx_einvoicing_document -
+		// an exact link to try below, before it is overwritten with the CDAR's IssuerAssignedID.
```
daGrumpf-bxp: The `trackingId` of an incoming lifecycle message is not the flowId of an outgoing `SupplierInvoiceLC` we sent - there is none, and that is the whole point of #857. It is the flowId of the flow of the **invoice** the status comments on.

Verified on the Hubtimize case behind the issue: the incoming 212 `8fbe219a-4579-4d11-8ad3-a91c41687243` carries the flowId of the received supplier-invoice flow (`INVOICFR_FA2607-0077.pdf`, received on 07/08), and no `SupplierInvoiceLC Out` flow exists for that invoice at all.

The lookup itself is right either way - both rows carry `fk_element_type = 'invoice_supplier'` - so this is a comment fix, repeated in the `findSupplierInvoiceByFlowId()` docblock (line 1234). Worth doing: read as written, the exact match looks useless for exactly the case it rescues, and the next reader will be tempted to drop it.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,263 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		if ($supplierInvoice->fetch($supplierInvoiceId) <= 0) {
+			// fetch() only scopes by entity when it searches by ref, not by rowid (core: "Don't use entity
+			// if you use rowid"), so an invoice moved to another entity is fetched fine - this only covers
+			// a row deleted without going through the trigger that clears fk_element_id (isEInvoice() false,
+			// a hand-made delete). The vendor reference may still resolve it, so try it before giving up.
+			$fallbackSupplierInvoiceId = ($vendorReference !== '' ? $this->findSupplierInvoiceByVendorReference($vendorReference, $vendorLegalId) : 0);
+			if ($fallbackSupplierInvoiceId <= 0 || $supplierInvoice->fetch($fallbackSupplierInvoiceId) <= 0) {
+				return array('res' => 0, 'message' => "FlowId " . $flowId . " - Failed to load supplier invoice id " . $supplierInvoiceId);
+			}
+			$supplierInvoiceId = $fallbackSupplierInvoiceId;
```
daGrumpf-bxp: Two small things in this block.

In the vendor-reference branch the fallback is a guaranteed no-op: `$supplierInvoiceId` came out of `findSupplierInvoiceByVendorReference($vendorReference, $vendorLegalId)` at line 1168, so calling it again with the same two arguments can only return the same id whose `fetch()` just failed - one wasted pair of queries, then the same `res=0`. Only the exact-flow-match branch can gain anything here, which is what your comment describes. Keeping the id the exact match returned in its own variable and running the fallback only when `$supplierInvoiceId` is that one makes the block do what it says.

And line 1185 is a dead assignment: `$supplierInvoiceId` is never read after it - line 1188 takes `$supplierInvoice->id`.

daGrumpf-bxp (commented):
Re-read at 650a4b58, after the postpone / exact-match round. The shape is right now: the helper marks the transient failure instead of aborting the batch, both `syncFlow()` implementations return rather than break, and the exact flow match is in. What follows is what is left - the first three are not cosmetic.

One point with no line to hang it on, since the file is not in the diff: `doc/function-map.data.json:430` still carries the method as `SuperPDPProvider::processIncomingSupplierInvoiceStatus` and states it "never returns a negative result" (rendered into `FUNCTION-MAP.md:626` and `:680`). Both are false after this PR - it lives on `AbstractPDPProvider` and returns `-1` with `postponeflow` on purpose. `function-map.data.json` is the source, `scripts/build_function_map.php` regenerates the `.md` and the `.html`, and no CI job does it for us.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274: Document::cronSyncFlows() now passes EINVOICING_SYNC_MARGIN_TIME_HOURS to getLastSyncDate(), same margin already used by the manual sync UI. The alreadyProcessedFlowIds pre-check discards the re-listed flows that are already stored, exactly as you described.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274: a single candidate is now checked against vendorLegalId too (rejected if it doesn't match), and the SIREN/SIRET/VAT comparisons are extracted into a shared matchesVendorLegalId() helper used by both branches.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274, applying your suggestion as-is: both write results are now read, rollback on either failure, res=0 returned instead of a false res=1.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274: split into two cases - an empty response body is now postponed (via the shared postponeIncomingSupplierInvoiceStatus()), a malformed/non-CDAR body stays at res=0.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274 as part of the matchesVendorLegalId() helper: VAT compared via removeAllSpaces() both sides, and a SIREN GlobalID also tried against the SIRET's first 9 digits.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274: corrected in both places (the comment here and the findSupplierInvoiceByFlowId() docblock) - it's the flowId of the received supplier invoice flow itself, not an outgoing SupplierInvoiceLC.

einvoicing/class/providers/AbstractPDPProvider.class.php
comaiteseb: Fixed in 64840274: the retry now only fires when the id came from the exact flow match (kept in $flowMatchedSupplierInvoiceId), and the dead assignment is removed.

einvoicing/class/document.class.php
```diff
@@ -1589,7 +1589,12 @@ public function cronSyncFlows()
 		}
 
 		if (isset($provider)) {
-			$syncFromDate = $provider->getLastSyncDate();
+			// A flow postponed on one run (nothing stored for it) is only re-listed by a later run if the
+			// cursor still reaches back to it - a margin, applied here since the manual sync of
+			// document_list.php can already be re-run with a hand-picked date, the cron cannot. The flows
+			// it re-lists that are already stored are cheaply discarded by the alreadyProcessedFlowIds
+			// pre-check in syncFlows(), which queries only the flowIds of the current listing.
+			$syncFromDate = $provider->getLastSyncDate(getDolGlobalInt('EINVOICING_SYNC_MARGIN_TIME_HOURS'));
```
daGrumpf-bxp: This is the right answer for SuperPDP and it cannot work for Esalink, which is the provider the issue comes from.

`SuperPDPProvider::syncFlows()` pages: `while (true)` at line 1874, `$params['where']['updatedAfter'] = $cursor` at line 1883, `MAX_SYNC_BATCHES = 50`. A 12 h window there is simply more batches.

`EsalinkPDPProvider::syncFlows()` issues one call. `$params` only ever holds `where.updatedAfter` and `limit` (lines 888-926), the response is read once into `$results` (line 954), and the loop at line 1001 walks that array and ends. `$limit` is `EINVOICING_FLOWS_SYNC_CALL_SIZE`, 100 by default.

So on Esalink the margin does not widen the reach, it spends it: every run now re-lists the flows of the last 12 hours, and `alreadyProcessedFlowIds` discards them *after* they have taken their place in the page of 100. An account with 100+ flows per 12 h gets a page made entirely of flows already stored - `syncedFlows=0`, `MAX(updatedat)` unchanged, same page next run, for ever. Not a corner case for an account that receives invoices and their lifecycle messages both ways.

The cheapest fix is the one SuperPDP already applies: when the page comes back full, re-issue the search with `updatedAfter` set to the newest `updatedAt` of the page, until a page comes back short. `updatedAtSortKey()` (line 1205) already gives the comparable key, and `alreadyProcessedFlowIds` keeps the re-listed rows cheap.

If you would rather not touch Esalink's paging in this PR, then the margin has to be conditional on the provider being able to page - a constant that fixes one provider and can deadlock the other is worse than the problem it solves.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,327 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+
+		$candidates = array();
+		while ($obj = $db->fetch_object($resql)) {
+			$candidates[] = $obj;
+		}
+		$db->free($resql);
+
+		if (count($candidates) == 1) {
+			$candidate = $candidates[0];
+			if ($vendorLegalId === '' || $this->matchesVendorLegalId($vendorLegalId, $candidate->siren, $candidate->siret, $candidate->tva_intra)) {
```
daGrumpf-bxp: This is what I asked for last round and the tie-break itself is right. One branch is missing at the end of it.

`matchesVendorLegalId()` answers "does this third party contradict the CDAR", and it returns `false` for both "it names someone else" and "it names nobody". Only the first is a reason to drop the match:

| candidate third party | `GlobalID` `908544638` | today | should be |
|---|---|---|---|
| `siren = 908544638` | matches | recorded | recorded |
| `siren = 123456789` | contradicts | dropped | dropped |
| `siren`, `siret`, `tva_intra` all empty | says nothing | **dropped** | recorded |

The third row is a supplier created by hand, imported from a CSV, or foreign - Dolibarr requires none of the three fields, and `_syncOrCreateThirdpartyFromEInvoiceSeller()` only fills them when the vendor was created from a received invoice that carried them. On `main` that status was recorded; here it is lost for good, since `res=0` stores the flow.

```php
$vendorHasNoLegalId = ((string) $candidate->siren === '' && (string) $candidate->siret === '' && (string) $candidate->tva_intra === '');
if ($vendorLegalId === '' || $vendorHasNoLegalId || $this->matchesVendorLegalId(...)) {
    return (int) $candidate->rowid;
}
```

Worth a `dol_syslog(..., LOG_WARNING)` on the `$vendorHasNoLegalId` path: it is the one case where the reference is trusted alone, and if it ever attaches a status to the wrong invoice that line is what will say so.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,327 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+		$flowTrackingId = (string) $document->tracking_idref;
+
+		$flowResponse = $this->fetchFlowData($flowId, 'Original');
+		if ($flowResponse['status_code'] != 200) {
+			// Some flows have no 'Original' on the access point, only the converted copy. Both carry
+			// the same CDAR, so fall back rather than fail the flow (and every flow behind it).
+			$flowResponse = $this->fetchFlowData($flowId, 'Converted');
+		}
+		if ($flowResponse['status_code'] != 200) {
+			return $this->postponeIncomingSupplierInvoiceStatus($flowId, "Failed to retrieve flow details for flowId: " . $flowId);
```
daGrumpf-bxp: `postponeflow` means "come back next run", and for a 5xx or a timeout that is exactly right. For a 4xx it is a loop with no exit: a flow the access point answers `404` for (purged, or an LC for which it never materialises a document) is never stored, so every run fetches it twice and postpones it again, and it stays at the head of every listing for ever.

`syncFlows()` does not surface it either: the postponed branch `continue`s without touching `$error`, so `$globalres` stays `1` and `cronSyncFlows()` never enters the branch that prints `$sync_result['actions']` (`document.class.php:1607`). Only the `TotalPostponedSync` counter reaches the job output - a number, with no flowId and no action.

`$flowResponse['status_code'] >= 500 || $flowResponse['status_code'] == 0` for the postpone, and `res=0` for a 4xx, bounds it: the transient case still comes back, the permanent one is stored once and stops costing two API calls a run.

This is also the third `postponeflow` case in the module, and eldy's ruling on #718 (06/09) was that postponing stays the exception, extended case by case on a proven blocked mailbox. The platform-failure case has that - the `404`/`Converted` fallback came out of a real flow. The empty-200 case does not, as far as I can tell; if you have seen one, saying so in the PR body would settle it in advance.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,327 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+	 * with only its SIRET filled in still carries the same SIREN (verified on the Hubtimize case behind
+	 * #857: GlobalID '908544638' under scheme 0002).
+	 *
+	 * @param	string	$vendorLegalId	Legal identifier carried by the CDAR, empty when it carries none
+	 * @param	string	$siren			Third party SIREN
+	 * @param	string	$siret			Third party SIRET
+	 * @param	string	$tvaIntra		Third party intra-community VAT number
+	 * @return	bool
+	 */
+	protected function matchesVendorLegalId($vendorLegalId, $siren, $siret, $tvaIntra)
```
daGrumpf-bxp: Nit, since you did exactly what I asked and the argument turns out to be wider than what I wrote.

`tva_intra` is normalised now, `siren` and `siret` are not - and the docblock's own reason ("a third party can be keyed in with spaces where a CDAR never carries any") does not single out the VAT number. Dolibarr validates `idprof1` only for a handful of countries, so `908 544 638` in `siren` is perfectly storable, and it never matches `GlobalID` `908544638`.

The SIRET/SIREN relation is also one-way: line 1320 tries a scheme-`0002` SIREN against the first 9 digits of the SIRET, but a scheme-`0009` SIRET in the CDAR is not tried against a third party that only has the SIREN filled - the other half of the same case.

Both fall out of normalising once at the top:

```php
$vendorLegalId = removeAllSpaces($vendorLegalId);
$siren = removeAllSpaces((string) $siren);
$siret = removeAllSpaces((string) $siret);
```

then comparing, plus `substr($vendorLegalId, 0, 9) === $siren` for the reverse direction.

daGrumpf-bxp (commented):
Re-read at `64840274`. The three points of the last round are closed, and closed properly: the guard now tests `ProcessConditionCode` instead of the array that is never empty, `readFromString()` is wrapped, the two writes are read and the `rollback()` is reachable, the empty `Suggested actions` block is filled, and `fr_FR` is out. CI is green on the 8 checks.

Two of the things I asked for have a second half that this commit does not have yet. Both are consequences of the fixes themselves, so this is not a new pile - it is the same two points, one notch further in.

**1. The margin does not reach the flow it was added for, on the provider this PR is about.**

`cronSyncFlows()` now passes `EINVOICING_SYNC_MARGIN_TIME_HOURS` (12 by default, `modEInvoicing.class.php:180`), which answers the cursor problem on SuperPDP: `SuperPDPProvider::syncFlows()` walks `updatedAfter` in batches (`while (true)` at line 1874, `MAX_SYNC_BATCHES = 50`), so a wider window only means more batches.

`EsalinkPDPProvider::syncFlows()` does not page. It builds `$params['limit'] = $limit` once (line 926) and issues a single `flows/search`; there is no offset, no cursor, no loop. So the margin does not add batches there, it eats the only one:

| flows updated in the last 12 h | page of 100 | result |
|---|---|---|
| 20 | 20 old + up to 80 new | fine, 20 slots burned per run |
| 100+ | 100 old | `alreadyExist=100`, `syncedFlows=0`, `MAX(updatedat)` unchanged - the next run fetches the same page |

The second row is a synchronization that stops for good, on the very account this PR was written against. Before the margin the window started at `MAX(updatedat)` and the page was almost entirely new flows, so it could not happen.

The fix is the one SuperPDP already has: when the page comes back full, re-issue the search with `updatedAfter` moved to the newest `updatedAt` of the page, until it comes back short. `updatedAtSortKey()` is already there for the comparison.

**2. The legal-id check I asked for now discards a match it has nothing against.**

You applied it to the single-candidate case, which is what I asked, and the reasoning holds: one match in this Dolibarr is not one match for this vendor. But a third party with `siren`, `siret` and `tva_intra` all empty disproves nothing, and it is now treated as a mismatch - `return 0`, `res=0`, flow stored, the 212 lost with no retry. That case used to be recorded. Detail inline.

Still open from the last round, with no line to hang it on since the file is not in the diff: `doc/function-map.data.json:430` still carries the method as `SuperPDPProvider::processIncomingSupplierInvoiceStatus` and still says it "never returns a negative result" (rendered into `FUNCTION-MAP.md:626` and `:680`). Both are false since `677b25c9`. `scripts/build_function_map.php` regenerates the `.md` and the `.html`; no CI job does it for us.

Two smaller ones inline (the 4xx that is postponed for ever, and `siren`/`siret` left un-normalised where `tva_intra` now is).

Last thing, unrelated to the code: the branch is **16 commits behind `main`**, and `AbstractPDPProvider` is the file the open PRs collide on most. Worth rebasing before this goes in.

einvoicing/class/providers/AbstractPDPProvider.class.php
```diff
@@ -1067,6 +1067,327 @@ public function addEvent($eventType, $eventLabel, $eventMesg, $object)
...
+			return 0;
+		}
+		if (empty($candidates) || $vendorLegalId === '') {
+			return 0;
+		}
+
+		// Several invoices carry that number: only the one whose vendor is the issuer of the status.
+		$matches = array();
+		foreach ($candidates as $candidate) {
+			if ($this->matchesVendorLegalId($vendorLegalId, $candidate->siren, $candidate->siret, $candidate->tva_intra)) {
```
eldy: I can't find this function in previous version.
The description of PR is that is mutualize code of SuperPDP fo EsaLink.
But it seems it also introduce new rule management with thie method.

This means the PR is mixing refactoring, fixing a bug to add a guard in superpdp to esalink and also fixing or changing something else.
The PRis not atomic so can't be validated. Thanks to split each topic in one different PR.

comaiteseb: Hi Eldy, 
Originally my PR was simple — it was daGrumpf-bxp who led me to make my PR more complex. You (all) need to get on the same page.
If you prefer I revert to my first commit (the one I would like at first)

daGrumpf-bxp: Hi eldy — the function is mine, not new: `findSupplierInvoiceByVendorReference()` came in with #526 (merged 01/08), and lives today in `SuperPDPProvider`, private. This PR only moves it to the abstract class so Esalink can reach it. I diffed the moved bodies: byte-identical.

Where you are right is why I stepped in. That code cannot be moved verbatim, because it is already wrong where it stands. It returns `-1` on two failure paths, against its own docblock (*"never returns a negative result ... the flow is stored either way"*). In `syncFlows()`, a `-1` aborts the whole batch — and `syncFlow()` stores the flow anyway, so the pre-check skips it for ever and the vendor status is lost. That is live in SuperPDP today. Grafted onto Esalink as-is, it also brings back #449 (flows the access point holds only as `Converted`).

So the atomic PR is: the move, plus those two `-1` turned into `0` — two lines, which is what the docblock always promised. Everything else I asked for is a separate subject and I agree it does not belong here: the `Converted` fallback, the postpone convention (your #718 ruling applies), the `trackingId` exact match, the identifier matching — `matchesVendorLegalId()` on the line you flagged is indeed new, it came out of my review. I will open them one by one, on top of main, once this one is in.

eldy: Ok, so can you split the PR to just have the move + the minimal change to have the move ok ?
And the rest into another PR ?

comaiteseb: Done — force-pushed. fix857 is now just the original move (57d659b8) plus the two-line fix daGrumpf described (0fdf7542): the two res=-1 returns in processIncomingSupplierInvoiceStatus() now match what its docblock always promised (res=0, stored and skipped, never aborting the batch). Everything else from the review (Converted fallback, postponeflow convention, trackingId exact match, matchesVendorLegalId, transaction result-checking, sync margin) is out of this PR — kept on my side for whenever daGrumpf opens the follow-ups.

comaiteseb:
Split done as requested: `fix857` now only has the move + the two-line fix (0fdf7542) - the original commit plus turning the two `res=-1` into `res=0`, matching what the docblock already promised.

The rest (postpone convention, sync margin, exact flowId match, `matchesVendorLegalId()`, transaction result-checking, `Converted` fallback) is in a follow-up PR: #877, stacked on top of this one.

Issue #850, opened 2026-09-07
[einvoicing] Total amount on Dolibarr does not reflect e-invoice total amount


hansemschnokeloch:
### Bug

Another issue, ~~not sure if it is a duplicate of https://github.com/Dolibarr/dolibarr-community-modules/issues/844~~

After testing, https://github.com/Dolibarr/dolibarr-community-modules/pull/848 doesn't fix this issue, this is also **not a duplicate** of https://github.com/Dolibarr/dolibarr-community-modules/issues/844

Invoice total should be 300.00 €

In dolibarr amount is 50 005.00 €

xml data of einvoice (personnal data redacted)

```
<?xml version="1.0" encoding="UTF-8"?>
<CrossIndustryInvoice xmlns="urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100">
  <ExchangedDocumentContext xmlns="urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100">
    <BusinessProcessSpecifiedDocumentContextParameter xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
      <ID xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">S1</ID>
    </BusinessProcessSpecifiedDocumentContextParameter>
    <GuidelineSpecifiedDocumentContextParameter xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
      <ID xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">urn:cen.eu:en16931:2017#conformant#urn:factur-x.eu:1p0:extended</ID>
    </GuidelineSpecifiedDocumentContextParameter>
  </ExchangedDocumentContext>
  <ExchangedDocument xmlns="urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100">
    <ID xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">2608</ID>
    <TypeCode xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">380</TypeCode>
    <IssueDateTime xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
[...]

Issue #844, opened 2026-09-07
[einvoicing] Unit price precision limit ?


hansemschnokeloch:
### Bug

Question : is there a precision limit for the unit price in electonic invoices ?

I stumbled on the following case with an OVH invoice

- real unit price is 0,00000548 € (on the order)
- qty is 1951593
- total amount is 10,69 €

In the e-invoice unit price is rounded to 0,000005 € (six digits after comma), and total amount remains 10,69 €, but calculated amount by Dolibarr is 9,76 €, and of course the total of the invoice does not match the total of the e-invoice.

E-Invoice xml

```
    <IncludedSupplyChainTradeLineItem xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
      <AssociatedDocumentLineDocument xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
        <LineID xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">478803531</LineID>
      </AssociatedDocumentLineDocument>
      <SpecifiedTradeProduct xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
        <Name xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">Stockage Standard Infrequent Access - Bucket xxxxx sur la région de</Name>
        <Description xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">xxxxx</Description>
        <DesignatedProductClassification xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
          <ClassCode xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100" listID="ZZZ">PUBLICCLOUD - STORAGE STANDARD IA</ClassCode>
        </DesignatedProductClassification>
      </SpecifiedTradeProduct>
      <SpecifiedLineTradeAgreement xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
        <NetPriceProductTradePrice xmlns="urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100">
[...]

Issue #781, opened 2026-09-04
[einvoicing] A received invoice is booked two cents above the document: the import ignores Dolibarr's VAT calculation mode


daGrumpf-bxp:
### Bug

A received invoice is booked **two cents above the document it stands for**, on every supported Dolibarr version, because the import lets the instance decide how the VAT of the invoice is computed instead of adopting the convention the document was issued with.

Dolibarr has two VAT calculation modes, and the supplier invoice card offers them as a link (`ReCalculate  Mode1 / Mode2`, `fourn/facture/card.php`):

* **mode 1, "total of round"** — round the VAT of each line, then add them up. This is the default (`update_price()` falls back to `'0'` when nothing is set).
* **mode 2, "round of total"** — add up the line amounts, then compute and round the VAT once. Selected by `MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND_SUPPLIER` (Dolibarr 20 and above; on 18 and 19 the core only reads the generic `MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND` for supplier documents too).

A received e-invoice does not leave that open: BT-110 and BT-112 are given by the sender. The import goes through `FactureFournisseur::updateline()` → `update_price(1, 'auto', …)`, so it applies whatever the instance is set to, and on a document whose issuer rounded the other way the recorded invoice no longer totals what the document announces.

**Measured on a real vendor invoice received through a PA** — CII, `urn:cen.eu:en16931:2017`, 46 lines, one VAT rate at 20 %, announced **146.57 / 29.31 / 175.88**. Module `main` (1.2.0, `fe9b2057`), same document, same instance data, only the mode changes:

| Dolibarr | import in mode 1 (default) | import in mode 2 | document |
|---|---|---|---|
| 18.0.10 | 146.57 / 29.33 / **175.90** | 146.57 / 29.31 / **175.88** | 146.57 / 29.31 / 175.88 |
| 19.0.4 | 146.57 / 29.33 / **175.90** | 146.57 / 29.31 / **175.88** | ” |
| 20.0.4 | 146.57 / 29.33 / **175.90** | 146.57 / 29.31 / **175.88** | ” |
| 21.0.4 | 146.57 / 29.33 / **175.90** | 146.57 / 29.31 / **175.88** | ” |
| 22.0.5 | 146.57 / 29.33 / **175.90** | 146.57 / 29.31 / **175.88** | ” |
[...]

Issue #778, opened 2026-09-04
Factur-X (CII) import: BasisQuantity is not taken into account on lines priced "per N units" → line amount multiplied by N


julce100:
### Bug

**Summary**

When importing a supplier invoice in Factur-X format (CrossIndustryInvoice / EN 16931), lines whose price is expressed "per N units" via the ram:BasisQuantity tag (N ≠ 1) are imported as if BasisQuantity were 1. The amount of the resulting Dolibarr invoice line ends up multiplied by N compared to the actual invoice amount.

This case is commonly found on surcharge-type lines expressed as a rate "per 100" (e.g. fuel/energy surcharge), a usage that is standard practice with several transport/logistics suppliers.
**Observed behavior**

The module imports the line with a total of 100.00 € (= ChargeAmount × BilledQuantity, i.e. 2.00 × 50), instead of the expected 1.00 €. The BasisQuantity tag appears to be ignored when computing the unit price / line total, effectively treating it as if it were 1.

On a real invoice containing several such lines, the discrepancy can result in a ×100 factor on the imported invoice's total net and gross amount.

### Module

einvoicing

### Dolibarr Version

20.0.4

### Environment PHP

8.2.30

### Environment Database

12.2.2-MariaDB-ubu2404

### Steps to reproduce the behavior and expected behavior

1) Import, via the e-invoicing module, a Factur-X invoice containing a line whose price is given "per N units", for example:
```
xml
<SpecifiedLineTradeAgreement>
  <NetPriceProductTradePrice>
    <ChargeAmount>2.000000</ChargeAmount>
    <BasisQuantity unitCode="C62">100.0000</BasisQuantity>
  </NetPriceProductTradePrice>
</SpecifiedLineTradeAgreement>
<SpecifiedLineTradeDelivery>
  <BilledQuantity unitCode="C62">50.0000</BilledQuantity>
</SpecifiedLineTradeDelivery>
<SpecifiedLineTradeSettlement>
  <SpecifiedTradeSettlementLineMonetarySummation>
    <LineTotalAmount>1.00</LineTotalAmount>
  </SpecifiedTradeSettlementLineMonetarySummation>
</SpecifiedLineTradeSettlement>

```
Here, the price is 2.00 € per 100 units, billed on a quantity of 50: expected total = (2.00 / 100) × 50 = 1.00 € (confirmed by LineTotalAmount in the source XML).

[...]

PR #863, opened 2026-09-07
NEW #861 [einvoicing] A received document the import cannot reproduce no longer reaches validation

### LINKED ISSUE #861: [einvoicing] A received document whose totals the module cannot reproduce should stop the import, not be validated and paid
### Feature

A received document whose totals the module cannot reproduce is imported anyway, silently. Issues #777, #778, #844 and #850 are all the same story: a line the import rebuilt differently from what the document announces, found only because somebody noticed the amount on screen. The invoice can then be validated, approved (205) and paid at the wrong figure.

The module already knows how to tell: `SupplierInvoiceHelper::checkDolInvoiceAndEInvoiceConsistency()` compares the currency, the total excluding VAT (BT-109), the total including VAT (BT-112), the total VAT (BT-110) and the basis and amount of **each VAT rate**, and it even replays VAT modes 1 and 2 to say which one would fit.

It just never runs when it matters:

- it is called **only** from the `BILL_SUPPLIER_VALIDATE` trigger, never at import;
- and only when `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` is set, which is **off by default** and whose setup entry is itself hidden behind `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION_AVAILABLE`, so an administrator never sees the option exists.

### What is proposed

1. **Check at import.** Run the same comparison at the end of `createSupplierInvoiceFromSource()`, on the invoice just created, and report the verdict in the sync result: the document reference and file, `document total = X` against `Dolibarr total = Y`, rate by rate, and an invitation to open an issue with that file — which stays reachable, since the XML is attached to the invoice and readable from the XML preview.
2. **Refuse to go further with an invoice known to be wrong.** A mismatch found at import is recorded on the invoice, and while it stands the invoice cannot be validated and no acceptance status (205) is offered for it. The check is replayed at validation, so an invoice the operator corrected to match the document validates normally and the mark is cleared: the block is "until the totals agree", not a dead end.
3. **Bring the existing option back into the open.** `EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` becomes a visible setting, still opt-in. It keeps its own meaning — re-check *every* e-invoice at validation, including the ones edited by hand afterwards — which is a different question from "the import itself could not reproduce this document".

Nothing new is computed: the arithmetic is the helper's, already tested and already translated.

### Why it matters

An imported invoice that does not total what the document bills is not a display problem: it goes into the accounts, it is what gets approved to the vendor and paid. Catching it at import, where the document and the file are still in hand, is the difference between a two-line issue report and someone finding the gap in a bank reconciliation.

daGrumpf-bxp:
Fixes #861.

## What was silent

`alignInvoiceTotalsWithDocument()` already confronts the invoice the import built with BT-110 and BT-112, and replays both VAT rounding conventions to find the one the vendor used (#781). When neither lands on the announced totals, it **gave up without a word** and kept the invoice — free to be validated, answered with a 205 and paid at a figure the document does not carry. #777, #778, #844 and #850 were all found the same way: somebody noticed an amount on screen.

## What happens now

The document is named, the two totals are printed, and the invoice is marked:

> The received document T861-… announces a total of 130.00 including VAT (30.00 of VAT), while the invoice built from it totals 120.00 including VAT. Neither VAT calculation mode rebuilds what the document announces, so the import could not reproduce it.
> That invoice cannot be validated, and no approval can be sent for it, until its totals are those of the document. The e-invoice file is attached to the invoice and readable from its XML preview: please report the problem with that file, both totals and your Dolibarr version.

While the mark stands, `validate()` is refused by the trigger and `getSendableStatusesForReceivedInvoice()` offers no acceptance status (205, 206) — refusing the document stays offered, which is the answer such a document deserves. The invoice itself is kept: the file is attached to it, and nothing else carries what the vendor sent.

The mark is not a dead end. It holds the announced totals, and they are confronted again at validation: an invoice corrected to what the vendor bills validates normally and drops its mark. No schema change — the mark is a row in `llx_einvoicing_extrafields`, the module's own property store.

## The option comes back into the open

`EINVOICING_SUPPLIER_INVOICE_CHECK_CONSISTENCY_ON_VALIDATION` is shown in the setup page again, still **off by default**. It was hidden in db421d40 for a stated reason:

> FIXME Disabled. […] Adding a rollback inside a more global transaction break all workflows. On Postgresql, it also cancel any following commits. A check to compare data is a readonly operation and should NEVER open transaction and try to modify data.

That objection no longer has a target: the comparison was rewritten to recompute in memory through `PriceHelper::calculatePriceTotal()`, and the whole path `checkDolInvoiceAndEInvoiceConsistency()` → `getInvoiceDetailsForComparison()` → `calculatePriceTotal()` now contains no `begin()`, no `rollback()` and no write — the one query left is the `SELECT` on `c_tva` the core's own `calcul_price_total()` makes. The stale "seriously bugged, do not use it" header goes with it. Its scope stays its own: re-checking *every* e-invoice at validation, including those edited by hand, which is a policy — hence opt-in.

`totalsAgreeWithDocument()` moves to `SupplierInvoiceHelper`, where the trigger and the status list reach it, rather than being copied.

## Measured

Real imports on **Dolibarr 23.0.4**, read back from the database:

| | document | invoice built | outcome |
|---|---|---|---|
| line 100.00 at 20 %, header announcing 30.00 of VAT | 130.00 | 120.00 | reported, marked, `validate()` → **-1** with the message above |
| same invoice, line corrected to 100.00 at 30 % | 130.00 | 130.00 | `validate()` → **1**, ref SI0001-0056, mark cleared |
| the document of #850, on a branch without its fix | 300.00 | 50 005.00 | reported and blocked |

## Tests

Three cases in `SupplierInvoiceHelperTest` (the mark blocking until the invoice agrees, the absolute-value comparison of a credit note, the acceptance statuses withdrawn) and the two cases of `ImportVatCalculationModeTest` that described the old silence, rewritten for the new one. Module suite file by file on **18.0.10, 19, 20, 21, 22, 23.0.4, 24.0 and 25**, each under the PHP of the CI matrix: **29/29 files on all 8 cores**. QA generation matrix, 9 instances × 14 documents: 126/126. E2E, 16 cycles both ways: all complete (200, 201, 205, 211, 212 — one carrying an extra 202).


daGrumpf-bxp:
About the red "PHPUnit on Dolibarr 18.0.0" on this PR: it is the runner, not the code. The job died in `Initialize containers` — `Failed to initialize container mariadb:11.4`, the MariaDB healthcheck never answered — so not a single test ran. I cannot re-run a job on this repository, so I dispatched the same workflow on the fork, where it covers the seven cores instead of the two a PR runs:

https://github.com/daGrumpf-bxp/dolibarr-community-modules/actions/runs/34146670100

- **18.0.0, 20.0.0, 21.0.0, 22.0.0, 23.0.0, 24.0.0: green.**
- 19.0.0: 28 files pass, `CdarDateFormatTest.php` fails — and it fails the same way on `main` (run 34113396464, same workflow, same core), which is what #846 fixes. Nothing here touches it.

Happy to push a rebase if a re-run of the upstream job is easier than trusting the fork.

PR #865, opened 2026-09-08
NEW: [einvoicing] Add EINVOICING_SEND_SUCCESS and EINVOICING_SEND_ERROR triggers after e-invoice transmission to Access Point

Boushabazakaria:
## NEW: [einvoicing] Add EINVOICING_SEND_SUCCESS and EINVOICING_SEND_ERROR triggers after e-invoice transmission to Access Point

## What does this PR do?

Fire two new triggers after each attempt to send an e-invoice to the Access Point (PDP):

- `EINVOICING_SEND_SUCCESS` — fired on successful transmission, exposes `flow_id` via `$object->context['einvoicing_flowid']`                                                                                                                  
- `EINVOICING_SEND_ERROR`   — fired on failed transmission, exposes errors via `$object->context['einvoicing_errors']`                                                                                                                         

Both triggers are fired from:
- the **auto-send path** (`afterPDFCreation` hook, when `EINVOICING_AUTO_SEND_ON_GENERATION` is enabled)
- the **manual/mass send path** (`sendOneInvoiceToAccessPoint`)
                                                                                                                                                                                                                                                
## Why?

Allow other modules to react to a successful or failed e-invoice transmission (e.g. update a status, notify a third-party system, log to an external service) by listening to these triggers.

### DIFF
--- a/einvoicing/class/actions_einvoicing.class.php
+++ b/einvoicing/class/actions_einvoicing.class.php
@@ -68,7 +68,7 @@ class ActionsEInvoicing extends CommonHookActions  // @phan-suppress-current-lin
 	 */
 	public function afterPDFCreation($parameters, $object, &$action, $hookmanager)
 	{
-		global $db, $langs;
+		global $db, $langs, $user;
 
 		dol_syslog(__METHOD__ . " Hook afterPDFCreation called for object " . get_class($object));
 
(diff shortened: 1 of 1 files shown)


einvoicing/class/actions_einvoicing.class.php
```diff
@@ -214,11 +214,15 @@ public function afterPDFCreation($parameters, $object, &$action, $hookmanager)
 									$sendres = $provider->sendInvoice($invoiceObject);
 									if ($sendres) {
 										setEventMessages($langs->trans("InvoiceSuccessfullySentToPDP") . ' - ' . $langs->trans("FlowId") . ': ' . $sendres, null, 'mesgs');
+										$invoiceObject->context['einvoicing_flowid'] = (string) $sendres;
```
eldy, line 217: Triggers are restricted to CRUD action (Create / Update or Delete an object in database).
When we call sendinInvoice, we also call the logCall function that run the CRUD action Create into table llx_einvoicing_call for object Call (that trackevent).
So the object in trigger will be $call and $call->call_type will be 'sendinvoice'
You may also find if it is a success or failure with into $object->context['statusCall']

PR #866, opened 2026-09-08
FIX [einvoicing] A received Factur-X no longer fatals on PHP below 7.3

daGrumpf-bxp:
Follow-up to #840, from the other end. That pull request raised the module's declared `phpmin`, and it was refused: upgrading an ERP is the user's decision, not ours. This one leaves that decision alone. It removes the fatal error and says what the choice is.

## What happens today

`horstoeko/zugferd` is resolved for PHP 7.3 (`vendor/composer/platform_check.php`), and `FacturXProtocol.class.php` requires its autoloader at file scope. Reception picks the protocol from the document that **arrived**, not from `EINVOICING_PROTOCOL`, so a site set to CII reaches that include as soon as a supplier sends a Factur-X:

```
syncFlows -> syncFlow -> fetchImportableFlowDocument -> ProtocolManager::getProtocol
          -> dol_include_once(FacturXProtocol) -> vendor/composer/platform_check.php:24
```

## What this changes

- `ProtocolManager::getProtocol()` answers `null` for a protocol whose libraries cannot run on this PHP, **before** the include. The greyed "PHP 7.3+" entry of the setup page now reads its minimum from the same place.
- The reception keeps the flow pending, as it already did for a syntax it cannot read, and adds a suggested action: set the conversion format of the Access Point account to CII, or upgrade PHP.
- `Document::cleanXmlData()` and `SupplierInvoiceHelper` dereferenced the protocol without checking it; they now fail the way they already did when no protocol is detected.

## Measured

Dolibarr 18.0.10, receiver on `EINVOICING_PROTOCOL=CII`, real invoices through an access point, five PHP built side by side.

| PHP | before | after, issuer sends CII | after, issuer sends Factur-X |
|---|---|---|---|
| 7.0 / 7.1 / 7.2 | fatal error, exit 255, HTTP 500 | imported (the CII `Original` is read) | flow kept, action shown |
| 7.3 / 7.4 | imported | imported | imported |

The same flow stayed pending under 7.0, 7.1 and 7.2, then was imported by the 7.3 run: nothing is lost. Over HTTP the synchronization page went from a 107-byte HTTP 500 carrying the raw Composer message to a normal page. PHPUnit: 29 files x 8 cores, 232 OK.

The same click on **Synchronize**, Dolibarr 18.0.10 on PHP 7.0, before :

before: HTTP 500

and after:
after: the flow is kept and the action is explained

### DIFF
--- a/einvoicing/class/document.class.php
+++ b/einvoicing/class/document.class.php
@@ -1660,6 +1660,9 @@ class Document extends CommonObject
 			throw new Exception(__METHOD__ . " : protocol not detected for XML data");
 		}
 		$protocol = $protocolManager->getProtocol($detectedProtocolName);
+		if (!is_object($protocol)) {
+			throw new Exception(__METHOD__ . " : protocol " . $detectedProtocolName . " is not available on this server");
+		}
 		$protocolClassName = get_class($protocol);
 
 		$cleanedXmlData = $protocolClassName::removeAttachmentFromXml($xmlData);
--- a/einvoicing/class/protocols/ProtocolManager.class.php
+++ b/einvoicing/class/protocols/ProtocolManager.class.php
@@ -37,6 +37,10 @@ class ProtocolManager
 	const EXCEPTION_UNSUPPORTED_FORMAT = -100;
 	const EXCEPTION_UNKNOWN_FORMAT = -101;
 
+	// Factur-X is read and written by horstoeko/zugferd, whose vendor tree refuses to load below this
+	// version with a fatal error raised by vendor/composer/platform_check.php, not with a message.
+	const PHP_MIN_FACTURX = '7.3';
+
 	/**
 	 * Initialize available protocols.
 	 * @param DoliDB $db db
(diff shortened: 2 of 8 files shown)


Daoud-mohamed:
@daGrumpf-bxp Using https://github.com/nikic/php-Parser can be a better alternative, as it supports PHP 7.0 and can be used to extract the XML file from a Factur-X Invoice.

daGrumpf-bxp:
Superseded by #868, closing this one.

@Daoud-mohamed, your point sent me measuring where the PHP floor actually comes from, and the answer made this branch unnecessary. On the inbound side `horstoeko/zugferd` does exactly one thing: pull the XML attachment out of the PDF/A-3 container — 25 lines of `ZugferdDocumentPdfReaderExt::collectAttachmentsFromPdfContent()`. What comes out is then parsed by the module's own CII parser, the same one a plain CII document goes through. So the whole 7.3 floor was being paid for an extraction, and no third-party library is needed for it: the core already ships a PDF parser, `htdocs/includes/tcpdi/tcpdi_parser.php`, self-contained, present unchanged from Dolibarr 18 to 25.

#868 does that. Measured on the 1885 Factur-X documents the bench holds: extraction byte for byte identical to the library, and identical again when the same 1885 are re-read under **PHP 7.0 on Dolibarr 18.0.10**. A received Factur-X is imported there, where this branch could only postpone the flow and explain why.

So on the receiving side there is nothing left to guard against and no reason to raise `phpmin`: reception loads no composer library at all. The library stays where it writes the container — emission — which is the only place a floor remains, and one the user walks into knowingly.

PR #867, opened 2026-09-08
NEW einvoicing: match vendor product references whatever their writing

euchmax:
Two fixes in `findProductFromEinvoiceLine()`, sharing one root cause: a product reference is compared as a raw string, while each system it travels through writes it its own way. They are independent — happy to split them into two pull requests if you prefer.

**1. Match the vendor reference on a canonical form**

Step 1 compares `pfp.ref_fourn` with `prodsellerid` by strict equality. A vendor that does not write its reference the same way on its order forms and on its invoices defeats it entirely: the reference recorded from the order form reads `A1234-10_42`, the invoice carries `A1234|10|42`. Same item, same segments, other separators — no match, and every line of every invoice of that vendor has to be bound by hand in the mapping screen.

This adds a fallback that only runs when the exact lookup returned nothing, comparing both sides on a canonical form: letters and digits only, upper case.

- The exact lookup stays first, so nothing changes for the vendors that already match, and the index on `ref_fourn` is still used there.
- The normalization is done in PHP, not in SQL. Removing every separator in SQL needs `REGEXP_REPLACE`, which is not available on every supported database; a function applied to the column would defeat any index anyway; and one query per supplier costs less than one scan per invoice line. The map is cached per supplier and entity, for the run.
- Canonicalization is lossy, so it can merge two distinct references. When it does, the line is left to the manual mapping and a warning is logged, rather than bound to an arbitrary row — the current `LIMIT 1` without `ORDER BY` would pick one silently.
- The fallback is deliberately limited to `prodsellerid`. Applying it to `prodname` as well would compare canonicalized labels, where collisions are far more likely than on references.

**2. Symmetric sanitization of the `EI-` lookup**

The same method looks up `'EI-' . $lineData['prodsellerid']`, while `_findOrCreateProductFromEinvoiceLine()` creates the product with `'EI-' . dol_sanitizeFileName($lineData['prodsellerid'])`.

As soon as the vendor reference holds one of the characters `dol_sanitizeFileName()` replaces — `|, <, >, /, \, ?, *, ", :, °, $, ;, \` \` — the two strings differ and the lookup never matches. A reference such as `A1234|10|42` creates the product `EI-A1234_10_42` and is then searched as `EI-A1234|10|42`, so the step is silently dead for that vendor: every invoice either creates a duplicate product or falls back to the manual mapping.

Fix: apply the same transform on the lookup side, so both sides build the same string.

The comment above that lookup also had a typo, `inmported`, corrected in passing.


daGrumpf-bxp:
Point 2 is a plain bug and welcome. Point 1 is another matter: you say it yourself, canonicalization is lossy — it can merge two distinct references, and that guesswork would then apply to every install by default.

If it goes in, it has to be an opt-in setup option, off by default. Silently matching products on an approximation is exactly the kind of behaviour I do not want running on my own production data.

euchmax:
Agreed on both counts :) !

Point 1 is now behind a setup option, `EINVOICING_PRODUCTS_MATCH_CANONICAL_REF`, off by default, next to the other import options and flagged with `warningifon` so the setup page warns when it is turned on. It is not declared in the module constants either, so an existing install stays on the current behaviour after an upgrade.

Two properties of the fallback are worth keeping in mind, since they are what makes the approximation bounded:

- it never overrules an exact match — the strict lookup runs first and returns as before, so an install with the option on behaves identically on every line that already matched;
- when several products of the supplier share the same canonical form, nothing is bound. The line goes to the manual mapping and a warning is logged, so the ambiguity is surfaced rather than resolved by chance.

It is not silent either: the mapping screen already shows the resolution message against each line, so a line matched on the canonical form reads "Product found by prodsellerid (canonical form)" and can be told apart from an exact match at a glance.

Point 2 is unchanged.

Issue #563, opened 2026-08-08
[einvoicing] New invoice tab: e-invoicing lifecycle tracking (status per actor + chronogram)


Charlymd:
### Feature Request

Add a dedicated tab on the customer invoice card ("E-invoicing tracking") to follow the e-invoicing lifecycle of *that specific invoice*, instead of relying only on the generic agenda filtered by label.

The tab shows, using data already stored in `llx_einvoicing_lifecycle_msg` (currently write-only: only the last message is read, by `fetchLastknownInvoiceStatus()`):

1. **Current status per actor** — 3 cards (Fournisseur / PDP-PA / Client) with the last known event for each, derived from `direction` (in/out) + the XP Z12-012 status code range. Before the first PDP exchange, the Fournisseur card falls back to the local Dolibarr-side status (not generated / generated / generation error).
2. **Chronogram** — an inline SVG sequence diagram (3 swimlanes, one row per event, arrows between actors, dashed/red for rejections, a bracket grouping events sharing the same timestamp with the transaction/flow_id in a tooltip).
3. **Detailed log** — a plain table with the full provider message, validation status and reason code for audit/debugging.

Screenshot of the working prototype (customer invoice, full lifecycle: Deposited → Issued → Received → Paid):

![E-invoicing tracking tab prototype](https://raw.githubusercontent.com/Charlymd/dolibarr-community-modules/img/einvoicing-tracking-prototype/.github/issue-assets/einvoicing-tracking-prototype.png)

### Module

einvoicing

### Use case

Today, diagnosing "why is this invoice stuck" means either reading `llx_einvoicing_lifecycle_msg` directly in the DB, or opening the generic agenda (Événements) filtered by the `EINVOICING` label, which only gives free-text entries with no structure (no status code, no actor, no visual sequencing). A dedicated tab makes the back-and-forth between Dolibarr, the Access Point and the buyer's platform legible at a glance, which is useful both for end users (support questions: "did my invoice reach my customer?") and for developers debugging a provider integration.

Scope is intentionally limited to customer invoices: supplier invoices have their own separate lifecycle/UI and are out of scope for this tab.

### Suggested implementation

[...]

PR #870, opened 2026-09-08
feat(einvoicing): add e-invoicing tracking tab on customer invoices

Charlymd:
New "E-invoicing tracking" tab on the customer invoice card: a status card per actor (Fournisseur / PDP-PA / Client), falling back to the local Dolibarr status before the first PDP exchange, plus a detailed log table of every llx_einvoicing_lifecycle_msg event.

Follows the review on #563:
- Reuses the existing "EinvoiceEventsTab" tab slot/translations (already shipped in 6 languages) instead of declaring a new tab; drops the other dead commented tab line (no translation, unused).
- Actor classification goes through EInvoicing::STATUS_* constants, not hardcoded status-code numbers.
- Adds EInvoicing::fetchLifecycleEvents($elementType, $elementId), an element-type-agnostic reader, instead of ad-hoc SQL in the page, so the supplier invoice side can reuse it later.
- New translation keys added to en_US only (fr_FR etc. come back from Transifex).

Chronogram (SVG sequence diagram) intentionally left out of this PR, to follow on top per the same discussion.


einvoicing/lib/einvoicing_lifecycle.lib.php
```diff
@@ -0,0 +1,92 @@
...
+{
+	$override = trim((string) $override);
+	if ($override !== '' && $override !== (string) $status) {
+		return einvoicingLifecyclePlain($override);
+	}
+	return $einvoicing->getStatusLabel($status);
+}
+
+/**
+ * $langs->trans() may return HTML entities (&eacute;, ...); decode to plain UTF-8 so callers that embed
```
eldy: We should not need to use such a function
If $langs->trans() is annoying you with htmlentities, just use
$langs->transnoentitiesnoconv()  instead.
Var in memory or from database ($evt['lc_status'], ...) should be utf8 without html entities.

Charlymd: Good point, thanks — removed einvoicingLifecyclePlain() entirely, einvoicingLifecycleLabel() now returns $override as-is. getStatusLabel() already goes through transnoentitiesnoconv(), and the provider override message comes straight from the DB (never through $langs->trans()), so there was nothing to decode. Fixed in 4f70b531.

PR #877, opened 2026-09-09
FOLLOW-UP #858: postpone convention, exact flow match, and other robustness fixes on incoming supplier invoice status

comaiteseb:
Depends on #858 (the minimal move + docblock fix). Scope agreed with @daGrumpf-bxp and @eldy on that PR: everything beyond the atomic move, so #858 could stay small.

- Postpone convention (`postponeflow`) instead of a bare negative result on a transient platform failure (GET error, empty response body), per the #718 rule - a malformed CDAR stays non-transient (`res=0`).
- `Document::cronSyncFlows()` now applies `EINVOICING_SYNC_MARGIN_TIME_HOURS` so a postponed flow isn't pushed out of the next cron run's sync window.
- Exact invoice match by flowId first (an incoming flow's trackingId is the flowId of the received supplier invoice flow itself, verified on Hubtimize), falling back to the vendor reference heuristic.
- `findSupplierInvoiceByVendorReference()`: check the vendor's legal identifier even on a single-candidate match, normalize VAT, compare SIREN against SIRET's first 9 digits, via a shared `matchesVendorLegalId()` helper.
- Read the results of `insertOrUpdateExtLink()`/`storeStatusMessage()` - neither throws on a SQL failure, so the rollback around them could never fire.
- `fetchFlowData()` `Original` -> `Converted` fallback for access points that only hold the converted copy.

Note: until #858 is merged, this diff includes its commits too (stacked on top of it) - it will shrink to just the new commit once #858 lands on main.


daGrumpf-bxp:
One thing worth stating plainly now that the split has landed, so this one does not sit here too long: **as `main` stands since #858, a transient failure of the platform call loses the vendor status for good.**

`processIncomingSupplierInvoiceStatus()` now returns `res = 0` when the `GET` on the flow does not answer 200 (`0fdf7542`, the two-line half of the split). The caller `break`s, and `syncFlow()` stores the flow at the end regardless of the outcome (`EsalinkPDPProvider.class.php:1803`, `SuperPDPProvider.class.php:2806`). The next run's `alreadyProcessedFlowIds` pre-check then discards that flowId silently: it is never read again, and the 212 the vendor sent is never recorded on the supplier invoice.

That is still better than what #858 replaced - `res = -1` **and** the flow stored anyway, so the batch aborted *and* the status was lost. But the trade is "abort" for "silent loss", not a fix, and this PR is what closes it:

- `return` instead of `break` on a negative result, so nothing is stored and the flowId comes back on the next listing;
- `postponeflow`, so the flows behind it are still processed instead of the batch aborting on this one (the #718 convention);
- the sync margin on `cronSyncFlows()`, so the postponed flow is still inside the window the next cron run asks for - without it, `getLastSyncDate()` being `MAX(updatedat)` over the *stored* documents, a flow deliberately left unstored can be pushed out of the cursor by any later flow that does get stored.

A malformed CDAR stays non-transient (`res = 0`, stored and skipped), so the "never stall on a flow we cannot read" guarantee of #858 is unchanged.

CI is green on the eight checks, and the branch is on `main` alone now that #858 has merged, so the diff here is just the follow-up commit.

eldy:
Sorry @daGrumpf-bxp, report is not clear for me? Does it means we need a change to merge this or is it a go for you ?

daGrumpf-bxp:
It is a go for me, no change needed: it goes in the right direction, and the risk of losing a vendor status is lower with this PR than without it.

Issue #675, opened 2026-08-30
[einvoicing] One un-importable incoming document stops the whole synchronization, with a message that does not say what to do


daGrumpf-bxp:
### Bug

One incoming document that cannot be imported stops the **whole** incoming synchronization of the
instance, for every supplier, until somebody intervenes by hand. And the message the user gets is a
raw technical line that does not say what to do.

The trigger measured here: a received invoice references a document the buyer does not hold. The
final invoice of a deposit references its deposit invoice (BT-25), a credit note references the
invoice it credits, a replacement references the one it replaces. When that document is not in the
buyer's Dolibarr, `createSupplierInvoiceFromSource()` returns `-1`
(`CIIProtocol.class.php` lines ~910, ~1027, ~1063, ~1243, same shape in `FacturXProtocol`), and
`syncFlows()` breaks out of its loop:

```
ERROR_SYNCFLOW - Failed to synchronize flow i_381565: Failed to create supplier invoice from
E-invoice document for flowId: i_381565. Document : TRI-AC-2608-0024 linked to document
TRI-FA-2608-0265 not found in Dolibarr
Aborting synchronization due to errors.
```

### What that costs, measured

Bench, `main` at 4b70999f, one instance issuing to another through the SuperPDP sandbox:

1. The failing flow stays at the head of the queue: **three consecutive synchronizations**,
   `syncedFlows: 0` each time, the same flow each time.
2. An **ordinary invoice issued afterwards**, with no relation whatsoever to the case
   (`TRI-FA-2608-0266`), **never arrives** while the block lasts. The whole incoming queue is down.
3. **The natural remedy does not work.** Having the supplier send the missing deposit invoice through
   the platform changes nothing: the synchronization aborts before it ever reaches the newer flow.
4. The only way out is to **create, by hand, a supplier invoice carrying the missing reference**.
   Done - the synchronization resumes at once and imports the four queued flows, the two waiting
   invoices included.
5. If that hand-made invoice does not match the amount, the guard added by 511901e3
   (`SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT`) stops the queue again - this time with a business
   message and an action link, which is exactly the shape the case above is missing.

The triggers are ordinary, not exotic: deposit followed by its final invoice is standard practice in
France; a credit note or a replacement of an invoice the buyer never received happens whenever the
module was installed after the fact, a flow was deleted, or an earlier import failed.

### Prior art in this repository

[...]

PR #884, opened 2026-09-09
FIX #880 [einvoicing] Skip InvoiceReferencedDocument check for standard invoices (TypeCode 380)

### LINKED ISSUE #880: [einvoicing] Standard invoice (TypeCode 380) with placeholder InvoiceReferencedDocument is indefinitely blocked on import
### Bug

## Summary

When a supplier sends a standard invoice (TypeCode 380) that contains an `InvoiceReferencedDocument` (BG-3/BT-25) with a placeholder value (e.g. `XXXX` or `0`), the module blocks the import indefinitely, asking the user to first create a supplier invoice with that placeholder reference in Dolibarr.

## Root cause

In both `CIIProtocol.class.php` and `FacturXProtocol.class.php`, the `invoiceRefDocs` existence check runs for **all** invoice types:
                                                                                                                                                                                                                               
```php
  // Applies to ALL types including TypeCode 380 — wrong
  if (!empty($parsedHeader['invoiceRefDocs']) && is_array($parsedHeader['invoiceRefDocs'])) {
   
  However, InvoiceReferencedDocument is only semantically required for:
  - 381 / 503 — credit notes (must reference the invoice being credited)
  - 384 — replacement invoices (must reference the invoice being corrected)

  For TypeCode 380 (standard invoice), BG-3 is optional per EN 16931.
  Some vendors populate it with a placeholder value when it is not applicable,
  instead of omitting the block entirely. The module should not block the import in that case.                                                                                                                                                                                                                                  

  Note: the code that actually uses invoiceRefDocs to set fk_facture_source (a few lines below) is already correctly restricted to credit notes and replacement invoices — making the preceding existence check inconsistent.

  Expected behaviour

  A standard invoice (TypeCode 380) with a placeholder or irrelevant InvoiceReferencedDocument is imported normally. The existence check is only triggered for TypeCodes 381, 503 and 384.                                                                                                                                                                                                      
   
  Proposed fix

  Restrict the check to the invoice types that genuinely require it:

  $refDocCheckTypes = ['381', '503', '384'];
  if (in_array($parsedHeader['documenttypecode'] ?? null, $refDocCheckTypes)
      && !empty($parsedHeader['invoiceRefDocs'])
      && is_array($parsedHeader['invoiceRefDocs'])) {

```
  Same change in both CIIProtocol.class.php and FacturXProtocol.class.php.

  Related
   
  - #675 — import failure stops the whole synchronisation (partially addressed by #676, but the root cause for TypeCode 380 was not fixed) 

### Module

einvoicing

### Dolibarr Version

_No response_

### Environment PHP

_No response_

### Environment Database

_No response_

## Steps to reproduce

1. Receive a CII/Factur-X invoice from a supplier with TypeCode 380 and an `InvoiceReferencedDocument` containing a placeholder value such as `XXXX`.
2. Run a synchronization.
3. The invoice is postponed with the message:
     > "The received invoice XXXXXXX refers to the invoice XXXX, which is not in
     > your database. It is needed to link the two, so the flow was not imported
     > and will be retried on the next synchronization."
4. Every subsequent synchronization produces the same result

### Attached files

_No response_

Boushabazakaria:
## FIX #880 [einvoicing] Skip InvoiceReferencedDocument check for standard invoices (TypeCode 380)
A standard invoice (TypeCode 380) containing an `InvoiceReferencedDocument` (BG-3/BT-25) with a placeholder value (e.g. `XXXX`, `0`) was indefinitely blocked on import: the module searched for that reference in Dolibarr, found nothing, and postponed the flow on every synchronization.

The existence check is only meaningful for invoice types that actually need a linked document in Dolibarr: credit notes (381/503) and replacement invoices (384). For standard invoices, BG-3 is optional per EN 16931 and some vendors populate it with a placeholder when it does not apply.

The fix restricts the check to TypeCodes 381, 503 and 384 in both CIIProtocol and FacturXProtocol, consistent with the downstream code that already limits `fk_facture_source` linking to those same types.
                                                                                                                                                                                                                                                
Fixes #880

### DIFF
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -924,7 +924,9 @@ class CIIProtocol extends AbstractProtocol
 		}
 
 		// Check if all referenced documents in the invoice exist in Dolibarr for the same supplier, if not return with error since we need them for correct linking in the invoice
-		if (!empty($parsedHeader['invoiceRefDocs']) && is_array($parsedHeader['invoiceRefDocs'])) {
+		// Only check for invoice types that require a linked document (credit notes 381/503, replacement 384) — for standard invoices (380) BG-3 is optional and may contain vendor placeholders.
+		$refDocCheckTypes = ['381', '503', '384'];
+		if (in_array($parsedHeader['documenttypecode'] ?? null, $refDocCheckTypes) && !empty($parsedHeader['invoiceRefDocs']) && is_array($parsedHeader['invoiceRefDocs'])) {
 			foreach ($parsedHeader['invoiceRefDocs'] as $invoiceRefDoc) {
 				$refDoc = $invoiceRefDoc['IssuerAssignedID'] ?? null;
 				$dateDoc = $invoiceRefDoc['FormattedIssueDateTime'] ?? null;
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -924,9 +924,7 @@ class CIIProtocol extends AbstractProtocol
 		}
 
 		// Check if all referenced documents in the invoice exist in Dolibarr for the same supplier, if not return with error since we need them for correct linking in the invoice
-		// Only check for invoice types that require a linked document (credit notes 381/503, replacement 384) — for standard invoices (380) BG-3 is optional and may contain vendor placeholders.
-		$refDocCheckTypes = ['381', '503', '384'];
-		if (in_array($parsedHeader['documenttypecode'] ?? null, $refDocCheckTypes) && !empty($parsedHeader['invoiceRefDocs']) && is_array($parsedHeader['invoiceRefDocs'])) {
+		if (!empty($parsedHeader['invoiceRefDocs']) && is_array($parsedHeader['invoiceRefDocs'])) {
 			foreach ($parsedHeader['invoiceRefDocs'] as $invoiceRefDoc) {
 				$refDoc = $invoiceRefDoc['IssuerAssignedID'] ?? null;
 				$dateDoc = $invoiceRefDoc['FormattedIssueDateTime'] ?? null;
@@ -937,9 +935,17 @@ class CIIProtocol extends AbstractProtocol
 					return ['res' => -1, 'message' => SupplierInvoiceHelper::refLookupErrorMessage($refDocInvoiceId, $refDoc, 'required by received document ' . ($parsedHeader['documentno'] ?? ''))];
 				}
 				if ($refDocInvoiceId == 0) {
-					// The document references an invoice this Dolibarr does not hold. Nothing has been created at this
-					// point, so the flow is postponed and retried on the next synchronization rather than failed, and
-					// the message spells out what to create with a link to the screen that creates it.
+					// An unqualified reference (no ram:TypeCode in the XML) is a placeholder that the import
+					// does not consume — some vendors (e.g. DSV Road) always emit BG-3 with a dummy value
+					// such as "XXXX" when no preceding invoice applies. Skip it silently so it does not block
+					// the import and does not reach the post-creation loop.
+					if (empty($typeDoc)) {
+						dol_syslog(get_class($this) . '::doCreateSupplierInvoiceFromSource Skipping unqualified InvoiceReferencedDocument ref="' . $refDoc . '" (no TypeCode) for ' . ($parsedHeader['documentno'] ?? ''), LOG_DEBUG);
+						continue;
+					}
+					// The document references a qualified invoice this Dolibarr does not hold yet. Nothing has
+					// been created at this point, so the flow is postponed and retried on the next
+					// synchronization rather than failed, and the message spells out what to create.
 					$langs->load("bills");
 					$action = $langs->trans('CreateTheMissingSupplierInvoiceToImport', $refDoc);
 					$action .= ' <a class="butAction small smallpaddingimp nomarginleft" href="' . DOL_URL_ROOT . '/fourn/facture/card.php?action=create&socid=' . (int) $socId . '&ref_supplier=' . urlencode($refDoc) . '" target="_blank">';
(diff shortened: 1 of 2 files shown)


daGrumpf-bxp:
I ran this branch on a bench before it lands, because the check it removes is a guard rather than a nuisance.

## What the pre-check at `CIIProtocol:927` protects

`invoiceRefDocs` is consumed a second time **after** `create()`, in the loop at `CIIProtocol:1063`, and that loop runs for **every** invoice type. Its not-found branch returns a bare failure:

```php
if ($linkedObjectId == 0) {
    return ['res' => -1, 'message' => 'Document … was not found in Dolibarr'];
}
```

No `postponeflow`. `syncFlows()` reads exactly that flag (`SuperPDPProvider:1973`): with it, the flow is postponed and the batch carries on; without it, `$error++` then `break` — *"Aborting synchronization due to errors."* (`SuperPDPProvider:2088`).

So skipping the pre-check for 380 does not remove the failure. It moves it past invoice creation, onto the branch that stops the whole synchronization.

## Measured

Dolibarr 22.0.5, PHP 8.3. Real import through `createSupplierInvoiceFromSource()`, inside a transaction that is rolled back. Two CII 380 documents, identical but for the BG-3 block:

* **A** — the placeholder of the issue: `<ram:IssuerAssignedID>XXXX</ram:IssuerAssignedID>`
* **B** — a French deposit chain: a standard invoice deducting a deposit invoice the buyer does not hold

main (`d7bfdd64`):

```
A  res=-1  postponeflow=1  actioncode=LINKED_INVOICE_NOT_FOUND   invoice created: none
   -> flow POSTPONED, the batch goes on, retried next run
B  res=-1  postponeflow=1  actioncode=LINKED_INVOICE_NOT_FOUND   invoice created: none
   -> flow POSTPONED
```

this branch (`5ed9a67a`):

```
A  res=-1  postponeflow=(absent)  actioncode=(none)
   invoice created before the failure: id 16100  ref_supplier=PROOF884-…-A  ttc=540.27
   -> SYNCHRONIZATION ABORTED, every flow behind it stalls
B  res=-1  postponeflow=(absent)  actioncode=(none)
   invoice created before the failure: id 16101
   -> SYNCHRONIZATION ABORTED
```

The error message is word for word the same in both columns. What changes is that the buyer now ends up with neither the invoice (the rollback in `syncFlow()` drops it) nor a working synchronization — which is #675 back in place.

And it is not only about exotic placeholders. Case B is the ordinary French deposit chain, and this module emits that BG-3 on standard invoices itself: `buildinvoicelines.inc.php` line 485 (deposit, 386), line 366 (previous situation — situation invoices are transmitted as 380), line 912 (source invoice of a discount). Any of those, received while the referenced document is missing, goes from "postponed" to "batch aborted".

The same diff on `FacturXProtocol` has the same effect through the same code.

## Where the fix probably is

The decision cannot rest on the type of the carrying document, because a 380 legitimately references documents the import does consume. It has to rest on what the document says about the reference itself.

**Correction to the first version of this comment**, which pointed at #855's rule — telling an invoiced object identifier from a document reference by the presence of a `ReferenceTypeCode` — as the model to follow. It is not one. UNTDID 1153 is largely documentary (`IV`, `ON`, `CT`, `AFL`…), so that code being present says nothing about whether a document is meant, and testing it *before* the lookup costs a deposit deduction: measured on a bench (Dolibarr 22, real deposit invoice, import rolled back), a line pointing at an existing deposit gets no discount at all as soon as its reference carries `AWV` — or `AFL`, which does designate a document. The qualification is only safe once the lookup has already come back empty, where it decides how loudly to report, not whether to look.

There is material for that here: the CII parser never reads the entry's own `ram:TypeCode` (`CIIProtocol:2004` fills only `IssuerAssignedID` and `issueDate`), so the `$invoiceRefDoc['TypeCode']` read at `:931` is always null. A reference qualified as a deposit (386) or as a source invoice (381/384) is one the import needs; an unqualified placeholder is not. Reading that code, and keeping the pre-check for the references that are genuinely used, would close #880 without giving up the guard — and without the second loop being reached with a reference it cannot resolve.

Glad to help with a patch along those lines if that is useful.

Boushabazakaria:
@daGrumpf-bxp @eldy The root cause is that DSV Road always emits `<ram:InvoiceReferencedDocument>` with `XXXX` as a placeholder on standard TypeCode 380 invoices, even when no preceding invoice applies. Since `XXXX` is never found in Dolibarr, the pre-creation check postpones the flow indefinitely.
                                                                                                                                                                                                                                                
The fix reads `ram:TypeCode` from within `InvoiceReferencedDocument`. When absent (as in any unqualified/placeholder reference), the entry is silently skipped.
When present (qualified reference to a deposit or source invoice, as used in Extended profiles), the existing check/postpone logic runs as before.

Applied to both CIIProtocol and FacturXProtocol, in both the pre-creation guard and the post-creation safety net.

daGrumpf-bxp:
The placeholder is a real case and the direction is right — do not stop the batch on a reference that designates nothing. One thing about the discriminator, though, checked against the documents this module itself emits.

**`ram:TypeCode` inside `InvoiceReferencedDocument` is forbidden below EXTENDED.** CII-DT-018 only tolerates a TypeCode on an `AdditionalReferencedDocument`, which is why the emission of this module writes it under `isExtendedProfile()` and nowhere else:

```php
// The document type (BT-X-...) is a fatal error under EN 16931, where CII-DT-018 only
// tolerates a TypeCode on an AdditionalReferencedDocument. The EXTENDED and
// EXTENDED-CTC-FR profiles reinstate it (CII-EXT-DT-018).
if ($this->isExtendedProfile($profile)) {
```

So on an EN 16931 document the TypeCode is always absent — not because the reference is a placeholder, but because the norm forbids writing it. The two committed specimens of this repository show it:

```
cii_creditnote.xml    InvoiceReferencedDocument: IssuerAssignedID, FormattedIssueDateTime   TypeCode: absent
cii_replacement.xml   InvoiceReferencedDocument: IssuerAssignedID, FormattedIssueDateTime   TypeCode: absent
```

Both are `urn:cen.eu:en16931:2017`, both generated by this module. And the rule CII-DT-018 is not a style preference — it is `flag="fatal"`: `not(ram:TypeCode) or (self::ram:AdditionalReferencedDocument ...)`, "TypeCode should not be present".

Correcting myself on the size of the consequence, having now read the loop rather than assumed it: a credit note loses nothing, because the document level loop only consumes the reference when the object it finds is a deposit — for anything else there is just the comment "Other linked document handling can be implemented here". The loss is the deposit, and only the deposit:

**a final invoice after deposit, in the EN 16931 profile, has its BG-3 reference skipped, so the discount is never created and the invoice is imported for the full amount** — silently, since the skip only writes a LOG_DEBUG line and nothing reaches the import messages. That is the defect that disqualified #855, arriving from the other side.

**What the reference designates cannot be read from a field the norm forbids.** What can be read is the document's own arithmetic, and it says when the reference matters: BT-113, `TotalPrepaidAmount`. If the document declares an amount already paid and points at an invoice we do not hold, that is a missing deposit and stepping over it imports an invoice short of its deduction. If it declares nothing paid, the reference costs nothing to step over — placeholder or contract number alike.

That gives the same three-way rule #911 lands at line level, one storey up:

1. reference found → link it, as today;
2. not found and BT-113 > 0 → postpone the flow, nothing is stored, `syncFlow()` rolls back and the next run takes it again;
3. not found and nothing declared paid → step over it and report it in the import messages.

The `XXXX` of DSV Road falls in case 3 and stops blocking, with no dependency on a profile-specific field. Happy to bench that against your document if you want a measurement before changing anything — the harness from #911 takes any CII file.

Unrelated, for merge order: this branch and #911 touch different loops of the same file — document level here, line level there — and `git merge-tree` merges them cleanly in both orders.

PR #889, opened 2026-09-09
Qual: buildXML() takes the output language it is already handed

daGrumpf-bxp:
Rewritten after the review: the argument stays, the signature takes it.

## Problem

`generateInvoice()` passes `$outputlangs` as a fourth argument to `buildXML()`,
which declares three:

```php
$xmlcontent = $this->buildXML($invoiceData, $linesData, $this->getBuildXmlProfile(), $outputlangs);
//  public function buildXML(array $invoiceData, array $linesData, $profile = '')
```

PHP accepts extra arguments on a userland call and drops them, so the language
never arrived. Since some of the document will have to be written in the
language of its recipient, declaring the parameter is the way to keep it, not
removing the argument.

## Fix

Declare `$outputlangs`, and normalise it the way `generateInvoice()` already
does. It stays apart from `$langs`: the only two strings the method translates
today are the "IBAN missing for the seller" errors, which are meant for the
user working, not for the recipient of the document.

## Testing

Bench, eight instances, Dolibarr 18.0.10 / 19.0.4 / 20.0.4 / 21.0.4 / 22.0.5 /
23.0.4 / 24.0.1 / 25.0.0 (development), each under the PHP version its core
targets (7.4 to 8.5):

- the **24 XML files regenerated from the same invoices** across the eight cores
  are byte for byte what they were, apart from the generation stamp that carries
  the commit;
- 256 PHPUnit runs, 126 e-invoice generations (7 invoice types x CII and
  Factur-X x 9 instances), 16 end-to-end cycles and 48 pages, all unchanged.


eldy:
In a future may be some content of the xml will need to be in the language of the customer.  So $outputlang will be used.

daGrumpf-bxp:
Rewritten the other way round, following your remark: `$outputlangs` is now a declared parameter of `buildXML()` instead of the argument being dropped at the call. It is normalised the way `generateInvoice()` does it, and kept apart from `$langs` so the two error messages the method translates stay in the language of the user working.

PR #896, opened 2026-09-09
NEW einvoicing: queue flows needing a manual action instead of aborting the whole synchronization

Pichinov-Jose:
## What

A received flow that needs a manual action still aborts the whole synchronization: `PRODUCT_NOT_FOUND`, `THIRDPARTY_NOT_FOUND` and `SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT` reach the `break` in `SuperPDPProvider::syncFlows()`.

Nothing is stored for that flow, so `getLastSyncDate()` — `MAX(updatedat)` over the **stored** documents — never moves past it. The next scheduled run stops on the same flow, and everything behind it (the outbound status refreshes included) stalls with it, run after run, until an operator fixes that one flow by hand.

This is the same stall #855 and #879 remove for the failures the operator cannot act on. This PR removes it for the ones that **do** need an operator action, without losing them or blocking everything else in the meantime.

## How

Instead of aborting, the flow is recorded in a persistent manual-action queue (`llx_einvoicing_sync_pending`) and the run carries on — the same spirit as the existing `postponeflow`, but for failures that need a manual step, so the flow is kept rather than retried blindly.

A new page (**Billing → E-invoicing documents → Manual-action queue**) lists the queued flows with, per row:
- the action(s) the protocol already computes — create the product/service or thirdparty, associate an existing product, set a default product for the issuer;
- a queue-only action to link the issuer to an **existing** thirdparty by its fiscal identifiers (SIREN / SIRET / VAT), with a field-by-field comparison and a uniqueness pre-check;
- a one-click **Retry**.

On the next successful sync the flow leaves the queue on its own, linked to the supplier invoice it created (`fk_element_type` / `fk_element_id`).

Touch points:
- `SuperPDPProvider::syncFlows()`: on the three manual-action codes, **queue + `continue`** instead of `break`; **resolve** on success; counted in the recap (with a link to the queue) and in the return array.
- New `EInvoicingSyncPending` CommonObject + table (**additive**, `entity`-scoped, unique on `(entity, provider, flow_id)`).
- `modEInvoicing`: one menu entry.
- fr/en translations.

Additive only: no existing schema or behaviour changes for flows that already synchronize; a flow that later succeeds is simply removed from the queue.

## Screenshots

_To add: the manual-action queue page (action icons + retry) and the synchronization recap showing the queued count with its link._
einv_pr_file
einv_pr_comparaison

## Test

_Validated end-to-end on a real received invoice (2 product lines, both `PRODUCT_NOT_FOUND`): the run no longer aborts, both lines are captured in the queue across retries, and once the products are created/associated a Retry imports the supplier invoice and the row is resolved._

---
Happy to split the behaviour change (queue + continue) from the management page if that reviews more easily.


eldy:
> Nothing is stored for that flow, so getLastSyncDate() — MAX(updatedat) over the stored documents — never moves past it. The next scheduled run stops on the same flow, and everything behind it (the outbound status refreshes included) stalls with it, run after run, until an operator fixes that one flow by hand.

Yes, this is the expected behaviour. Because the order is very important to avoid update action that will fails if not in the same order, we need to stop as soon as a manual action is required.
Trying to going forward is a avery dangerous behaviour (clearly not recommended for the moment when process l generates critical situation).
I recommend you for the moment that all the code of this feature is conditionned by a hidden option so we can enable/debug it without having end users experience the troubles this may bring...

Pichinov-Jose:
Hi @eldy, thanks — that makes sense: the ordering constraint is a good reason to keep the abort as the default.

I've gated the whole feature behind a hidden option `EINVOICING_ENABLE_MANUAL_ACTION_QUEUE`, off by default:
- **Off (default):** `syncFlows` aborts on the first flow needing a manual action exactly as before — nothing is queued, nothing is resolved — and the "File d'attente synchro" left-menu entry is hidden.
- **On:** the skip-and-continue queue is active, so it can be enabled and debugged without changing the default experience.

I also merged the latest `main` (the `fr_FR` lang conflict is resolved). Ready for another look.

Pichinov-Jose:
@daGrumpf-bxp @eldy — flagging the overlap with #932 so we don't end up with two parallel backlogs.

#932 persists and replays the **postponed** flows (a reference not yet in the base, a missing deposit) — non-actionable failures where the fix is to wait or to record the missing document — and its replay-by-identifier addresses the root cause I built this PR around: a flow dropping out of the `getLastSyncDate()` window as newer traffic pushes `MAX(updatedat)` forward. That replay is a more general answer to "don't lose the flow" than the queue here.

What #896 adds that #932 does not is the **actionable** side: the business errors an operator can resolve directly — `PRODUCT_NOT_FOUND`, `THIRDPARTY_NOT_FOUND`, `SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT` — surfaced in a dedicated list with the operator actions (create the product, associate an existing product/thirdparty, set a default, then retry that single flow). It is now gated behind the hidden option `EINVOICING_ENABLE_MANUAL_ACTION_QUEUE` (off by default), per the review above.

The real overlap is the persistence + panel. Rather than ship a second table (`llx_einvoicing_sync_pending`) and a second page next to #932's `llx_einvoicing_postponed`, I'm happy to **rebase #896 on top of #932 once it lands** — reusing its backlog infrastructure and keeping only the actionable-UI layer here — or to fold the manual-action rows into #932's model if you'd prefer a single backlog.

How would you like to sequence this? I can wait for #932 to merge and realign #896 onto it.

eldy:
I understand your feature of storing flow we can't process, so we can process them later.

For me this should replace completely the postpone feature (the actual postpone feature must be removed in favor of a table storing solution, activated on option only)
This means this PR make the PR #932 useless.

@daGrumpf-bxp Can you confirm ?

daGrumpf-bxp:
@eldy No, I cannot confirm that — #896 does not make #932 useless, for three reasons.

**1. #896 is off by default, so it cannot carry the fix.** On 12 September you asked for the whole
feature to sit behind a hidden option, and @Pichinov-Jose did exactly that on the 15th:
`EINVOICING_ENABLE_MANUAL_ACTION_QUEUE`, off by default, with `syncFlows()` aborting on the first
flow needing a manual action exactly as before when it is off. That is the right call for a queue
that changes the processing order. But it means that if `postponeflow` is removed in favour of that
table, then for every user who does not turn the option on — that is, everyone, by default — a
postponed flow is no longer taken again at all. The bug below would be fixed for nobody.

**2. The bug it would leave open is not theoretical — I measured it.** This is the point you raised
yourself on #932 and #933 on 14 September, and you were right. The window starts at
`getLastSyncDate()` = `MAX(einvoicing_document.updatedat)` over the **stored** documents, minus
`EINVOICING_SYNC_MARGIN_TIME_HOURS` (12 by default), and the API filter is `updatedAfter`. So a
postponed flow is listed again only while

```
updatedAt(postponed flow) > MAX(updatedat of stored flows) - 12h
```

`MAX` counts every stored flow, the lifecycle messages of our own outgoing invoices included, so the
cursor is pushed forward by traffic that has nothing to do with the postponed document: the busier
the instance, the sooner it drops out.

On my bench, on real data I did not stage, **four received supplier invoices were already lost that
way** — all `LINKED_INVOICE_NOT_FOUND`, one postponed across five runs between 10 and 13 September,
each run writing *"it will be retried on the next synchronization"*, and by then sitting 74 hours
behind the start of the window; three more on another instance, 262 hours behind. A real run of the
cron path then answers *"Synchronization completed successfully — 10 new flows synchronized"*,
**does not mention the lost flow at all**, and pushes `MAX` further away. The promise the message
makes to the operator is not kept, silently.

#932 fixes precisely that, and unconditionally: `replayPostponedFlows()` takes each backlog row
again **by `flowId`**, which uses no date window at all, at the head of `syncFlows()` in both
providers. Recetted before and after on the same bench instance: before, two cron runs say
"completed successfully, 0 new" and the flow is never mentioned; after, "Total of flows postponed: 1"
with `nb_attempts` climbing; then with the missing deposit recorded, one run imports the invoice at
400.00 net / 480.00 gross — the 120.00 deducted — and `MAX(updatedat)` never moved.

**3. The two do not cover the same failures.** They are complementary, not redundant:

| | failure | what unblocks it | whose job |
|---|---|---|---|
| **#932 / #933** | a reference the base does not hold yet: an awaited deposit, a linked invoice, a duplicate reference on the lookup | **waiting**, or recording the missing document in the normal course of business | no operator action on the flow itself — it must simply be taken again, forever, until it works |
| **#896** | `PRODUCT_NOT_FOUND`, `THIRDPARTY_NOT_FOUND`, `SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT` | an **operator decision**: create the product, associate an existing one, link the issuer to a thirdparty | a dedicated screen, and worth an option because it changes the processing order |

The second needs a queue an operator works through. The first must not need anyone: there is nothing
for a human to do about a deposit that has not arrived yet, so putting it in a manual-action queue
behind an option would be asking an operator to act on flows that will resolve by themselves.

**What I do agree with**, and it is most of your point: there should be **one** storage mechanism,
not two parallel backlogs. #932 already owns a table for this (`llx_einvoicing_postponed`), and
@Pichinov-Jose said the same thing on the 15th when he flagged the overlap. If you want a single
table serving both, I am happy to do the work — #932 landing first with the replay unconditional,
then #896 storing its actionable failures in that same table rather than in
`llx_einvoicing_sync_pending`, keeping its option and its screen. What I would ask not to do is
remove `postponeflow` and leave the replay behind an option that is off: that trades a bug the
operator cannot see for a bug nobody is protected from.

#933 is a separate matter and does not depend on this: without it, a reference lookup answering `-3`
still stops the whole synchronization on every run, for every user. It only makes two more failures
reach the existing `postponeflow` convention.

Pichinov-Jose:
Thanks @eldy, and thanks @daGrumpf-bxp for the measured analysis — I've read both.

On "#896 should replace the postpone feature completely": I don't think #896 can *replace* it, and @daGrumpf-bxp's measurements are the reason. #896 is gated **off by default** (as you asked on 12 Sep, since it changes the processing order), so if the postpone path were removed in its favour, postponed flows would stop being retried for everyone who doesn't switch the option on — i.e. by default. And the two cover different failures: #896 is for the **operator-actionable** ones (`PRODUCT_NOT_FOUND`, `THIRDPARTY_NOT_FOUND`, `SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT`) that need a screen and a decision; the postpone / #932 case (an awaited deposit, a linked invoice) has nothing for an operator to do and must just be replayed automatically.

So I'd keep them **complementary in behaviour** — but I fully agree we should **not** ship two parallel backlog tables. The clean convergence: let #932 land its `llx_einvoicing_postponed` backlog, and **I rebase #896 on top to reuse that single table** (the manual-action rows becoming one kind of backlog entry), keeping only the operator screen and the retry/associate actions here. That removes the duplication you're pointing at while keeping both behaviours.

If that works for you, **#932 can go ahead and I'll adapt #896 onto it.** CI here is green in the meantime.

PR #897, opened 2026-09-10
FIX [einvoicing] A referenced document unknown here no longer aborts the synchronization

ATM-NicolasV:
## Problem

A received line may reference another document through BT-128 (`ram:AdditionalReferencedDocument`). The import required **every** one of them to match a supplier invoice already present in Dolibarr, whatever the referenced document actually was.

Seen in production: a vendor carried a **contract number** there. `findIdByRef()` returned 0, the import returned an error, and the whole synchronization run stopped on that flow. The invoices created earlier in the same run were reported as synchronized while they were not kept, which is confusing for the operator and leaves the queue stuck on a document that will never resolve.

The type of the referenced document is read right above, into `$lineRefDocType`, and never used — a single occurrence in the module.

## What changes

That reference is how a deposit gets attached to a line. It is not a condition for the invoice to be valid, and an issuer is free to carry an order or a contract number there. A reference matching no supplier invoice is now **reported in the messages of the import and stepped over**, and the invoice is created.

A deposit that really is missing is not silent either: the invoice then totals less than its document announces, which `alignInvoiceTotalsWithDocument()` already catches and flags on the invoice.

Filtering on the type code was the other option. It was not taken: it only covers the type codes we happen to know about, while an unresolved reference should not fail a valid invoice whatever its type.

Unchanged on purpose:

- a lookup that fails technically (`findIdByRef()` < 0) still fails the import;
- a document found but not readable still fails the import;
- deposit handling is untouched.

## Structure

The loop moves out of `createSupplierInvoiceLinesFromSource()` into `resolveLineLinkedDocuments()`, next to the other `resolveLine*` helpers, so the decision can be tested on its own. An existing `// TODO` about deposits was kept, next to the deposit branch where it applies.

## Tests

`test/phpunit/LineLinkedDocumentTest.php`, 4 cases: an unknown reference does not fail the line, it is reported and names the reference, a line referencing nothing stays silent, an empty identifier is skipped without a message.

Written before the fix and watched failing. Full module suite run before and after: same failures either way, minus the four this adds — no regression. `php -l` clean, phpcs clean against `dev/setup/codesniffer/ruleset.xml`.

### DIFF
--- a/einvoicing/ChangeLog.md
+++ b/einvoicing/ChangeLog.md
@@ -4,6 +4,7 @@
 
 ## 1.2.0
 
+FIX: [einvoicing] A document referenced by a received line and unknown here no longer aborts the synchronization
 FIX: #853 [einvoicing] The dates of a received document keep the day they state
 FIX: [einvoicing] The CDAR date test errors on Dolibarr 19 since it landed on main
 FIX: [einvoicing] Enhance status handling
--- a/einvoicing/class/protocols/CIIProtocol.class.php
+++ b/einvoicing/class/protocols/CIIProtocol.class.php
@@ -1242,61 +1242,12 @@ class CIIProtocol extends AbstractProtocol
 			// Add supplier ID to line for later use in product sync
 			$parsedLine['supplierId'] = $supplierInvoice->socid;
 
-			$is_deposit_line = 0;
-			$fk_remise = 0;
-			// --------------------------------------------------
-			// Loop on linked documents at line level
-			// --------------------------------------------------
-			if (!empty($parsedLine['additionalRefDocs']) && is_array($parsedLine['additionalRefDocs'])) {
-				foreach ($parsedLine['additionalRefDocs'] as $refDoc) {
-					$lineRefDocId = $refDoc['IssuerAssignedID'] ?? null;
-					$lineRefDocType = $refDoc['typeCode'] ?? null;
-					$lineRefDocDate = $refDoc['issueDate'] ?? null;
-
-					$linkedObjectId = SupplierInvoiceHelper::findIdByRef($lineRefDocId, (int) $parsedLine['supplierId']);
-					if ($linkedObjectId < 0) {
-						return ['res' => -1, 'message' => SupplierInvoiceHelper::refLookupErrorMessage($linkedObjectId, $lineRefDocId, 'linked to line ' . $parsedLine['lineid'])];
-					}
-					if ($linkedObjectId == 0) {
-						return [
-							'res' => -1,
-							'message' => 'Document "' . dol_escape_htmltag((string) $lineRefDocId) . '" linked to line ' . dol_escape_htmltag((string) $parsedLine['lineid']) . ' was not found in Dolibarr. Please verify why this document is missing (deleted, not imported, or not provided by the supplier). To resolve this issue, you must manually create the invoice using the supplier invoice reference "' . dol_escape_htmltag((string) $lineRefDocId) . '".'
-						];
-						// TODO: Add a check before sending a final invoice after deposit to ensure that the deposit invoice has been properly sent to the PDP and successfully received.
-					}
-
-					// Load linked supplier invoice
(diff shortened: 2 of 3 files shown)


einvoicing/ChangeLog.md
```diff
@@ -4,6 +4,7 @@
 
 ## 1.2.0
 
+FIX: [einvoicing] A document referenced by a received line and unknown here no longer aborts the synchronization
```
daGrumpf-bxp, line 7: `.agents/AGENTS.md:198` asks contributors not to touch this file: *"Do not update the `ChangeLog` file (this file will be generated before the release from all commit titles)"*. Your commit title already carries the wording, so the line can just go.

```suggestion
```

daGrumpf-bxp (commented):
One point of repo hygiene, the rest of the change is being looked at separately.

daGrumpf-bxp:
Heads-up: **#855** fixes the same loop (BT-128 at line level) with a narrower rule — it skips the reference *before* the lookup when the document qualifies it with a `ReferenceTypeCode`. Each merges cleanly on `main`, but one after the other they conflict in `CIIProtocol.class.php` (`git merge-tree`, both orders), and neither thread mentions the other.

Bench, Dolibarr 18→25, both branches merged on current main: PHPUnit green on both (this one 33 files × 8 cores, 0 KO). On the deposit path the two do differ, which an earlier version of this comment got wrong. Measured on Dolibarr 22, real deposit invoice, seven documents, import rolled back:

- this PR keeps the deposit linked in every case — a line pointing at an existing deposit gets its `(DEPOSIT)` discount whether or not the reference carries a `ReferenceTypeCode`;
- #855 loses it as soon as the reference is qualified: with `AWV`, and with `AFL` too, no discount is created at all, because the reference is never looked up. That is the false positive raised in review on that thread.

So the suggestion of re-proposing #855's rule as a `continue` *above* `resolveLineLinkedDocuments()` was wrong and I withdraw it: placed before the lookup it silently drops a deposit deduction. Placed *after* it, on the "not found" branch only, it costs nothing and is still worth having — a telecom invoice would otherwise add one import message per line.

One claim in the new docblock the bench does not confirm: *"A deposit that really is missing still shows up: the invoice then totals less than its document announces, which alignInvoiceTotalsWithDocument() catches and flags."* It does not. That method compares BT-110 and BT-112 only, and a deposit moves neither — it moves BT-113 and BT-115. Measured: a document announcing `TotalPrepaidAmount 120.00` / `DuePayableAmount 420.27` whose deposit is absent from Dolibarr is imported at the full `540.27`, no discount, no totals-mismatch mark, validation not blocked. On `main` the same document aborts the batch.

That leaves one case where stepping over a reference costs money and nothing catches it. The signal separating it from a harmless reference is in the document itself: BT-113 > 0. Postponing the flow there rather than importing keeps the batch going and stores nothing (`syncFlow()` rolls its transaction back), which is what the document level already does for BG-3 with `LINKED_INVOICE_NOT_FOUND` — and it keeps the postpone an exception, in the spirit of #718.

daGrumpf-bxp:
Opened #911 with this fix after this PR was closed: the defect it aimed at is still on `main`, and a line reference that matches no supplier invoice still aborts the whole batch.

It is not this branch as it stood — an unqualified reference (a contract number) still aborted here, and a document declaring an amount already paid whose deposit is missing was imported short of its deduction, which the totals do not catch since a deposit moves BT-113 and BT-115 rather than BT-110 and BT-112. So: step over what is not ours, postpone the flow when BT-113 says money is at stake, and read BT-128-1 only to decide how loudly a missing reference is reported.

The reading that a reference we do not hold must not stop the batch is yours, and #911 says so.

daGrumpf-bxp:
The guard on an entry whose `IssuerAssignedID` is empty has been carried from this branch into #911 and credited there. `findIdByRef()` answers `0` for an empty reference exactly as for one it cannot find, so without it a document declaring an amount already paid had its flow postponed for ever. Measured on the branch: postponed without the guard, imported with it.

PR #904, opened 2026-09-10
FIX [einvoicing] The import guard also confronts the amount the document says was already paid

daGrumpf-bxp:
## The hole

`CIIProtocol::alignInvoiceTotalsWithDocument()` confronts the imported invoice with **BT-110** and **BT-112**, and marks it when they disagree so it cannot be validated or approved (issue #861).

A deposit moves neither of those. It moves **BT-113** (`TotalPrepaidAmount`) and **BT-115** (`DuePayableAmount`), and on the Dolibarr side it is not a total at all — it is a `DiscountAbsolute` attached to the invoice. So a document that says *"540.28 including VAT, of which 120.00 already paid"* whose deposit the import could not link builds an invoice totalling **exactly** what the document announces, carrying **no deduction**, and passes the guard untouched.

BT-113 is parsed already, into `$parsedHeader['totalPrepaidAmount']`. Nothing on the reception side reads it — it is used only when the module *emits* a document.

## Measured

Dolibarr 22.0.5, PHP 8.3, real import through `createSupplierInvoiceFromSource()` in a transaction that is rolled back, one supplier deposit invoice of 100.00 excl. / 120.00 incl. VAT recorded beforehand. Six documents, identical but for their BG-22:

| document | BT-107 | BT-108 | BT-113 | `main` | this branch |
|---|---|---|---|---|---|
| plain | 0 | 0 | 0 | not flagged | not flagged |
| document level allowance | 50.00 | 0 | 0 | not flagged | not flagged |
| document level charge | 0 | 30.00 | 0 | not flagged | not flagged |
| **prepaid, deposit not linked** | 0 | 0 | **120.00** | **not flagged** | **flagged** |
| **the three together** | 50.00 | 30.00 | **120.00** | **not flagged** | **flagged** |
| prepaid, deposit linked | 0 | 0 | 120.00 | not flagged | not flagged |

The import itself is faithful on everything else: BT-106, BT-107, BT-109, BT-110 and BT-112 all match what Dolibarr stores, allowances and charges included. Only BT-113 had no reader.

Validation, same bench:

```
bg22-prelinked.xml   mark: none                                       validate() = 1   VALIDATED
bg22-pre.xml         mark: {"tva":90.05,"ttc":540.28,"prepaid":120}   validate() = -1  REFUSED
   The e-invoice this invoice was imported from announces 120 including VAT as already
   paid, and the invoice carries 0 of deduction. Record the deposit invoice the vendor
   refers to and attach it as a discount, or refuse the document to the vendor.
```

An invoice whose deposit **is** linked validates normally: the check is on the deduction actually attached, not on the presence of BT-113.

## What changes

* `SupplierInvoiceHelper::linkedDepositAmount()` — what the deposits attached to an invoice deduct from it, VAT included. A deposit is told from a document level allowance by `fk_invoice_supplier_source`, which only the former carries.
* `totalsAgreeWithDocument()` takes an optional announced BT-113 and confronts it with that amount. Passing nothing keeps the previous behaviour exactly.
* The mark carries BT-113 when that is what it is about, so the re-evaluation at validation still sees it — without that, the totals agree and the mark would be cleared on the spot. **A mark written before this carries no prepaid amount and is read as `null`**, which leaves it on the two totals alone.
* The trigger names what is missing rather than reporting a total difference that does not exist, since the figures do match.

The VAT calculation modes are not replayed for this case: the totals are already right, so replaying them would prove nothing and would only produce a misleading message.

`FacturXProtocol extends CIIProtocol`, so both protocols are covered.

## Where the check sits

BT-113 is answered **once, before the totals**, and not at each point where they are found to agree: no rounding convention explains a deposit that is not attached, so there is nothing for the conventions to say about it. That is one hook instead of three, and it leaves the VAT machinery untouched.

One consequence, stated plainly: a document that is **also** off on its VAT convention keeps the totals of the instance, since the guard returns before the conventions are replayed. The invoice is marked and blocked either way, and #910 is what makes those totals agree.

## Tested

* Bench, Dolibarr 18 to 25, PHPUnit of the module: **256 runs, 256 OK, 0 KO**.
* The six documents above, on `main` and on this branch, with the validation attempt.
* `phpcs` clean with the ruleset of the repo.

No test file is added here, per the Critical Rules of `.agents/AGENTS.md`. Glad to push the fixtures and the test class if you want them.


eldy:
@daGrumpf-bxp Sometime an standard invoice just fill prepaid to says invoice was already paid.
In this case, we don't need to link invoice to a deposit that does not exists.
So we need a better detection to know if this invoice need a cedit note to ba attached rst before beeing validated or not.
For the moment i disabled the test by forcing $isprepaidannouncedforadeposit=0 in commit  87b9ac57808cc91f5e7df05319317d20fbbe47a9

PS: i started to add exemple of anonymized cii files found in real life into the test/samples dir.

daGrumpf-bxp:
@eldy You are right, and the sample settles it. Thanks for taking the trouble to anonymise a real one — it is what made the answer findable. Follow-up in #947.

**Your question has a normative answer, and it is BT-23.** BT-113 on its own discriminates nothing: EN 16931 says only `BR-CO-16` (BT-115 = BT-112 − BT-113 + BT-114), `BR-DEC-16` and `CII-SR-484`, and no rule requires a BG-3 when BT-113 > 0. The invoicing framework is what carries the information. In the FNFE `France_RFE` schematron (V1.4, `CII/EN16931/schematron/BR-FR-Flux2-Schematron-CII.sch`):

* **`BR-FR-CO-09`**, line 1155, `flag="fatal"` — *if BT-23 is B2, S2 or M2 (invoice already paid), then BT-113 = BT-112, BT-115 = 0, and BT-9 is the payment date*;
* **`BR-FR-CO-08`**, line 1142, names the other half — *B4, S4, M4 = final invoices after a deposit*.

Your document carries `<ram:ID>B2</ram:ID>`, `TotalPrepaidAmount = 48.41 = GrandTotalAmount`, `DuePayableAmount = 0.00`. It matches `BR-FR-CO-09` to the letter. So the answer to *"May be we have also ref of deposit ?"* is yes, twice over: the framework says it outright, and failing that, **BG-3** is the only thing the import ever attaches a deposit from (`invoiceRefDocs` → invoice of type deposit → `insert_discount()`), so a document without one can carry no deduction the import missed.

**Two things I found while measuring, that are worth knowing.**

*The hole was wider than the fix.* None of the six documents I used for #904 carries a BG-3 — I checked. So that PR never actually told a missing deposit from an invoice the vendor had cashed in; it could not, its own bench documents had the shape of yours. That is on me.

*Forcing the constant to zero does not unblock every such invoice.* The prepaid branch returns before the VAT alignment of #781 runs, so an invoice a cent short of what its document announces stays short and is refused on its totals instead. Measured on 22.0.5, one of my BG-22 documents still came out at 540.27 against 540.28 announced and `validate()` still returned -1 with `87b9ac5` in place. #947 answers BT-113 only where it is a deposit, which lets the alignment run again: same document now carries 540.28 and validates.

Measured on 18.0.10, 22.0.5 and 25.0.0-alpha, real imports in transactions that are rolled back, then `validate()` on each invoice: your sample and a `B2` + BG-3 variant come out clean and validate; a document referencing an invoice that exists but is not a deposit stays marked and refused, with the message that names the missing deduction. PHPUnit of the module, 8 cores, 256 runs 256 OK.

Two things left out of #947, deliberately:

* `FacturXProtocol` does not put `businessProcessId` in its parsed header, so the BT-23 rule is blind on that path and only BG-3 applies there. Say the word and I add the reader.
* A deposit the import attaches **correctly** still cannot be validated, which predates #904: `insert_discount()` writes a negative line, so the invoice totals BT-112 − BT-113 while the guard of #861 compares it with BT-112. Reported separately in #948 with the measurement.

One last thing about the sample itself: the PHPUnit workflow copies `einvoicing/test/samples/*.cii.xml`, and the file is named `..._cii.xml`, so it is not being validated by anything yet. #949 renames it.

Issue #772, opened 2026-09-04
[einvoicing] Negative amounts on incoming invoice are seen as 0.00


apio-sys:
### Bug

Connected with PA SuperPDP. Incoming invoice (vendor OVH) contains a lot of postives and negatives that cancel them. Sounds crazy but that is how OVH support invoices work. 6 pages of them ( ! ). The endline should be 50 EUR w/o VAT. Looking at the invoice in the SuperPDP back-office, the total amount is OK. In Dolibarr though, the total is 56,90. The extra 6,90 is the sum of all the "Free starter account with
domain" lines for - 0,30 each but imported as 0,00. Hence I think negative amounts are not properly taken into consideration. Even though the OVH invoice is really messy, I think we should still be able to ingest this. I can upload sample files if needed.

### Module

einvoicing

### Dolibarr Version

23.0.3

### Environment PHP

8.4.24

### Environment Database

MariaDB 11.8.6

### Steps to reproduce the behavior and expected behavior

As explained above. I can upload the invoice if needed. Probably not here inside a Github issue. It's not very confidential but I don't think we want invoices and PII data here.

### Attached files

Let me know where I can privately send invoice samples.

daGrumpf-bxp:
This one looks fixable, and I have a candidate ready — but I need your document to confirm the mechanism before proposing it, because there are several ways an issuer can write a negative line and they do not all break the same way.

**What I think happens.** BR-27 forbids a negative item net price (BT-146), so a line that takes money off the invoice has to carry its sign somewhere else: either on the invoiced quantity (BT-129, negative), or on the net line amount (BT-131) alone. The import does not store BT-131: it hands quantity and unit price to `FactureFournisseur::updateline()` and lets the core recompute the line, so whatever `qty x price` gives is what the invoice ends up with. On a "Free starter account with domain" line — unit price `0.00`, BT-131 `-0.30` — that rebuilds `0.00` and the credit vanishes. 23 such lines x 0.30 = the 6.90 you are seeing, which fits your report exactly.

There is already the same class of repair in the module for a line whose *quantity* is zero (#726). My patch extends it to the two remaining shapes the couple (quantity, unit price) cannot express at all — the rebuild is zero, or it is of the opposite sign — and carries BT-131 as a single unit, the way you would key it in by hand. Lines that already add up are untouched.

[...]

PR #909, opened 2026-09-10
TEST [einvoicing] The documents the rules accept come back in and rebuild the same amounts

daGrumpf-bxp:
## The half nobody asks

The CI generates the five specimens and has the three FNFE stages accept them, which proves what the module **emits** is a document a platform will take. That same document is one a vendor could send **us**, so importing it back has to rebuild the invoice it came from.

Every amount defect this module has shipped was found by somebody noticing a figure on a screen: a line with an amount but no quantity imported as zero (#726), a document level charge dropped from the total (#731), a line charge lost and corrupting its line's discount (#735), a negative line dropped (#772), a vendor rounding on the other VAT convention (#781), an allowance sign and a discount without BT-137 (#783), a unit price unable to state its amount (#844), a line imported at another amount than its BT-131 (#850). Not one by the CI.

## What lands

**A gate, and nothing else.** No module code is touched: this is a CI step and the script it runs.

`import-amounts.php` imports each accepted document for real, in a transaction that is rolled back, and confronts the invoice with the document. The confrontation is **the module's own comparison**, `checkDolInvoiceAndEInvoiceConsistency()`, called unchanged — so the gate covers whatever that method covers and grows with it instead of restating it.

That method reads the document back from `einvoicing_document`, where only a synchronization puts it. The row a synchronization would have written is written by the script, so the comparison runs on its usual footing; it goes back with everything else at the rollback.

**BT-113 is read on top**, because it moves neither BT-110 nor BT-112: an invoice short of what the vendor says he already holds rebuilds every total that comparison reads, and passes. What the script reads back is **the mark the import left** — so whether that amount is a deposit to attach at all stays BT-23's business, decided by the import (#904, #947) in one place, not re-derived here where a second reading would drift from the first.

## The controls

A green gate says nothing until it is shown to refuse something. Two controls, both built from the documents that were just accepted, and both measured on the bench against Dolibarr 24:

| control | what it damages | gate |
|---|---|---|
| grand total moved by one cent | `cii_standard.xml` | refused on the total including VAT |
| a deduction announced and not attached | `cii_replacement.xml`, `BT-113 = 540.27` against the invoice it names (BG-3) | refused naming the deduction |

The second one needs building: **not one of the five specimens announces an amount already paid at document level** — `cii_deposit.xml` carries `BT-113 = 0` and references no preceding invoice — so nothing in the corpus reaches that reading.

And the controls are themselves controlled: run against a script that does not read BT-113, the second one comes back `exit=1`, which is what makes its green worth something.

No false positive: the five specimens still rebuild their amounts, on the module exactly as `main` has it.

Two details the corpus needs: `--id-from-name` gives each specimen back a document number of its own — `normalizeSampleInvoiceXml()` flattens the timestamp, so the five carry the same BT-1, and imported as they are they collide and one invoice is confronted with another document. And a document whose import found no referenced invoice is tried once more after the others are in, so a credit note lands after the invoice it corrects.

## Tested

Bench, Dolibarr 24, on the module as `main` has it: the five specimens rebuilt, both controls red where they have to be, the BT-113 control proven to fail against a gate that is not looking. No PHPUnit test file is added, per the Critical Rules of `.agents/AGENTS.md` — this is a CI step next to the schematron ones, in the shape those already use.

## Where this is going

The three amount dimensions were split across three branches so they could be reviewed apart: **BT-113** and the deposits in #904, the **vendor's rounding convention** in #910, both merged on 12/09, and the loop closed here.

What this pull request deliberately leaves out is anything that would change the module to make the gate say more: confronting **BT-106** against the lines, **BT-107** against the attached allowances, and the document with itself on **BR-CO-13** and **BR-CO-16**. Those widen what the module reports on validation, they are worth having, and they are worth judging on that — not on the fact that they make a CI step more talkative. They will come as their own pull request.


daGrumpf-bxp:
The one document this gate cannot rebuild is answered by #910: the situation specimen announces a VAT the core produced through `calcul_price_total()` at 50% progress (`45.03` on a base of `225.11`), which recomputing from the base never gives. With that branch the import carries the announced amounts onto the invoice and this gate reports **5 of 5** on the same corpus.

So this stays draft until #910 lands, then it goes green without a change of its own.

PR #910, opened 2026-09-10
FIX [einvoicing] A received invoice carries the VAT its document announces

daGrumpf-bxp:
## The defect

A received invoice had to be rebuildable **by calculation** or it was marked as not totalling its document, which keeps it out of validation and approval (#861). A vendor invoice does not owe us that: its own tool may round the VAT differently, and what it announces is what it bills.

The module emits such a document itself. Its situation specimen announces a taxable base of `225.11` and a VAT of `45.03`, and that figure is the **core's**:

```
calcul_price_total(5, 100.05, 10, 20, 0, 0, 0, 'HT', 0, 0, null, array(), 100)
    total_ht = 450.23   total_tva = 90.04   total_ttc = 540.27
calcul_price_total(5, 100.05, 10, 20, 0, 0, 0, 'HT', 0, 0, null, array(),  50)
    total_ht = 225.11   total_tva = 45.03   total_ttc = 270.14
```

The full VAT rounds to `90.05`, half of it is `45.025` → `45.03`; recomputing from the base gives `45.02`. So a customer receiving a situation invoice issued by Dolibarr got it marked and could not validate it.

## What it does

When the invoice does not total what the document announces, the announced VAT is written onto the lines that carry each rate, and the header sums them. `update_price()` sums the VAT the lines hold rather than recomputing it, so the invoice totals what was sent.

It is **not** called afterwards: `'auto'` falls back to the mode that rebuilds every line from its unit price when the instance carries no `MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND_SUPPLIER`, and would undo what was just written.

## It replaces the replay of the two VAT conventions

`alignInvoiceTotalsWithDocument()` used to recompute the invoice in mode 2, then in mode 1, looking for the vendor's convention. That is gone: 100 lines down to 71, one mechanism instead of two.

Measured first, on the cases the replay exists for — a vendor rounding on mode 2 read by an instance on mode 1, and the other way round. Three lines of `1.09 / 7.79 / 1.59` at 20%, one document announcing `2.09`, another announcing `2.10`:

| | header (ht / VAT / incl.) | VAT per line |
|---|---|---|
| replay, instance mode 1, document mode 2 | 10.47 / 2.09 / 12.56 | 0.22 1.56 0.31 |
| **carry**, instance mode 1, document mode 2 | 10.47 / 2.09 / 12.56 | 0.22 1.56 0.31 |
| replay, instance mode 2, document mode 1 | 10.47 / 2.10 / 12.57 | 0.22 1.56 0.32 |
| **carry**, instance mode 2, document mode 1 | 10.47 / 2.10 / 12.57 | 0.22 1.56 0.32 |

The same invoice, line VAT for line VAT, in all four combinations, without recomputing anything twice. They agree because the core, in mode `'1'`, carries the residual difference onto one line exactly as this does.

## What it refuses

So that "carry what the document says" does not become "believe anything", `carryAnnouncedVatOntoLines()` writes nothing and the invoice stays marked when:

* the **taxable base of a rate is not the announced BT-116** — that is a line read wrong, not a rounding convention;
* the VAT differs by **more than a cent per line of the rate** — the most a convention can move, each line VAT being rounded once and their total once more. A document announcing `3.00` of VAT on a base of `10.47` is still marked and still blocks;
* a rate the document announces has no line to carry it, or the document carries no VAT breakdown.

## The rounding setting does not do this instead

`MAIN_ROUNDOFTOTAL_NOT_TOTALOFROUND_SUPPLIER` decides whether a stored amount survives a recomputation, not where it comes from. On `main`, the situation document imports at `45.02 / 270.13` with the setting **absent, at `0` and at `1`** alike. It is also an accounting choice belonging to the instance, for every supplier invoice it holds.

Worth knowing about the amounts written here: `update_price()` is called by `create()`, `addline()`, `updateline()`, `deleteLine()` and `insert_discount()`, **not** by `validate()` nor by a payment. A received invoice keeps the announced totals through validation, approval and payment, and loses them only if a line is edited or a discount attached — at which point it no longer matches its document anyway, and the mark returns on the next comparison.

## Tested

* Bench, Dolibarr 18 to 25: PHPUnit **256 runs, 256 OK**.
* The five documents the CI generates, imported back: all five rebuild their amounts (that is what #909 measures, and what it is waiting for).
* Controls: a grand total moved by a cent, a taxable base moved by ten euros, and a VAT of `3.00` on a base of `10.47` are all three refused.
* `phpcs` clean.

## Two things to know

**`ImportVatCalculationModeTest` is touched.** Its stub aligned against a header carrying `taxTotalAmount` and `grandTotalAmount` and nothing else — a shape no received document has, BG-23 being mandatory in EN 16931 and the place the VAT per rate is read from. With the breakdown added, the six tests pass here, and the two that asserted a message naming `Mode 1` / `Mode 2` now assert the total the document announces. Glad to split that off or drop it if you would rather the file stayed untouched.

**#904 hooks a check into the replay loop this branch removes**, so the two conflict for a real reason. Whichever lands first, I rebase the other and put the BT-113 check on the paths that remain.


daGrumpf-bxp:
Moved `carryAnnouncedVatOntoLines()` to the end of the group of import-side writers (`createHeaderChargeLines`, `buildHeaderChargeLines`, `createHeaderDiscounts`, `getOrCreateDepositDiscount`) and its call up to right after the VAT calculation modes have been tried, instead of both sitting where #904 also lands.

Reason: #904 and this one each merge on `main`, but not one after the other — `git merge-tree` conflicted in `CIIProtocol.class.php` in both orders, and that is not something to leave for the merge. Both orders are clean now. The change is a pure relocation, the added and removed lines being identical to the previous revision, and the bench was rerun on it: the five generated documents rebuild their amounts, a grand total moved by a cent and a taxable base moved by ten euros are both still refused, PHPUnit 256/256 on Dolibarr 18 to 25.

daGrumpf-bxp:
Two things after a closer look at the shape of this, since it does add a fourth rung to a ladder that already had three: totals agree, replay mode 2, replay mode 1, and now carry.

**I tried deleting the replay instead of following it.** If carrying the announced VAT always lands on the document's figures, the two mode replays would be dead weight. It does not hold as it stands: `ImportVatCalculationModeTest` comes back with 2 failures of 6 without them. Worth knowing why, though — the stub that test aligns with carries no BG-23, so the carry never fires there and the test does not actually settle whether one could replace the other. What it does settle is that the replay is deliberate, covered behaviour, so collapsing the two into one is your call and not something to slip into this branch. Say the word and I will measure it properly on real documents.

**The method itself was fat and is now 15 lines shorter.** It read its lines by hand in SQL where `fetch_lines()` already exposes the `tva` column as `total_tva`, and it scanned every rate to decide one sign where each rate carries its own. What remains is forced rather than chosen: the write has to land on the lines, because the comparison reads the VAT per rate from them (`getVatDetails()`), and the header has to be written straight, because `update_price()` would recompute it away.

Rerun on the bench after the rewrite: five generated documents rebuilt, both damaged controls still refused, PHPUnit 256/256 on Dolibarr 18 to 25.

daGrumpf-bxp:
**This now replaces the replay of the two VAT conventions instead of following it**, and the branch is a different shape from the one reviewed above. `alignInvoiceTotalsWithDocument()` goes from four successive attempts to two, 100 lines to 71.

Measured before deciding, on the cases the replay exists for — a vendor rounding on mode 2 read by an instance on mode 1, and the other way round. Three lines of `1.09 / 7.79 / 1.59` at 20%, a document announcing `2.09` (mode 2) and another announcing `2.10` (mode 1), imported on an instance in each mode:

```
                             header (ht/vat/incl)      VAT per line
replay (main)  mode 1  m2    10.47 / 2.09 / 12.56      0.22 1.56 0.31
carry only     mode 1  m2    10.47 / 2.09 / 12.56      0.22 1.56 0.31
replay (main)  mode 2  m1    10.47 / 2.10 / 12.57      0.22 1.56 0.32
carry only     mode 2  m1    10.47 / 2.10 / 12.57      0.22 1.56 0.32
```

The same invoice, line VAT for line VAT, in all four combinations — and without recomputing the invoice twice. The reason they agree is that the core, in mode `'1'`, carries the residual difference onto one line exactly as this does.

**Two things it refuses**, so that "carry what the document says" does not become "believe anything":

* a taxable base differing from BT-116 — a line read wrong, not a convention;
* a VAT differing by **more than a cent per line of the rate** — the most a rounding convention can move. A document announcing `3.00` of VAT on a base of `10.47` is still marked and still blocks. That bound was missing from the version reviewed above, which would have written the `3.00` in; the repo's own `testADifferenceThatIsNotARoundingConventionIsReportedAndBlocks` caught it once the stub was made representative.

**`ImportVatCalculationModeTest` is touched**, which I would not do lightly. Its stub aligned against a header carrying `taxTotalAmount` and `grandTotalAmount` and nothing else, a shape no received document has — BG-23 is mandatory in EN 16931, and it is where the VAT per rate is read. With the breakdown added the six tests pass on this branch, and the two that asserted a message naming `Mode 1` / `Mode 2` now assert the total the document announces. Happy to split that into its own commit or drop it if you prefer to keep the file untouched.

One consequence for **#904**: it hooks a check into the replay loop this branch removes, so the two now conflict for a real reason and not a cosmetic one. Whichever lands first, I rebase the other and put the BT-113 check on the paths that remain.

PR #911, opened 2026-09-10
FIX [einvoicing] A line reference to a document we do not hold no longer aborts the batch

daGrumpf-bxp:
#897 was closed by its author this afternoon without landing, and #855 has not moved since the review it received on 09/09. The defect they both aimed at is still on `main`, so here it is on its own, with what a bench says about it.

## The defect

A received line carries its referenced documents in `ram:AdditionalReferencedDocument` (BT-128). The import looks each one up as a supplier invoice, and when it finds none it returns:

```php
return ['res' => -1, 'message' => 'Document "…" linked to line … was not found in Dolibarr…'];
```

No `postponeflow`. `syncFlows()` reads exactly that flag: without it, `$error++` then `break`. **The whole batch stops**, and every flow behind it stays pending for good — the synchronization cursor being the `updatedat` of the documents actually stored, a flow never stored comes back into the window at every run and stops it again.

What ends up in that slot in the field is rarely a document: a contract number, an order number, and on telecom or utility invoices the object the line bills — a phone number, a meter — which EN 16931 calls an *invoiced object identifier* and qualifies with BT-128-1 (`ram:ReferenceTypeCode`).

## What lands

The lookup runs first, exactly as today. Only the not-found branch changes:

1. **BT-128-1 is present** — the issuer is saying, about this very reference, that it is written in an identification scheme of its own: what the line bills, and not a document. Nothing to look for, and nothing to report, since a telecom issuer puts one on every line;
2. no qualifier, and the document **declares an amount already paid** (BT-113) while pointing at an invoice we do not hold → that is the missing deposit, and stepping over it would import an invoice short of its deduction. The flow is **postponed**: nothing is stored, `syncFlow()` rolls its transaction back, and the next run takes the document again, with the business action that asks for the missing invoice to be recorded. That is how BG-3 is already handled at document level (`LINKED_INVOICE_NOT_FOUND`);
3. no qualifier and nothing declared paid → stepped over **with a message** in the import report: a contract number, an order number.

The qualifier is read **before** BT-113, and that order is the whole point. Tested the other way round first, and it cost the case this started from: a telecom invoice whose lines carry a phone number under `AWV` **and** which declares any amount already paid — a credit, an advance, nothing exotic — had its flow postponed, so the batch went on but that invoice never came in. Measured:

```
telecom line 0639981234 + AWV, TotalPrepaidAmount 120.00
    BT-113 read first      flow POSTPONED, retried for ever
    qualifier read first   imported
```

A missing reference that carries a qualifier is therefore skipped even when money is declared paid. It costs nothing: #904 confronts BT-113 with the deposits actually attached and marks the invoice when they disagree, so a deposit that really is missing is still caught, wherever its reference was written.

## Measured

Dolibarr 22.0.5, real import through `createSupplierInvoiceFromSource()` in a transaction that is rolled back, a supplier deposit invoice of 100.00 excl. / 120.00 incl. VAT recorded beforehand. Six documents, identical but for the line reference:

| line reference | `main` | this branch |
|---|---|---|
| contract number, nothing declared paid | **batch aborted** | imported, reported |
| `0639981234` + BT-128-1 `AWV`, nothing declared paid | **batch aborted** | imported, silent |
| the deposit, no BT-128-1 | deducted | deducted |
| the deposit + BT-128-1 `AWV` | deducted | deducted |
| the deposit + BT-128-1 `AFL` | deducted | deducted |
| missing reference, `TotalPrepaidAmount 120.00` | **batch aborted** | **flow postponed** |

The three deposit rows are the point of doing the lookup first. The reviewer of #855 asked on 09/09 whether `AFL` does not designate a document, and it does: UNTDID 1153 holds 806 codes, most of them documentary — `IV`, `ON`, `CT`, `DQ`, `AFL`. Deciding from the presence of that code **before** looking up costs the deduction: measured on this bench, a deposit referenced with `AWV` or `AFL` gets no discount at all. The author of #855 reached the same conclusion himself in his answer that day. So BT-128-1 is used here only to decide how loudly a *missing* reference is reported, never whether to look it up.

PHPUnit of the module, Dolibarr 18 to 25: **256 runs, 256 OK**. `phpcs` clean. `FacturXProtocol extends CIIProtocol`, so both protocols are covered.

## An entry carrying no identifier

Taken from #897 after it closed, and it earns its place: `findIdByRef()` answers `0` for an empty reference exactly as it does for one it cannot find. Without the guard, an `<ram:IssuerAssignedID/>` left empty on a document declaring an amount already paid **postpones the flow for good** — nothing will ever make an empty reference resolvable. Measured on this branch:

```
empty reference + TotalPrepaidAmount 120.00
    without the guard   flow POSTPONED, retried next run, for ever
    with it             imported
```

The seven other cases are unchanged, PHPUnit stays at 256/256 on Dolibarr 18 to 25.

## Credit

The case for not aborting on a reference that is not ours is **#897**'s, from ATM Consulting, and the case for reading BT-128-1 is **#855**'s, from Eoxia. Both were closed by their authors in favour of this branch, and what each contributed is here: the empty-reference guard is #897's code, and the rule that reads BT-128-1 only once the lookup has come back empty is the one the author of #855 described in his review answer of 09/09. Neither is reused as it stood — one aborted on an unqualified reference, the other lost the deposit deduction — but the two readings are theirs.

No test file is added, per the Critical Rules of `.agents/AGENTS.md`. Glad to push the fixtures and the test class if you want them.


daGrumpf-bxp:
Fresh evidence that this is still live on `main`, found while benching something else.

A real round trip through the access point, dd22 on Dolibarr 22.0.5 / PHP 8.3, `main` at `41622d8c`. One instance issues a final invoice whose deposit is deducted **as a line**, so the deposit reference travels at line level. The receiving instance does not hold that deposit. The synchronization answers:

```
ERROR_SYNCFLOW - Failed to synchronize flow i_525339: Failed to create supplier invoice from
E-invoice document for flowId: i_525339. Document "TRI-AC-2609-0095" linked to line 1 was not
found in Dolibarr. Please verify why this document is missing (deleted, not imported, or not
provided by the supplier). To resolve this issue, you must manually create the invoice using the
supplier invoice reference "TRI-AC-2609-0095".
Aborting synchronization due to errors.
```

`res: -1`, `syncedFlows: 25`, and the batch stops there — every flow behind it stays unread. The queue only moved again after the missing deposit was recorded by hand on the receiving side; the very next run drained it.

Two things worth noting for this PR:

* **The document-level loop no longer does this.** The three `return ['res' => -1` of the line loop (`CIIProtocol.class.php:1299`, `:1302`, `:1337`) are now the remaining ones in the import that stop a batch on a reference the buyer does not hold — which is what this PR is about.
* **The trigger is ordinary**, not exotic: this is a Dolibarr-emitted final invoice after deposit, the standard French shape. The deposit reference goes to line level precisely because the deduction is a line.

Also measured on the same run, since it bears on the rule this PR applies at line level: when the deposit is deducted **as a line**, the emitted document carries `TotalPrepaidAmount` = **0.00** — the deduction is in the lines, so nothing is "already paid". When it is imputed **at document level** instead, the same chain emits BT-113 = 120.00 and the reference goes to BG-3. So the two shapes are distinguishable by BT-113 at both levels, not only at line level.

eldy:
The workflow we want is to avoid as much as possible the "postponing".
If we have an error that need a manual action to fix, the process must stop with advice on how to fix it. If we run again, we must stop with eactly the same error as same point.
It seems this PR does not match this rule ? The invoice can't be imported and we bypass so may import the next one. And because the process will start after the last imported, the postponed flow will just be forgotten . This may have no bad effect if it is a cycle life status message but for an import of an invoice, we should never continue if we reach an import error.

daGrumpf-bxp:
Thanks — two points, one of which is the whole reason this PR exists.

**Documents do not arrive in order.** If we stop the entire import on a missing reference, the later invoice that *supplies* that missing document can never be imported: it sits behind the blocked flow, and the synchronization stops before reaching it. Measured on 30/08: asking the supplier to send the missing document unblocks nothing at all. The only way out is to record the missing invoice **by hand** — so "stop with advice on how to fix it" is not a workflow that converges here, it is a dead end that also freezes every other supplier's flows in the meantime. That is exactly what `main` does today on an ordinary Dolibarr final-invoice-after-deposit (`TRI-AC-2609-0095`, `syncedFlows: 25`, comment above).

**A postponed flow is not forgotten.** `syncFlow()` wraps each flow in its own transaction, so a postponed flow has stored nothing, and the synchronization cursor is the `updatedat` of the documents actually stored. It therefore comes back into the window at **every** run and stops at the same point with the same message — which is literally the rule you state ("if we run again, we must stop with exactly the same error at same point"). That is the mechanism already on `main` for the document-level case (#540, #676) and for an unreadable payload (#751/#764, which you merged), and it is the one case you declared acceptable in #718.

So this PR postpones only in the case you already accepted: a reference we do not hold **while an amount is declared paid** (BT-113) — the missing deposit, where importing would book an invoice short of its deduction. The two other branches are not import errors being bypassed: one is a reference the issuer explicitly qualifies as *not* a document (BT-128-1 — a phone number, a meter), the other is a contract or order number, imported with a message in the report. Nothing is ever stored half-linked.

Happy to narrow it further if you prefer — e.g. keep only the postpone branch and leave the unqualified case aborting — but the out-of-order point above stands whichever way we go.

biedseb1:
Hello,
Following the information provided by Esalink:
On a line with the status code "INFORMATION," self-reference is expected under the French profile for informational lines. Your connector should therefore interpret it as "no prior invoice."
This should not be blocking, as this reference can correspond to different elements depending on the case. For example, it sometimes corresponds to the transport order number: this is notably the case with FedEx, which transmits this information in this form. Here too, this should not cause a blockage.
Please don't hesitate to get back to me if you need any clarification.
Best regards,

daGrumpf-bxp:
Thank you — this is directly usable, and it points at a discriminant the module already reads.

`parseInvoiceLines()` already picks up **BT-X-7 / BT-X-8** (`ram:AssociatedDocumentLineDocument/ram:LineStatusCode` and `ram:LineStatusReasonCode`, `CIIProtocol.class.php:265-266`), and `main` already acts on them elsewhere: `isDetailLine()` uses `LineStatusReasonCode` to tell a regular item from a line that carries no amount (BR-FREXT-BR-22 / BR-FREXT-CO-10). So "an `INFORMATION` line means *no prior invoice*" is a rule this import can apply as it stands, and it would be a much better discriminant than the ones this PR currently uses.

Before I rewrite the PR around it, I would rather measure it on real documents than infer it. **Could you send the XML you mention — the Esalink ones, and the FedEx transport-order one?** Factur-X PDFs are fine too, anonymised as much as you need: only the structure around the reference matters, the party names and amounts can be replaced.

What I need to read on them, and cannot guess:

1. **The guideline URN the document declares** (`ram:GuidelineSpecifiedDocumentContextParameter/ram:ID`). BT-X-8 exists from **EXTENDED** upwards only; under plain EN16931, `ram:LineStatusReasonCode` is forbidden, so `INFORMATION` would simply not be readable there. If these documents are EN16931 rather than EXTENDED / EXTENDED-CTC-FR, the rule holds but the discriminant is not in the file, and I need another one for that case.
2. **Where exactly the reference sits on the line**, and with what qualification: `ram:SpecifiedLineTradeSettlement/ram:AdditionalReferencedDocument` — its `ram:IssuerAssignedID`, and whether it carries a `ram:TypeCode` and/or a `ram:ReferenceTypeCode`. (Related: on #884 we established that `ram:TypeCode` on that element is unusable as a discriminant, because EN16931 forbids it; `ReferenceTypeCode` is what this PR keys on today.)
3. **Whether the self-reference really equals the invoice's own BT-1**, and whether it appears only on `INFORMATION` lines or on `DETAIL` lines as well. If it is exactly BT-1, that is a check that needs no profile at all and works everywhere.
4. **What the document declares in BT-113** (`ram:TotalPrepaidAmount`) in these cases — 0.00 or not. This PR currently postpones only when an amount is declared already paid, on the grounds that importing then books an invoice short of its deduction; if your samples carry a self-reference *and* a prepaid amount, that branch has to change too.

Once I have them I will add them to the bench (the module's own harness replays a received document against Dolibarr 18 to 25) and report what each one produces on `main` and on this branch, before and after the change.

To be explicit about the outcome: what you describe makes the **unqualified branch of this PR too blunt**. Today it postpones an unqualified reference whenever BT-113 > 0, and steps over it otherwise. If an `INFORMATION` line is a self-reference by construction under the French profile, that line should be skipped outright — no lookup, no message, no postponement — and the postponement kept only for what is genuinely a missing deposit. I would rather write that against your files than against my reading of the spec.

biedseb1:
Here is the file, however it has been anonymized
[facture_anonymisee.xml](https://github.com/user-attachments/files/32186942/facture_anonymisee.xml)

daGrumpf-bxp:
Thank you — the file answers more than I asked. I benched it before replying; here is what it contains, and what it changes.

**What the document is.** UBL 2.1, `CustomizationID urn:cen.eu:en16931:2017#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr`, `ProfileID S1`. An express-freight invoice of four lines, 377.67 excl. / 404.03 incl. VAT, two lines at 20% and two exempt (VATEX-EU-151).

### 1. It carries no line status code

There is no `cbc:LineStatusCode` anywhere in the file. So the rule Esalink states — an `INFORMATION` line means *no prior invoice* — may well be right, but this document does not carry the discriminant, and I will not key the import on a field I cannot see in a real file. If one of your other samples does carry it, send it and I will measure that too; otherwise this PR should not be rewritten around it.

### 2. The line reference is already qualified

Each of the four lines carries:

```xml
<cac:DocumentReference>
   <cbc:ID schemeID="AAM">000000000001</cbc:ID>
   <cbc:DocumentTypeCode>130</cbc:DocumentTypeCode>
</cac:DocumentReference>
```

That is BT-128 with **BT-128-1 = `AAM`** (waybill number) — the tracking number of the shipment the line bills. In the UBL binding `DocumentTypeCode` 130 is precisely what marks the element as an *invoiced object identifier* rather than a document reference; the CII equivalent is `ram:AdditionalReferencedDocument` carrying `ram:ReferenceTypeCode`. This is exactly the **first branch of this PR**: a reference the issuer qualifies is not a document, so it is skipped without a lookup message.

### 3. The self-reference is at document level, not line level

`cac:BillingReference/cac:InvoiceDocumentReference/cbc:ID` = `900000001` = the invoice's own `cbc:ID` (BT-1). That is BT-25, one storey above what this PR touches, and `main` already steps over it since #927 — you can see its message in the import report below.

### 4. BT-113 is absent

No `cbc:PrepaidAmount` in `cac:LegalMonetaryTotal`. The postpone branch of this PR is therefore never reached by a document of this shape.

### One caveat about this file itself

The module has no UBL reader — `ProtocolManager::getProtocol('UBL')` returns `null` — so this exact file is not importable today: at reception it is postponed with `ERROR_FLOW_NOT_SUPPORTED_PROTOCOL` and the action asking to set the conversion format on the access point account. A flow of this shape reaches the code this PR changes only once the access point delivers it *Converted* into CII / Factur-X.

So I transposed it into CII line for line to bench it: same four lines, same amounts, same four references (`ram:ReferenceTypeCode AAM`, `ram:TypeCode 130`), same document-level self-reference, same totals, `TotalPrepaidAmount 0.00`.

### Measured

Real import through `createSupplierInvoiceFromSource()`, inside a transaction rolled back afterwards:

| document | `main` | this branch |
|---|---|---|
| as sent — 4 qualified references | **batch aborted** on line 1 | imported, 404.03 incl. VAT |
| same, `ReferenceTypeCode` removed | **batch aborted** on line 1 | imported, one message per line |

Dolibarr 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1 and 25.0.0-alpha (PHP 7.4 to 8.5): **8 aborts on `main`, 8 imports on this branch**. On `main` the message is the one you know:

```
Document "000000000001" linked to line 1 was not found in Dolibarr. Please verify why this
document is missing (deleted, not imported, or not provided by the supplier). …
```

The import report of the qualified document names the references nowhere:

```
Thirdparty … updated successfully.
Document FDX-Q-900000001 names itself as the invoice it follows; the reference was ignored.
Product found by prodsellerid          (x4)
Supplier Invoice created or updated with ID: 20774
Einvoice file saved as attachment
```

while the same document with the qualifier removed produces one line per reference:

```
Document "000000000001" referenced by line 1 matches no supplier invoice in Dolibarr and was
stepped over: the invoice is imported without it.
…  x4, one per line
```

That settles something the body of this PR asserted without measuring: on a four-line document the qualifier is the difference between **zero messages and four** — on a 200-line transport or telecom invoice, between zero and 200. All my earlier test documents had a single line, so this is the first time the claim is actually proved.

### Conclusion

Your file does not ask for a rewrite — it is the first **real** document confirming the two rules this PR already applies, on a shape I had never benched: several lines each carrying their own reference. Thank you for it.

Still welcome, if you have them: a sample where the reference sits on an `INFORMATION` line, and one where this same shape arrives as CII rather than UBL.

daGrumpf-bxp:
Heads-up: this one has a hole we had not foreseen, and it is reachable with a **valid** document. The fix is in #947, which already answers the same question at the other site.

## What happens

The line level rule postpones the flow when a reference is unqualified and `totalPrepaidAmount > 0`. That reads BT-113 as *"a deposit is missing"*. It is not always. **`BR-FR-CO-09`** (FNFE `France_RFE` V1.4, `CII/EN16931/schematron/BR-FR-Flux2-Schematron-CII.sch` line 1155, `flag="fatal"`) reads **BT-23 in B2, S2 or M2** as *"invoice already paid"*: BT-113 equals BT-112 and BT-115 is zero. The vendor cashed it in as he issued it — a card, a direct debit. There is no deposit, and there never will be.

Such a document, carrying an unqualified BT-128 on a line — a tracking number, the shape a carrier or a utility issues — is **postponed at every run, for ever**.

## Measured

Two documents built from the BG-22 bench of #904, identical but for BT-23, each with one line carrying `AdditionalReferencedDocument` / `IssuerAssignedID` + `TypeCode 130` and **no `ReferenceTypeCode`**. Both are legal:

```
p911-b2-lineref.xml    VALID    urn:cen.eu:en16931:2017      (fnfeval, FNFE France_RFE V1.4)
p911-b1-lineref.xml    VALID    urn:cen.eu:en16931:2017
```

| document | BT-23 | BT-113 | BT-112 | BT-115 | `main` (9efd1448) |
|---|---|---|---|---|---|
| already paid | **B2** | 540.28 | 540.28 | 0.00 | **flow POSTPONED**, `LINKED_INVOICE_NOT_FOUND` |
| after a deposit | B1 | 120.00 | 540.28 | 420.28 | flow POSTPONED — correct, the deposit is real |

Dolibarr 22.0.5 / PHP 8.3, real import through `createSupplierInvoiceFromSource()` in a transaction that is rolled back.

The first row is the defect: the invoice waits for `TRACK-…`, which nothing will ever record, because the document says precisely that nothing is owed.

## How it got through — the chronology is the answer

| | |
|---|---|
| **10/09 17:21** | the line level rule is written here |
| 12/09 | #904 lands, encoding the same premise at document level |
| **14/09 16:07** | the maintainer disables #904 on a real counter-example and leaves a TODO |
| **14/09 afternoon** | `BR-FR-CO-09` and BT-23 are found, answering that TODO |
| **14/09 18:37** | this PR is merged |

The premise *"BT-113 > 0 means a deposit is missing"* is ours: #884 → #931 → #904 → this one, four PRs in four days. Nobody questioned it because **every bench document we built carries `B1`** — the six of #904 included. A wrong model stays invisible when every specimen comes out of the same mould.

What makes it sting: the body above *describes* this exact failure mode (*"postponed, retried for ever"*) and guards two cases against it — the qualified reference, and the empty one. The remaining case, unqualified and non-empty on an already-paid document, was never built. And the real document that motivated this PR carries no BT-113 at all, so it never exercised the postpone branch: only a synthetic `B1` did, which is the case where the rule is right.

## The fix

#947 now carries one rule read by both sites — `depositAnnouncedByDocument()`: an amount announced under B2, S2 or M2 is not a deposit, so it is neither waited for at line level nor confronted at document level. Nothing changes for B4/S4/M4, which `BR-FR-CO-08` names as the final invoices after a deposit, nor for any other framework.

A `B2` specimen goes into the bench alongside the `B1` ones, so the mould is no longer single.

PR #914, opened 2026-09-10
stancer: let a customer pay again after a card payment was refused

hello-lemon:
Found in production this morning: a customer whose card had just been declined could no longer pay that order at all, and every further click on "Pay by card" answered with a blank page. Four separate things had to line up for that, and they are fixed here. Everything below was reproduced and verified on a live instance (Dolibarr 22, module 2.0.23).

### 1. No retry after a refusal (the blocking one)

Stancer keeps a `unique_id` reserved for good, **even when the payment carrying it was refused**. The module rebuilt the same one, so Stancer answered:

```
HTTP 409 {"error":"Payment already exists, duplicate unique_id","unique_id":"CUS=1986.ORD=1276"}
```

and no further attempt could ever be created for that object. `canBeReused()` only returns true for `STATUS_DRAFT`, so a refused payment fell into the branch that leaves `$mustNewUUID` false and nothing produced a fresh attempt id.

A new `hasFinallyFailed()` covers the final failure states (error, expired, failed, refused, canceled) and starts a new attempt. The suffix now comes from `stancerNextFreeTag()`, which looks for one that is actually free and fits the `varchar(36)` column, instead of `substr($sp->getNextNumRef(), -2)`.

### 2. The blank page

On that 409 the error path called:

```php
$urlPayment = getOnlinePaymentUrl(0, $object->element, (string) $object->ref);
header("Location: " . $urlPayment);
```

`$object->element` is `commande` or `facture`; `getOnlinePaymentUrl()` only answers to `order`, `invoice`, `member`, `membersubscription`, `contractline` and `free`. It returned an empty string, and the empty `Location:` header showed the customer a blank page. A proposal, which has no online payment page at all, hit the same wall.

The element is now mapped to a payment type, and when there is no page to return to the reason is printed rather than redirecting to nowhere.

The message that went with it was:

```php
$message = "Please try with an other payment provider like Stripe";
```

A hardcoded English string recommending another payment provider to the customer, next to an `ErrorStancer` key that has no translation and therefore printed its own name. Replaced by a translated sentence saying that nothing was charged and that the payment can be tried again.

### 3. An email or a mobile can be taken from the contacts

A Stancer customer needs an email or an international mobile. The check read the thirdparty alone, and its landline field alone (`substr($societe->phone, 0, 1) != '+'`), while the contacts of that same thirdparty usually carry both. On the instance this was found on, **1011 of the 1099 thirdparties that could not pay online had a contact with an email or a mobile**, and 17 more had a French mobile stored as `06...` rather than `+336...`.

`stancerResolvePayerContact()` now walks the thirdparty, then the contacts of the paid object (billing first), then the contacts of the thirdparty, and `stancerNormalizePhone()` puts a national number into international form. The same helper is used by the guard that shows the button and by `stancerAddCustomerIfNeeded()`, which had drifted apart: the guard read `phone` only while the customer builder preferred `phone_mobile`, so the button could be hidden for a payment that would have gone through.

**Nothing is written.** The values feed the Stancer customer for that payment only; copying an address onto a record stays a human decision.

### 4. What the customer reads when it fails

- the reference was printed from `$object->ref`, but `$object` is only loaded on the paths that could identify the document, so it showed empty. It now falls back to the decoded link and to the payment record.
- the summary passed the Stancer word to `getLabelStatus()`, which expects the local integer status: `"refused"` resolved to `0` and a declined payment was labelled **"Draft"**. Statuses are translated instead.
- a refusal now says whether the bank declined it (with its response code) or whether strong authentication could not run at all — `auth.status: unavailable`, where the bank is never asked and no response code exists. Those are two different problems for the customer: try again, versus use another card.
- it now states that nothing was charged.

### Two additions

- **"Send the payment link"** on orders and invoices: mails the permanent payment page to the resolved contact and records it in the agenda, with the source of the address. Goes through `formconfirm()`, so the mail leaves on a POST carrying a CSRF token, never on a GET a link prefetch could fire.
- **`STANCER_CB_DISABLE_3DS`**, off by default and exposed in the card setup: some corporate cards never manage to authenticate (`auth.status: unavailable` twice in a row on the order above), and the payment is refused before the bank is asked for anything. The tooltip states plainly that turning authentication off moves the liability for a fraudulent chargeback to the merchant.

### Verified

- the click that returned an empty `Location:` now redirects to a real Stancer payment page;
- an order whose thirdparty has no email and a national landline shows the payment button again;
- the error screen shows the reference, "Payment declined" and the authentication explanation;
- `php -l` clean on every touched file, all new keys present in `fr_FR` and `en_US`.

Happy to split this into separate pull requests if you would rather review them one at a time.


rycks:
hello @hello-lemon 
yes could you please split that PR in many little PR ?
thanks for your contribs
Éric

hello-lemon:
yep. you can close this one.

hello-lemon:
Thanks @rycks, done: split into seven pull requests, one subject each, that merge cleanly in any order.

- #1063 retry after a refusal, and the return URL that names the new attempt (the blocking one)
- #1064 blank page when a payment cannot start
- #1065 what the return page tells the customer
- #1066 option to send card payments without 3-D Secure (off by default)
- #1067 missing `require_once` for FormMail
- #1068 payer email or mobile taken from the contacts
- #1069 "Send the payment link" button (depends on #1067 and #1068)

Together they are exactly the code of this PR, so I am closing it.

PR #922, opened 2026-09-10
FIX: [einvoicing] Name the missing field instead of shipping a placeholder in the document

daGrumpf-bxp:
## The defect

An incomplete third party record did not stop the generation — it was papered over with literal text:

```php
'sellercountry' => $mysoc->country_code ?? 'COUNTRY NOT SET',
'buyercountry'  => $buyerCountryCode    ?: 'COUNTRY',
'buyername'     => $buyerName           ?: 'CUSTOMER',
'buyerpostcode' => $buyerZip            ?: 'ZIP',
```

`COUNTRY` in a field of two characters is refused as fatal:

```
profile FACTUR-X_EN16931.xslt : 100 check(s), 1 failure(s)
    [FX-SCH-A-000568] Value of 'ram:CountryID' is not allowed.
```

The two operators were not even consistent: `??` only fires on `null`, so an empty seller country went
through untouched and produced `<ram:CountryID/>` — a different failure, on the same missing field.

The buyer registration identifier had no length control either, where the seller one has had one for a
while. `G1.63` makes the SIREN of **both** parties mandatory, `G1.89` makes it nine digits, and
`BR-FR-32` refuses anything else under scheme `0002`, as fatal. `idprof()` truncates a SIRET to nine
characters, so the case that gets through is a SIRET typed into the SIREN field — which the import of
a received document can itself produce.

## The change

Three checks in the control block that already refuses a bad professional identifier, each naming the
party and the field so the operator knows which record to open:

* country of both parties (BT-40, BT-55), mandatory under `BR-09` and `BR-11`;
* names of both parties (BT-27, BT-44), mandatory under `BR-06` and `BR-07`;
* format of the registration identifier of both parties: `BR-FR-32` asserts `^\d{9}$` on every party
  under scheme `0002` and `G1.89` says the SIREN is nine digits, so both tests measure the format and
  not a length. The core never validates `idprof1` when a third party is saved, so `12345678A` is
  storable and a length test lets it through into a document the access point refuses as fatal.

The country and name tests are one per party, each naming the record to open, and they compare a
trimmed string rather than calling `empty()`, which also refused a third party named `0`.

And the placeholders are gone. The post code (BT-38, BT-53) and the city (BT-37, BT-52) are optional
terms, so an absent one is now written as nothing at all instead of an empty element that
`PEPPOL-EN16931-R008` reports. Measured: a document without either is `VALID` on the three stages.

## Measured

On the eight cores of the bench (18.0.10 → 25.0.0-alpha, each under the PHP version the CI matrix
gives it, 7.4 → 8.5), four cases:

| | result |
|---|---|
| nominal | generated |
| seller country empty | refused — `BADADDRESS: The country of your company is empty…` |
| buyer SIREN of 14 digits | refused — `BADTHIRDPARTYPROFID: … is not made of exactly 9 digits.` |
| SIREN `12345678A`, either party | refused — same message; on the previous head it reached the document |
| buyer post code and city empty | generated, no placeholder and no empty element |
| buyer or seller name empty | refused — one message per party |
| third party named `0` | generated |

8/8 cores as expected, each under the PHP version its CI matrix gives it (7.4.33 to 8.5.10). The
fifteen documents collected from the no-post-code-no-city case were put through the three stages of
the FNFE chain (XSD CII D22B, profile XSLT, BR-FR-Flux2, v1.4.0.04): 15/15 `VALID`. PHPUnit 32/32 files on each of the eight (256/256), and
`regenerate_einvoicing_fixtures.php` reports the five reference documents `unchanged`.

## Note

The messages are plain English strings, like the other `BAD*` of that block which the caller already
prefixes with the translated `ErrorGeneratingXML`.


daGrumpf-bxp:
Heads-up on the merge order: this one and #919 both touch the address block of `buildParty()`, and whichever lands second will conflict.

The two changes are independent and the resolution is to keep both — #919 routes the escapes through `einvoicingXmlText()`, this one makes the post code and the city conditional:

```php
if (!empty($data[$prefix . 'postcode'])) {
    $addr->appendChild($doc->createElement('ram:PostcodeCode', einvoicingXmlText((string) $data[$prefix . 'postcode'])));
}
```

Same shape for `ram:CityName`. I have merged the six branches of this batch together and run the full recette on the result — 8 real invoices × 4 cores byte for byte identical, the six CII profiles identical, PHPUnit 32/32 on eight cores, one complete E2E cycle with the statuses back — so the combination holds. Happy to rebase this branch on whichever of the two you take first.

daGrumpf-bxp:
Rebased on `main` now that #926 has landed — my earlier note about the merge order is moot, the conflict is resolved here.

The one place the two met is `'buyername'` / `'buyerids'`: #926 decides the identifier, this branch drops the `'CUSTOMER'` placeholder because the buyer name is now checked before anything is built.

```php
'buyername'  => $buyerName,
'buyerids'   => (empty($buyerGlobalIds) ? '' : $idprof),
```

Both intentions are kept, and the `ram:ID` guard #926 added is untouched.

Recette replayed on the rebased branch, on the eight cores of the bench (18.0.10 → 25.0.0-alpha, PHP 7.4 → 8.5): the four cases as expected on 8/8, PHPUnit 32/32 files per core, and the five reference documents `unchanged`.

daGrumpf-bxp:
I have removed the `PR to fix (CI or conflict error)` label: the conflict it reported was real, and it is gone since the second rebase. The chronology, in UTC:

| time | event |
|---|---|
| 17:48 | first rebase, on a `main` that went up to #923 |
| 18:02 | #919 merged — the address block of `buildParty()`, the conflict announced in my comment above |
| 18:03 | label posted, 90 s later — correct at that instant |
| 19:07 | second rebase, conflict resolved |
| now | head `b946ba52` is a single commit on top of `main` (41622d8c, #919 included), CI green |

einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -1067,15 +1067,15 @@
...
 
-	'buyerlineone'              => $buyerAddressLines[0] !== '' ? $buyerAddressLines[0] : 'ADDRESS',
+	'buyerlineone'              => $buyerAddressLines[0],
 	'buyerlinetwo'              => $buyerAddressLines[1],
 	'buyerlinethree'            => $buyerAddressLines[2],
-	'buyerpostcode'             => $buyerZip         ?: 'ZIP',
-	'buyercity'                 => $buyerTown        ?: 'TOWN',
-	'buyercountry'              => $buyerCountryCode ?: 'COUNTRY',
+	'buyerpostcode'             => $buyerZip,
+	'buyercity'                 => $buyerTown,
```
eldy, line 1070: For mysoc->country_code, we can force to set it, but not for customer side.
What will happen if we generate an invoice for a thirdparty with zip or town unknown ?
If xml still valid with no value ?

daGrumpf-bxp: Yes, the XML stays valid with no value — as long as the element is **absent**, which is what this patch does. An **empty** element is the one that gets reported, and that is exactly what the old `??` produced.

**The norm.** In annex A of XP Z12-012 (V1.4), the post code and the city are `0..1` in all three cardinality columns — EN 16931, CIUS FR and EXTENDED-CTC-FR — for both parties:

| BT | EN16931 | CIUS FR | EXTENDED FR | attached rule |
|---|---|---|---|---|
| BT-37 / BT-52 city | 0..1 | 0..1 | 0..1 | BR-FR-CPRO-33 (length <= 400, B2G only) |
| BT-38 / BT-53 post code | 0..1 | 0..1 | 0..1 | none |
| BT-40 / BT-55 country | **1..1** | **1..1** | **1..1** | BR-9 / BR-11, G2.01, G6.07, **G6.08** |

And G6.08 ("systematically mandatory from 01/09/2026") enumerates what it makes mandatory: inside BG-5 and BG-8, the country code and nothing else.

**Measured, on documents the code produced.** Bench run on the eight cores, each under the PHP its CI matrix gives it (18.0.10/7.4.33 -> 25.0.0-alpha/8.5.10), buyer `zip` and `town` emptied in the base inside a rolled back transaction:

```
E buyer no zip no town | PASS | generated, CityName=1 PostcodeCode=1 emptyelements=0 placeholders=0
```

One `CityName` left, the seller's; the buyer's is gone and no empty element takes its place. The fifteen documents collected that way, put through the three stages (XSD CII D22B, profile XSLT, BR-FR-Flux2, FNFE v1.4.0.04): **15/15 VALID**. The same document with the URN switched to `...#conformant#urn.cpro.gouv.fr:1p0:extended-ctc-fr` is VALID too, on `EXTENDED-CTC-FR-CII.xslt`.

**Why the country is treated differently.** The three shapes, measured on the same document:

| written | verdict |
|---|---|
| no `CityName` / `PostcodeCode` at all | VALID |
| `<ram:CityName/>` (empty) | `PEPPOL-EN16931-R008` — document MUST not contain empty elements |
| `<ram:CountryID/>` (empty) | **`BR-09` fatal** + R008 |
| `<ram:CountryID>COUNTRY</ram:CountryID>` | **`FX-SCH-A-000580` fatal** |

So yes: the customer country cannot be forced, and that is why it is now refused with a message naming the third party instead of being papered over — the two placeholders it used to ship are the last two lines of that table. The post code and the city need nothing: they are simply left out.

**Follow-up commit just pushed** (`a9f7231`), from re-reading the block:

* `BR-FR-32` asserts `^\d{9}$`, nine **digits**. Both SIREN tests only measured a length, and the core never validates `idprof1` when a third party is saved, so `12345678A` is storable. On the previous head it went straight through into the document — `<ram:ID schemeID="0002">12345678A</ram:ID>` — and the access point refuses it fatally on `BR-FR-10/BT-30` and `BR-FR-32-LEGALID`. Both tests now use the regex.
* The country and name tests are one per party, each naming the record to open, instead of one message announcing that two terms are mandatory.
* `trim()` instead of `empty()`, which also refused a third party named `0`.

Recette of the follow-up: 80/80 cases on the eight cores (nominal, both SIREN shapes, SIRET in the SIREN field, no zip/town, no country either side, no name either side, name `0`), PHPUnit 33 files x 8 cores = **264/264**, and `regenerate_einvoicing_fixtures.php` reports the five reference documents `unchanged`.

Issue #920, opened 2026-09-10
[einvoicing] BT-29 / BT-46 are emitted under scheme 0225, which the v3.2 specifications reserve to 0231: make the scheme an option


daGrumpf-bxp:
## What the module writes today

For a French seller, `getIEC6523Code()` returns `0225` for the party identifiers, and the value put
under it is `idprof()`, that is the **SIREN**. The same value therefore appears twice on the seller,
under two different schemes:

```xml
<ram:SellerTradeParty>
  <ram:GlobalID schemeID="0225">000000001</ram:GlobalID>          <!-- BT-29 -->
  <ram:SpecifiedLegalOrganization>
    <ram:ID schemeID="0002">000000001</ram:ID>                    <!-- BT-30, correct -->
  </ram:SpecifiedLegalOrganization>
  <ram:URIUniversalCommunication>
    <ram:URIID schemeID="0225">00000000100010</ram:URIID>         <!-- BT-34, correct -->
```

Same thing on `BuyerTradeParty` (BT-46), and it leaks onto `ShipToTradeParty` (BT-71) because
`buildParty()` builds the deliver-to party from the buyer data and the `GlobalID` block is guarded by
the profile only, not by the `$minimal` flag — `CIIProtocol.class.php:2416` and `:2798`.

## Why it is worth looking at

No validator refuses it: `BR-FREXT-CL-10` only checks that the scheme belongs to the ISO 6523 ICD
list, where `0225` is. The five reference documents are `VALID` on the three stages, with FNFE
`v1.4.0.03` and `v1.4.0.04` alike. So this is not a bug report — it is a semantic divergence that will
only surface downstream.

The DGFiP external specifications **v3.2 of 30/04/2026** reserve that element, in annex 1 (*Format
sémantique FE e-invoicing — Flux 1*, v1.2), sheet `FE - Flux 1 - CII`:

| Row | Path | Text |
|---|---|---|
| `BT-29d` | `…/SellerTradeParty/ram:GlobalID` | "Ce champ est utilisé **uniquement** pour renseigner le SIREN d'un assujetti unique" |
| `BT-29d-1` (1..1) | `…/ram:GlobalID/@schemeID` | "Valeur = **0231**. Si ce champ est utilisé, il ne peut comporter qu'une seule valeur : le SIREN de l'assujetti unique." |

with `S1.14` of annex 7 saying the same, and `G1.101` adding that the SIREN declared there must exist
[...]

PR #926, opened 2026-09-10
NEW: [einvoicing] #920 The scheme of the party identifier becomes an option

daGrumpf-bxp:
Closes the proposal of #920.

## Why

The module declares the SIREN in **BT-29** and **BT-46** under scheme `0225`, which is the *electronic
address* scheme of the national directory — the same value it already, and rightly, writes in BT-34 /
BT-49. Annex 1 of the DGFiP v3.2 specification reserves `SellerTradeParty/ram:GlobalID` to the SIREN of
a VAT group under **`0231`** ("ce champ est utilisé uniquement pour renseigner le SIREN d'un assujetti
unique", rule `S1.14`), and annex A of XP Z12-012 gives `0009` for the SIRET.

Nothing refuses the current form: `BR-FREXT-CL-10` only checks that the code belongs to the ISO 6523
ICD list, and `0225` does. The divergence would surface downstream, at a platform extracting the
regulatory data, and not on a bench.

## What this does

`EINVOICING_PARTY_IDENTIFIER_SCHEME` decides the scheme, and **its default is what the module has
always written**: an installation that upgrades changes nothing at all.

| | |
|---|---|
| French company | a list: `0225` (current behaviour, default), `0009` (the SIRET, when the record holds one), *do not declare a party identifier* |
| any other country | a free field for the ISO 6523 ICD code of the register the professional identifier comes from — empty keeps the code the module answers today |

The free field is the other half of the reason for this change: outside France, Belgium and Germany,
`getIEC6523Code()` answers `0060` — a D-U-N-S number — for a value that never is one. A Spanish NIF or
an Italian partita IVA was therefore declared as something it is not, and there was no way to say so.

**The option reaches BT-29 and BT-46 alone.** The legal registration identifier (BT-30, BT-47) stays on
`0002` as the specification requires, and the electronic address (BT-34, BT-49) stays on `0225`, where
`BR-CL-25` accepts nothing outside the CEF EAS list — moving it would break the routing of every
invoice. The bench harness asserts both under every value of the option.

`0231` is deliberately **not** offered. `BR-FR-CO-14` and `BR-FR-CO-15`, both fatal, make the `TXD`
note `MEMBRE_ASSUJETTI_UNIQUE` and the fiscal representative block (BG-11) mandatory as soon as it
appears; offering it alone would offer a document that gets refused. It belongs with the support of
the VAT group case.

## Measured

On the eight cores of the bench (18.0.10 → 25.0.0-alpha, each under the PHP version the CI matrix
gives it, 7.4 → 8.5), the same invoice generated under each value:

```
''     option unset (historical behaviour)
   seller : 0225:000000002       BT-30/47=0002:000000002   BT-34/49=0225:000000002
0225   default of the list
   seller : 0225:000000002       BT-30/47=0002:000000002   BT-34/49=0225:000000002
0009   SIRET
   seller : 0009:00000000200010  BT-30/47=0002:000000002   BT-34/49=0225:000000002
none   no party identifier
   seller : (none)               BT-30/47=0002:000000002   BT-34/49=0225:000000002
```

8/8 cores as expected, BT-30 and BT-34 identical under all four. The four documents are `VALID` on the
three stages of `validate-cii.py` (`--strict`), no empty element is produced in the *none* case, the
five reference documents are regenerated `unchanged`, and PHPUnit is 32/32 files on each core
(256/256). The six new language keys resolve on all eight, and `$mysoc->country_code` is readable
where the setup page reads it.

One thing the bench caught and the linters did not: constants in a **trait** need PHP 8.2, and this
module runs from 7.x — the first version of this change fataled on Dolibarr 18, 19 and 20. The two
values are written inline instead.


daGrumpf-bxp:
Heads-up on the merge order: this one conflicts with #919 and #922 if it lands after them, in two places, and both resolutions keep every intention.

**`buildinvoicelines.inc.php`** — #922 drops the `'CUSTOMER'` placeholder because the buyer name is now checked before anything is built, this one decides the identifier:

```php
'buyername'  => $buyerName,
'buyerids'   => (empty($buyerGlobalIds) ? '' : $idprof),
```

**`CIIProtocol.class.php`** — the `ram:ID` fallback gets the guard from here and the escape helper from #919:

```php
} elseif (!empty($data[$prefix . 'ids'])) {
    $node->appendChild($doc->createElement('ram:ID', einvoicingXmlText((string) $data[$prefix . 'ids'])));
}
```

I have merged the six branches of this batch together and run the full recette on the result: 8 real invoices × 4 cores byte for byte identical before and after, the six CII profiles identical and their md5 the same on the four cores, PHPUnit 32/32 on eight cores, the documents of the four validatable profiles — EN16931, BASICWL, EXTENDED and **EXTENDED-CTC-FR** — all accepted by the three stages, and one complete E2E cycle: new invoice, routed by the access point, received on the other instance, 200/201 back and a 205 sent from the receiving side, all `validation=Ok`.

Happy to rebase in whatever order suits you.

Issue #912, opened 2026-09-10
einvoicing: deposit references are emitted at line level in the BT-128 slot (TypeCode 130)


nicolas-eoxia:
Follow-up promised in the review of #855, so the point is not lost when that PR is closed.

## The problem

At line level, EN 16931 defines exactly one `AdditionalReferencedDocument`: **BT-128, the Invoiced object identifier**, with `TypeCode` fixed to **130** and `ReferenceTypeCode` (BT-128-1) as the optional qualifier of its scheme. BT-128 identifies *what the line bills* — a phone number, a meter, a subscription — and never a document.

This module emits its own deposit references into that slot:

```php
// CIIProtocol::buildLineItem()
if (!empty($line['isDepositLine'])) {
    $refNode = $doc->createElement('ram:AdditionalReferencedDocument');
    $refNode->appendChild($doc->createElement('ram:IssuerAssignedID', ... $line['depositInvoiceRef'] ...));
    $refNode->appendChild($doc->createElement('ram:TypeCode', '130'));
```

An invoice reference in the BT-128 slot is not conformant.

## Why it matters beyond conformance

It is what makes the *reading* side ambiguous, and that ambiguity has already cost work:

- a conformant reader receiving one of our invoices treats the deposit reference as the identifier of the billed object, and drops the link;
- and on our own reading side, `TypeCode == 130` cannot be used as the discriminator it is meant to be — it would stop linking deposits on every invoice this module has already issued. That is precisely the objection raised in #855, and the reason the fix there had to be keyed on something else. See #911 for the rule that ended up holding.

## What a fix involves

Emitting the deposit reference under a code that means "referenced document" rather than "invoiced object" — `TypeCode 916` at line level in the EXTENDED profile, or at document level as a preceding invoice reference (BG-3).

It cannot be a straight swap: invoices already issued by this module carry `130`, so the reading side needs to keep accepting an unqualified `130` as a document reference for as long as those are in circulation. Worth deciding deliberately rather than as a side effect of another PR.

daGrumpf-bxp:
Checked against the FNFE rules themselves (`France_RFE` 1.4.0.04, CII EN16931 and EXTENDED-CTC-FR schematrons), and the diagnosis holds — with one correction to the proposed target, and one nuance that makes the reading side less ambiguous than it looks.

**Where a preceding invoice belongs: BG-3, at document level.** The French rules name it in those words:

```
[...]

PR #928, opened 2026-09-10
Qual: [einvoicing] drop the writing side of XmlPatcher and rename it EmbeddedXmlReader

daGrumpf-bxp:
## What

`XmlPatcher` was written to patch the XML the horstoeko builder produces: switch its guideline URN to EXTENDED-CTC-FR, and inject the BG-3 deposit reference (`AdditionalReferencedDocument` with `TypeCode` 130) on a line. That whole side lost its caller in **6c79eeb8**, *"One XML builder for Factur-X and CII, not two"*, when the Factur-X path stopped patching a foreign document and started building its own.

What is left of the class is a reader — `getLineAdditionalReferencedDocuments()`, called by `FacturXProtocol` on a received document. This removes the rest.

| removed | reached by |
|---|---|
| `patchXmlString()` | nothing |
| `getPatchedXml()` | nothing — its own docblock says `TODO Seemsnot used` |
| `addDepositLineReference()` | nothing |
| `injectDepositLineRef()` | `patchXmlString()` only |
| `patchGuidelineId()` | nothing — **its only call site was already commented out** inside `patchXmlString()` |
| `URN_EXTENDED`, `URN_EXTENDED_CTC_FR` | `patchGuidelineId()` only |
| `$builder`, `$depositRefs` | the methods above |

PHPStan had already reported `patchGuidelineId()` as unused; the baseline entry that hid it goes with the code, along with the one on `patchXmlString()`'s parameter type.

The constructor keeps only the argument the reader uses (`new XmlPatcher($embeddedXml)` instead of `new XmlPatcher(null, $embeddedXml)`), and its single caller follows. The class docblock loses the sentence about receiving the horstoeko builder, which it no longer does.

## What does not change

The reader, its XPath and the shape it returns are untouched. On the same document, before and after:

```
main    : constructor takes 2 parameters, public API = __construct, addDepositLineReference,
          getPatchedXml, patchXmlString, getLineAdditionalReferencedDocuments
          -> [{"IssuerAssignedID":"FA2609-0007","typeCode":"130","issueDate":"2026-09-01"}]

branch  : constructor takes 1 parameter,  public API = __construct,
          getLineAdditionalReferencedDocuments
          -> [{"IssuerAssignedID":"FA2609-0007","typeCode":"130","issueDate":"2026-09-01"}]
```

## Measured

- PHPUnit on the eight cores of the bench (18.0.10 → 25.0.0-alpha, PHP 7.4 → 8.5): **32/32 files green on each, 256/256**;
- PHPStan level 10 on the two files touched: no error;
- QA emission matrix (CII and Factur-X, every invoice shape of the harness) on Dolibarr 23 and 24: 14 generations, 14 files, unchanged;
- all of it replayed after the rename: same results, and the reader answers the same array under its new name.

## Second commit: the class is renamed

`XmlPatcher` patches nothing any more, so it becomes **`EmbeddedXmlReader`** — the name says what is left of it, next to `PdfAttachmentExtractor`, the step that hands it that XML. The file is renamed with it and the two `dol_include_once()` follow. Kept as a separate commit so the removal can be read without the rename.

Whole pull request: **+100 / −318** over five files, 304 lines of class down to 96.

## Merge order — conflicts with #939

#939 (`fix/facturx-import-reuses-cii`) moves the one remaining call site of this class into a method of
its own, while this PR renames the class under it. The two conflict in either order, on
`FacturXProtocol.class.php` only — and resolving it the wrong way is a fatal, not a cosmetic slip,
since this PR deletes `XmlPatcher.class.php`.

Resolution: **take #939's `FacturXProtocol.class.php` whole, then apply this PR's rename to the moved
call site** in `parseReceivedDocumentWithExternalReader()` — `new XmlPatcher(null, $embeddedXml)`
becomes `new EmbeddedXmlReader($embeddedXml)`. The `dol_include_once()` merges by itself. After the
resolution `XmlPatcher` occurs 0 times in the module. Checked in both orders: same tree to the bit,
`php -l` green.

The rebase is mine to do, whichever lands first.


daGrumpf-bxp:
The `build` check is red on a defect of the phpcs job, not on this change: it hands phpcs the PHP files of the pull request, deletions included, and phpcs stops on the path of the file this pull request removes.

```
ERROR: The file "einvoicing/class/utils/XmlPatcher.class.php" does not exist.
Process completed with exit code 2
```

#930 fixes that job (the list is filtered to the files still present). Once it lands I rebase this one and the check goes green — every other check is already green here, phan and both PHPStan included.

einvoicing/class/utils/XmlPatcher.class.php
```diff
@@ -1,304 +0,0 @@
...
-	 * Find the trade line with the given LineID and append an AdditionalReferencedDocument identifying it as a deposit recovery line.
-	 * @param DOMDocument $dom dom
-	 * @param DOMXPath $xpath xpath
-	 * @param string $lineId line id
-	 * @param string $invoiceRef invoice ref
-	 * @param DateTimeInterface $invoiceDate invoice date
-	 *
-	 * @return void  No value return
-	 */
-	private static function injectDepositLineRef(
```
eldy, line 195: The XMLPatcher was introduced to be able to inject such lines into an invoice
(we need and want to have the deposit declared as the line level as recommanded by First option - chapter 3.2.19 Cas n°20 et 21 : Facture d’acompte et facture définitive (ou finale) après acompte
of document XP_Z12-014_CAS_USAGE_Annexe_A_V1.2.pdf).

Are you sure we can remove this ?

daGrumpf-bxp:
Yes — and the check is worth spelling out. `patchXmlString()` lost its only caller in 6c79eeb8 (#652), when the horstoeko builder went; before that it was reachable only under `EINVOICING_USE_EXTERNAL_FACTURX_BUILDER`, which no screen and no install script ever set — so no installation has ever emitted it.

It also does not write what you are pointing at: it injects `SpecifiedLineTradeSettlement/ram:AdditionalReferencedDocument` with `TypeCode 130`, the BT-128 slot (invoiced object identifier) — exactly what #912 reported and what #921 removed from the native builder.

XP Z12-014 §3.2.19 first option asks for **EXT-FR-FE-BG-06**: `SpecifiedLineTradeSettlement/ram:InvoiceReferencedDocument` with `IssuerAssignedID` (EXT-FR-FE-136), `TypeCode` **386** (EXT-FR-FE-137) and its date (EXT-FR-FE-138). A different element, and EXTENDED only — the EN16931 XSD declares no `InvoiceReferencedDocument` in `LineTradeSettlementType`, the EXTENDED one declares both.

That one we do not write, and you are right that we should: `buildinvoicelines.inc.php` states it follows the first method, and it still collects `depositInvoiceRef` / `depositInvoiceDate` — which nothing has written anywhere since #921. We do not read it on import either: the only line level XPath we have is `additionalRefDocs`. So #921 was right about the slot and incomplete about the need, and that is on me.

So this PR removes nothing reachable and nothing we still want to produce. The follow-up is mine to open: emit `ram:InvoiceReferencedDocument` with `TypeCode` 386 on the deposit line under EXTENDED-CTC-FR, and read it back on import. #921 is in `main` and in no tag yet, so nothing shipped has lost it.

eldy:
PR confirmed by @Daoud-mohamed that worked on XMLPatcher.

PR #932, opened 2026-09-10
NEW: [einvoicing] A backlog of the documents waiting to be imported, with the reason and the wait

daGrumpf-bxp:
## The gap

A postponed flow deliberately stores nothing — that is what makes it safe to retry — so **nothing recorded that it was ever seen**.

What the user got instead:

* an aggregate count, *"Total of flows postponed: N"*, that dies with the request;
* `$actions[$res['actioncode']]`, keyed by action code, so ten flows waiting for the same reason collapse into **one** line showing the last one's reference;
* one `LOG_WARNING` in the synchronization log, for whoever reads logs.

Close the page and even that is gone. A document could wait for weeks with no way to know it existed, which reason it was stuck on, or since when — and no way to decide.

## What this adds

`llx_einvoicing_postponed`, one row per waiting flow: the reason the import gave and its action, the supplier and document it is about, the synchronization call that carried it, when it was **first** postponed, and how many runs have met it since. Both providers write it in their postpone branch and drop the row on every other outcome, so a line never outlives its cause.

The synchronization page shows that backlog **on every visit**, not only after a run — a flow that keeps being postponed is invisible otherwise.

### Reaching a flow that never landed

Each line links to its synchronization call (`call_list.php?search_call_id=…`). A postponed flow has no `llx_einvoicing_document` row to select — that is the whole point — so the "Export for support" mass action of #805 cannot see it. The call that fetched the batch **is** logged, and its response carries the document, so that is where support can get at it.

### Indexes

Two, covering every read, and no third because no query needs one:

* `uk_einvoicing_postponed_flow (flow_id, entity)` — the write looks the row up by flow, the drop deletes by flow. Unique, so one row per flow whatever the number of runs, which is also what makes the write an upsert.
* `idx_einvoicing_postponed_entity_date (entity, date_creation)` — the panel lists the entity's backlog oldest first and counts it.

### Portability

The upsert is a read then a write. `ON DUPLICATE KEY` is MySQL only and `ON CONFLICT` PostgreSQL only, and the module runs on both. One synchronization runs at a time, so the read-then-write has nothing to race against, and the unique key guards it anyway.

## Measured

Full round trip through the access point, dd22 on Dolibarr 22.0.5 / PHP 8.3, one instance issuing to another.

| step | result |
|---|---|
| Table creation | through the real path — module deactivation/reactivation replays `llx_einvoicing_postponed.sql`; both indexes in place |
| A genuinely postponed flow | `i_525348`, `LINKED_INVOICE_NOT_FOUND`, batch **carried on**, no transaction left open |
| Row written | `provider=SUPERPDP reason_code=LINKED_INVOICE_NOT_FOUND fk_soc=1 document_ref=TRI-FA-2609-0976 nb_attempts=1 date_creation=19:17:47` |
| Three synchronizations in a row | `nb_attempts=3`, `date_creation` **unchanged**, `date_last_attempt` moved |
| Missing deposit recorded, sync again | invoice imported at 480.00 TTC with its `(DEPOSIT) -120.00` line, **backlog back to 0** |
| Synchronization page | HTTP 200, no error, no warning; the block renders with the flow, the supplier, the reason, the action button and the waiting time |

## Not in scope

The bench run also showed a flow **aborting the whole batch** on a line-level reference (`"Document … linked to line 1 was not found"` → *"Aborting synchronization due to errors"*). That path still answers with a bare `-1` and no `postponeflow`, so it never reaches this backlog. Reported separately.


daGrumpf-bxp:
## The screens

Captured on a bench instance running this branch (Dolibarr 24.0.1), on a genuinely postponed flow: one instance issued a final invoice deducting a deposit at document level, the receiving instance does not hold that deposit, so the flow was postponed three times.

### The backlog on the synchronization page

Documents waiting to be imported

What is waiting, for whom, why, since when, how many runs have met it, and the action that unblocks it. It is shown on every visit, not only after a synchronization.

### Where the link goes

The synchronization call that carried the flow

*"Open the synchronization call that carried it"* lands on the call list filtered on that call — one row, whose processing result reads *"Flow i_525674 postponed, it will be retried on the next synchronization"*. Selecting it offers the mass action, **Export in a zip archive** (#805), which is how support can get at a flow that has no record of its own to select.

## One thing the screenshots show that is worth stating

The interface here is forced to English, but the per-flow message stays French: `business_message` and the action are **stored as rendered text**, in the language of the synchronization run that produced them. In normal use that is the instance's own language, so the page is coherent — it only looks mixed because the URL forces another one. Storing the key and its parameters instead would make the backlog follow the reader, but the postpone path hands the providers an already-rendered message, so that is a change further up the chain, not here.

daGrumpf-bxp:
Follow-up commit on this branch: the link to the synchronization call in the backlog panel was built with `DOL_URL_ROOT.'/custom/einvoicing/call_list.php'`, which assumes the module sits under `htdocs/custom`. It now goes through `dol_buildpath('/einvoicing/call_list.php', 1)`, like the rest of the module (`call_card.php` included).

Checked on the bench, Dolibarr 24.0.1, with a postponed flow carrying a `call_id`: the panel renders and the link is `/custom/einvoicing/call_list.php?search_call_id=…` — unchanged for the usual deployment. With the same instance serving the module through an alternate root directory (`/modules`), the two expressions part ways:

```
-- module under htdocs/custom
  hardcoded     : /custom/einvoicing/call_list.php
  dol_buildpath : /custom/einvoicing/call_list.php

-- same instance, module served from an alternate root
  hardcoded     : /custom/einvoicing/call_list.php   <= 404
  dol_buildpath : /modules/einvoicing/call_list.php
```

PHPStan and phpcs green.

eldy:
There is something i don'tunderstund with the postpone flow (currently).
We do not store it ok.
But what happen if the following flow is received 2 days after and ok. The postponed is not store but we store the next one.
So at next run, the date to start is the one of last ok. So AP return all after the last ok and the postponed is forgot.
Did i miss someting ?

daGrumpf-bxp:
You are right, and it is not a theory: it was already happening on my bench.

### The rule

The start of a synchronization window is `getLastSyncDate()` = `MAX(einvoicing_document.updatedat)` over the documents that **were stored**, minus `EINVOICING_SYNC_MARGIN_TIME_HOURS` (12 by default), and the API filter is `updatedAfter`. So a postponed flow is listed again only while

```
updatedAt(postponed flow) > MAX(updatedat of stored flows) - 12h
```

`MAX` counts every stored flow, the lifecycle messages of the invoices we send included, so the cursor is pushed forward by traffic that has nothing to do with the postponed document: the busier the instance, the sooner it drops out. A flow that *aborts* the batch does not have the problem, precisely because nothing newer is stored behind it - which is why replacing aborts with postponements is what made this reachable. It is the `postponeflow` convention itself, not #933.

### Measured before touching anything

A survey of what the access point still holds against what each instance stored: **four received supplier invoices had already dropped out of the window**, on two instances.

| instance | flow | invoice | postponed | behind MAX |
|---|---|---|---|---|
| dd22 | `i_525378` | TRI-FA-2609-0977, 600.00, 120.00 already paid | 5 runs, 10 -> 13 Sep | 73.9 h |
| dd25 | `i_381740`, `i_381748`, `i_381788` | TRI-FA-2608-0269 / 0271 / 0276, 840.00 each | 6 Sep | 262 h |

dd22's synchronization log carries *"it will be retried on the next synchronization"* five times for that one flow. Running the scheduled path then reported *"Synchronization completed successfully - 10 new flows synchronized"*, zero postponed, **without mentioning it at all**, and moved `MAX` further away. The access point still held it; the window simply no longer asked for it.

The document itself is never lost: `flows/<flowId>` still serves it. All four were recovered with one call each, including one that had been out of reach for 15 days.

### What this branch now does

Both providers start a run by taking the backlog again **by identifier** - `syncFlow($flowId)` fetches `flows/<flowId>` and uses no date window at all - and only then poll the window, which is left exactly as it was. A flow that fails again is kept and never counted as an error: one row nobody can resolve would otherwise abort every run from there on, which is the very failure this replay exists to end. `EINVOICING_FLOWS_SYNC_REPLAY_SIZE` caps it at 50 per run, oldest first.

### Before / after

On dd22 (Dolibarr 22.0.5), with that same real invoice put back in the state it was in: postponed once while the window still reached it, then left 4 days behind the cursor. Margin at its default of 12 h, window start `2026-09-14T07:35:19Z`, flow at `2026-09-10T19:18:54Z`.

| | run | what it said | backlog | invoice |
|---|---|---|---|---|
| **before** | scheduled path, twice | "completed successfully", 0 new, 0 postponed - the flow is not mentioned | `nb_attempts` frozen at 1, "waiting since" frozen | absent |
| **after** | scheduled path, twice | "Total of flows postponed: 1" | `nb_attempts` 2 then 3, last attempt moves | absent, the deposit is genuinely missing |
| **after** | the deposit is recorded, then one run | "Total of waiting flows taken again and imported: 1" | empty | imported, 400.00 HT / **480.00 TTC** - the 120.00 deducted |

`MAX(updatedat)` never moved during any of it: the recovery owes nothing to the window.

PHPUnit 33/33 on that instance with the branch deployed, phpcs, phan and PHPStan green.

This way the backlog is not only a record of what is waiting, it is what makes the promise in its own wording true. I would rather close it here than leave a hole we know about waiting on another PR: #896 has a manual Retry doing the same call, but for the flows it queues, and only when someone clicks.

eldy:
Waiting answer of #896

daGrumpf-bxp:
@eldy Answered on #896: https://github.com/Dolibarr/dolibarr-community-modules/pull/896#issuecomment-5720864056 — short version: #896 sits behind `EINVOICING_ENABLE_MANUAL_ACTION_QUEUE`, off by default, so it cannot replace `postponeflow`: removing the postpone in its favour would leave every default install losing postponed flows for good.

That loss is the very hole you pointed at here on 14 September, and it is measured, not theoretical — four received invoices had already dropped out of the `updatedAfter` window on my bench. #932 closes it by taking the backlog again by `flowId`, with no date window involved, and it stays useful under #896 rather than competing with it.

On the conflict: the branch merges cleanly on today's `main` — GitHub reports no conflict on it now.

eldy:
@daGrumpf-bxp i merged the log table of PR #896.

So for future need of postpone feature, thanks to reuse the same table for your need than the one introduce by pichinov for manual action queue.

Note however that the postpone feature is nto the priority. I added the option INVOICING_ENABLE_POSTPONE_FLOWS (not set by default) so postpone feature is currently not enabled by default for end users.
Priority is to focus as much as possible on solution to progress in the flows by
- manual action each time there is a blocking situation
- find a custom "dolibarr hackor bypass" to solve each point where unlocking manually seems not possible.

For example, when we receive a final invoice that reference a deposit never received or that will be receive into another flow, we don't need to postpone. We can ask user to create it. So we can continue the flow.
And later, if we receive the real deposit, as the reference was already recorded, we can update it (we can retreive it with the supplier ref). So as long as we can have process workign without postponig, priority must be on implementing solution for this.

We should always find FIRST a way to allow a sequential flow (one case per one case). 
We must focus on this. And only when we have a manual "flow by flow" process working on a long term, we can work on a smoother process where more things are done automatically like when using postpone features.

I may not merge PR related to enhancement of postponing, because priority must be to implement first the manual fallbacks for each blocking situation and after only the automatic smoothed solution.

daGrumpf-bxp:
@eldy Done as you asked: the waiting flows are queued in `llx_einvoicing_sync_pending`, the table Pichinov introduced — this pull request now adds **no table and no schema change at all**, and is half the size it was.

Two things had to change in that shared code: `queueFromFlow()` no longer nulls the columns its caller does not bring (there are two writers on one row now, and the second does not know the actions the first computed), and a row is resolved even when `EINVOICING_ENABLE_MANUAL_ACTION_QUEUE` is off — a postponed flow is queued without that option, so its row would have stayed pending for ever.

What is left is only the part #896 does not cover: taking the waiting flows again **by identifier**, before the window is polled. That is your question of 14 Sep, and it was not a theory — the window starts at `MAX(einvoicing_document.updatedat)` over what *was* stored, so every document imported after a postponed one pushes the start past it, and once `EINVOICING_SYNC_MARGIN_TIME_HOURS` is spent the access point is never asked for that flow again. Measured on the bench: four documents out of the window, one for 262 hours. `flows/<flowId>` uses no date window, so it is what makes "retried on the next synchronization" true.

Bench, dd23 on Dolibarr 23.0.4 / PHP 8.4, manual-action option off throughout: flow postponed and queued while the batch carried on (14 flows imported), taken again on the next run (`attempts=2`, first-seen date unchanged), imported once the prerequisite existed, row resolved. Page loads with 0 error and 0 warning. PHPStan clean on 18 and 24, phan clean.

One consequence you should weigh: the replay takes every pending row of the provider, so the rows **your manual-action queue** writes are retried automatically at each run instead of on demand. It costs one `flows/<flowId>` call per waiting row, capped by `EINVOICING_FLOWS_SYNC_REPLAY_SIZE` (50), and a row that fails again is kept, never counted as an error.

If you would rather it only took the rows it wrote itself, say so and I will restrict it — the queue has the reason code to tell them apart.

PR #933, opened 2026-09-10
FIX: [einvoicing] A reference lookup that fails no longer stops the whole synchronization

daGrumpf-bxp:
## The defect

`SupplierInvoiceHelper::findIdByRef()` answers a reference lookup with `-1` on a database error,
`-2` when several supplier invoices carry the reference, and `-3` when the one it found does not
hold the amount the document announces. The call sites at document level turned all three into a
bare failure:

```php
return ['res' => -1, 'message' => SupplierInvoiceHelper::refLookupErrorMessage(...)];
```

`syncFlows()` reads exactly the absence of `postponeflow` there: `$error++`, then `break`.
**The whole synchronization stops**, and every flow behind that one stays unimported until
somebody notices and clears the cause by hand.

`-3` needs no option at all to happen: the reference of a received document already exists here
on an invoice of another amount — someone recorded the invoice manually before the flow arrived,
or the vendor reused a number. `-2` is the tolerant-match ambiguity. Neither clears itself on a
retry, so today the queue stays shut for as long as nobody looks at it.

None of these has stored anything. The import runs inside the one transaction
`createSupplierInvoiceFromSource()` rolls back whole, so the flow can simply stay pending and come
back — which is what `postponeflow` is for, and what BG-3 already does one branch away.

## Measured, on a real synchronization

Dolibarr 23.0.4, real access point, three invoices emitted from the partner instance. Before the
synchronization, `TRI-FA-2609-0978` is recorded by hand for that vendor at 8.40 while the document
announces 240.00.

Before — `syncFlows()` on `main`:

```
res : {"res":-1, ... "Aborting synchronization due to errors."}
      Synchronisation interrompue sur le flux i_525903
      imported: TRI-FA-2609-0978  8.40   (the one recorded by hand)
                TRI-FA-2609-0979  absent
                TRI-FA-2609-0980  absent
```

After, same three flows, same bench:

```
res : {"res":1, ... "Flow i_525903 postponed, it will be retried on the next synchronization"}
      La synchronisation s'est terminée avec succès - flux reportés : 1
      imported: TRI-FA-2609-0978  8.40   (still the one recorded by hand)
                TRI-FA-2609-0979  240.00
                TRI-FA-2609-0980  240.00
```

The operator gets the same recommended action as before, with the same link to the invoice to fix.
What changed is that the two documents behind it are in.

That bench also showed the defect on its own, without being asked: the queue of the receiving
instance was already shut by an older flow whose line referenced a document nobody holds. It had
to be cleared by hand before these three could even be reached.

Per call site, verdict of the import (`res`, `postponeflow`, and what the batch loop does with it),
Dolibarr 22.0.5 / PHP 8.3:

| case | site | before | after |
|---|---|---|---|
| reference already here with another amount (`-3`) | already-imported check | ABORTED | POSTPONED |
| several invoices match the reference (`-2`) | already-imported check | ABORTED | POSTPONED |
| several invoices match the reference (`-2`) | BG-3 pre-check | ABORTED | POSTPONED |

Nothing was stored in any of the failing runs — checked in `llx_facture_fourn` after each.

A received Factur-X reaches the same sites: since #939, `FacturXProtocol` no longer duplicates the
import path and goes through `CIIProtocol::doCreateSupplierInvoiceFromSource()` like any other CII.
Replayed on that path — a Factur-X built from the same CII by the module's own merger, imported
inside a transaction that is rolled back, with a supplier invoice of that vendor already carrying
`DUP-260910-201717` at another amount:

| core | `main` | this branch |
|---|---|---|
| 18.0.10 / PHP 7.4 | `res=-1 postponeflow=NULL` → ABORTED | `res=-1 postponeflow=1` → **POSTPONED** |
| 22.0.5 / PHP 8.3 | `res=-1 postponeflow=NULL` → ABORTED | `res=-1 postponeflow=1` → **POSTPONED** |

Same `actioncode` on both sides (`SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT`), and nothing stored.

## What this changes

`refLookupPostponedResult()` sits next to `refLookupErrorMessage()`, which it uses for the
technical half of the message, and returns the postponed shape: the reason, the action, and a link
to the supplier invoices carrying that reference. `LINKED_INVOICE_LOOKUP_FAILED` is a separate
action code from `LINKED_INVOICE_NOT_FOUND`, because the document is not missing — the lookup is
what could not be answered.

The two `fetch()` failures on an id the lookup did return are answered the same way: the reference
was matched, so this is a loading problem, not a missing invoice.

`SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT` keeps its action code and its message, now built where
the amounts are known instead of in each provider — the two provider branches that built it are
removed, they can no longer be reached. Its wording said the amount in the message was the one of
the invoice found here; it is the one of the invoice being imported, as the bench above shows.

## The trade-off

A database error postpones instead of stopping. If the database were down for every flow, the run
would report success with N flows postponed rather than an error on the first. That is visible —
the count of postponed flows is on the panel — and a database down does not get that far anyway.
Against it: one hiccup no longer costs a whole batch.

## Not in here

- **The line-level site**, in the BT-128 loop of `createSupplierInvoiceLinesFromSource()`. It has
  the same defect and it is measured — an ambiguous reference on a line aborts the batch the same
  way — but answering it means widening the `@return` of that method, the line #911 rewrote to add
  its own parameter. #911 having landed, it is a follow-up of its own rather than a conflict to
  hand the maintainer at merge time.
- The `== 0` branches, a reference this Dolibarr does not hold: #911 at line level, #931 at
  document level, both merged.
- Making a postponed flow visible after the page is closed: #932.

## Tests

PHPUnit file by file, the module suite on dd18 (PHP 7.4) and dd25 (develop), and the six files
covering the import path on the eight cores dd18 → dd25: 98 green, 0 failure, 2 unchanged skips
(`MultidirOutputCompatTest`). The six import files replayed on dd18 / dd22 / dd25 after the scope
was narrowed: 130 tests green each.

Rebased on `main` (d979e470): no conflict. Of the open branches, `merge-tree` answers one, #1000,
and the section below settles it.

## Merging with #1000

Both branches edit the same `return` of `doCreateSupplierInvoiceFromSource()`, the one answering `findIdByRef() == -3`, on adjacent lines: #933 adds `postponeflow` and builds the `businessmessage` there, #1000 changes the amount that block announces. `merge-tree` conflicts in both directions, so the resolution is written here rather than left for the maintainer to decide at merge time.

Take the array of #933 whole, then carry the value of #1000 on **both** amounts it names - the `expectedamount` of `actiondata`, and the argument of `SupplierInvoiceFoundButWithdifferentAmount`:

```php
			return [
				'res' => -1,
				'postponeflow' => 1,
				'message' => SupplierInvoiceHelper::refLookupErrorMessage($supplierInvoiceId, $parsedHeader['documentno'] ?? '', 'while checking whether it was already imported'),
				'actioncode' => 'SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT',
				'actionurl' => 'none',
				'actiondata' => array('supplierref' => $parsedHeader['documentno'], 'socid' => (int) $socId, 'expectedamount' => $announcedTotalTtc),
				'action' => $action,
				'businessmessage' => $langs->trans('SupplierInvoiceFoundButWithdifferentAmount', $parsedHeader['documentno'] ?? '', $announcedTotalTtc)
			];
```

`$announcedTotalTtc` and not `grandTotalAmount`, on the message as well: once #1000 is in, the lookup searches on BT-115 with the BT-114 tolerance, so announcing BT-112 would name the operator an amount the module never looked for - 25.48 where 25.47 was searched, on the very document of #994. Control on the resolved block: no `grandTotalAmount` left in it, `$announcedTotalTtc` twice.

Nothing else collides: `SupplierInvoiceHelper` and `einvoicing.lang` merge on their own, and the three other lookup sites #933 rewrites are untouched by #1000. Merged both ways with that resolution, the two results are the same tree to the bit (`cce4a0c28be09aa0dd56fec1211d032e12782723`), `php -l` green on every file the merge touches.


eldy:
There is something i don'tunderstund with the postpone flow (currently).
We do not store it ok.
But what happen if the following flow is received 2 days after and ok. The postponed is not store but we store the next one.
So at next run, the date to start is the one of last ok. So AP return all after the last ok and the postponed is forgot.
Did i miss someting ?

daGrumpf-bxp:
Answered in detail on #932, where the fix belongs: you are right, a postponed flow drops out of the window for good once `MAX(updatedat)` of the stored documents has moved more than `EINVOICING_SYNC_MARGIN_TIME_HOURS` past it. Four received invoices had already been lost that way on my bench. #932 now takes the backlog again by `flowId`, which uses no date window, so the promise the message makes is kept.

It is the `postponeflow` convention, not this PR: this one only makes two more failures reach it, and without it a reference lookup answering -3 still stops the whole synchronization on every run.

eldy:
Waiting answer for #896 and #932 to take a decision

daGrumpf-bxp:
@eldy The answer is posted on #896: https://github.com/Dolibarr/dolibarr-community-modules/pull/896#issuecomment-5720864056 — #896 is off by default (`EINVOICING_ENABLE_MANUAL_ACTION_QUEUE`), so it cannot replace `postponeflow`, and #932 stays needed to keep a postponed flow from dropping out of the `updatedAfter` window for good.

This PR is independent of that decision: it only stops a reference lookup answering -3 from aborting the entire synchronization, whatever the storage of the pending flows ends up being.

Without it, one unresolved reference still kills every run for every user — including those who never turn #896's option on.

daGrumpf-bxp:
## Bench report — 20 Sep 2026

Rebased on today's `main`, so on top of `fea521bb`, and run end to end on a real round trip through the access point.

**Bench**: dd23, Dolibarr 23.0.4 / PHP 8.4. A fresh invoice was issued from another instance of the bench (`TRIB-FA-2609-0049`) and the receiving side was put in the state this pull request is about: `EINVOICING_TOLERANT_SUPPLIER_REF_MATCH=1` and two supplier invoices carrying the reference as a delimited substring, so `findIdByRef()` answers **-2** — the reference resolves to several invoices and the import cannot tell which one it is.

**Before this pull request**, that path returns a bare `['res' => -1, 'message' => …]`: the run stops on `ERROR_SYNCFLOW`, the whole batch with it, the user gets a technical sentence and no action, and the cause does not go away on its own — so the next run stops at the same place.

**With it**, the same flow (`i_656301`) answered with an action code and something to do, and the run carried on and imported its 14 other flows, with no transaction left open:

* action — *"Check the supplier invoices carrying the reference TRIB-FA-2609-0049, so that exactly one of them answers to it"*, with a button to `fourn/facture/list.php?search_refsupplier=TRIB-FA-2609-0049&socid=1`;
* business message — *"The received invoice TRIB-FA-2609-0049 refers to TRIB-FA-2609-0049, and that reference could not be resolved to a single supplier invoice…"*;
* technical detail kept in the tooltip — *`Several supplier invoices match reference "TRIB-FA-2609-0049" while checking whether it was already imported, cannot determine which one to use`*.

Once the ambiguity was removed — exactly the manual action the message asks for — the flow imported cleanly and the supplier invoice was created.

### What the rebase on `fea521bb` changed here

@eldy added a **generic relay** of `$res['businessmessage']` in both providers, with the technical message in the picto. That is the hook this pull request needed: the hard-coded `SUPPLIER_INVOICE_FOUND_WITH_BAD_AMOUNT` block it removes is now genuinely redundant — there is a single producer of that action code (`CIIProtocol`, the return this pull request completes with its `businessmessage`), the values are the same ones (`actiondata.supplierref` / `expectedamount` are `documentno` / `grandTotalAmount`), and the relay adds the picto that the hard-coded block used to overwrite.

Reading the same code: with `EINVOICING_ENABLE_POSTPONE_FLOWS` **off**, the flow falls through to the `isset($res['action'])` branch, where that relay serves the message just the same — so what this pull request brings does not depend on the postpone feature being enabled. The bench run above was made with the option **on** (that is what makes the flow observable as postponed), so that particular path is read from the code, not measured; happy to measure it too if it matters.

### Scope of the run

The bench ran this branch merged with #932, because it is the queue of #932 that makes a postponed flow observable at all. The part measured here is this pull request's: the action code, the message, the action, and the batch carrying on.

CI green (10/10) on the rebased head. PHPStan 2.1.12 clean on Dolibarr 18 and 24 and phan 5.5.2 clean were run on the merged tree, not on this branch alone.

PR #936, opened 2026-09-13
FIX [einvoicing] The last look no longer calls a line of work a discount

P-tite-tete:
Follow-up to #723, on the last review round of that PR — the comment landed a few hours before the merge, so neither point made it in.

## The last look called a line of work a discount

#723 fixed the line an operator names `(DEPOSIT)` while it carries no discount: it keeps the name it was given and the sign is right. The last look at the bottom of `buildinvoicelines.inc.php` was not told, and reported that same line on its way out:

```
ERR EInvoicing: line 1 of DD23-FA-2609-0459 carries the unresolved discount marker (DEPOSIT)
    in BT-153. The line is a discount whose source piece could not be read.
```

It is not a discount and there is no source piece — `einvoicingDiscountLabelOfLine()` established that two hundred lines above, on `fk_remise_except`. Nothing reaches the customer and the document is correct, so this is a log line only; but it is `LOG_ERR`, on the one shape the previous PR declares legitimate, and whoever chases the message lands on a line with nothing wrong with it.

The report now applies the same rule as the resolution: `fk_remise_except` is kept per line beside the rowid the loop already collects, and the BT-153/BT-154 report asks for it before firing. The equality against the sentinel list is unchanged — this only stops the report on a line that carries no discount.

The BT-97 half needs no such test: `einvoicingDiscountLabel()` returns a text for all four sentinels even with no discount object, so `$discountReason` is empty exactly when the description is not a sentinel and the fallback can never put a sentinel into `'reason'`. That half is a net for a construction path that does not exist yet, which is what its comment says.

## Comment blocks brought back to the house rule

`.agents/AGENTS.md` asks for five lines at most per block, and no more comment lines than lines of code touched. The blocks written for the sentinels ran to nine and fourteen. The long reasoning moves here, the comments keep the sentence the reader of the code cannot deduce from it — including the two paragraphs of prose at the top of `einvoicingDiscountSentinels()` and `einvoicingDiscountLabelOfLine()`.

Net effect on the three files: 32 lines added, 61 removed.

## Tests

`DiscountSentinelTest`: 8 tests, 45 assertions, OK on 23.0.4. `php -l` clean on the three files. No behaviour change beyond the report above, and the report is not covered by a test — the case needs a full invoice object through the include, which the suite does not build.


daGrumpf-bxp:
Measured on the eight cores, and it does what it says. Thank you for picking both points up after the merge rather than letting them drop.

## The fix, on the bench

Same harness as #723 (`/opt/t524/t723.php`, cases A→H, and `t723b.php`, cases I→M), same invoices regenerated under three trees exported side by side: `main`, this branch merged onto it, and a third one described below. Negative control printed at every run (`einvoicingDiscountSentinels()` defined, and the realpath of the module in use). Cores 18.0.10, 19.0.3, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1, 25.0.0-alpha.

On `main`, case G — a line of work an operator named `(DEPOSIT)`, carrying no discount, invoice `DD23-FA-2609-0459` — writes this into `documents/dolibarr.log` at every generation:

```
ERR EInvoicing: line 1 of DD23-FA-2609-0459 carries the unresolved discount marker (DEPOSIT)
    in BT-153. The line is a discount whose source piece could not be read.
```

On this branch it is gone, and nothing else moved: A and H keep their BT-97 reasons, B its two names, C and D their BT-113 at 600.00 and 480.00, E and F are still refused by BR-25 with the rank and the rowid (`#2 (id 5053)`, `#2 (id 5055)`), G still goes out named `(DEPOSIT)` at `250.00 / 300.00`, and I→M are unchanged (J on the `…NoSource` wording, K naming the invoice, M keeping `Reprise (DEPOSIT) du chantier`). `DiscountSentinelTest`: 8 tests, 45 assertions, OK on all eight, each under the PHP its core gets in CI (7.4 → 8.5).

## One simplification: the value is already in `$linesData`

`$lineDiscountIds` duplicates something the loop already writes. Line 749, in the same array literal, under the same `$numligne`, from the same expression:

```php
'fk_remise'                 => $line->fk_remise_except ?? null,
```

and the last look already walks `foreach ($linesData as $numligne => $vals)`. Nothing rewrites that key between the two — the only later writes on `$linesData[$numligne]` are `grosspriceamount` and the two `grosspricebasisquantity*`, just below. So the guard can read it directly and the parallel array goes away:

```diff
-$lineDiscountIds   	= [];	// Document line number => llx_facturedet.fk_remise_except, 0 when the line is not a discount
@@
-	$lineDiscountIds[$numligne] = (int) ($line->fk_remise_except ?? 0);
@@
-		if (!empty($lineDiscountIds[$numligne]) && in_array((string) ($vals[$field] ?? ''), $discountSentinels, true)) {
+		if (!empty($vals['fk_remise']) && in_array((string) ($vals[$field] ?? ''), $discountSentinels, true)) {
```

That is the third tree of the run above. On the eight cores its output is **identical** to this branch's, character for character, on A→H and on I→M — the only difference in the whole capture is the pid in the log lines. `php -l` clean.

Two reasons beyond the three lines: `'fk_remise'` is old (it predates the module rename), so the value is not something this branch has to start keeping; and one array fewer is one array that cannot drift from `$linesData` the day a line is added to it from somewhere else. `!empty()` reads `null`, `0` and `'0'` the same way, so the cast the new array does is not buying anything either.

Your call — the branch is correct as it stands, this only removes state.

## On the rest

The comment pass is right, and the docblocks of `einvoicingDiscountSentinels()` and `einvoicingDiscountLabelOfLine()` keep the sentence that cannot be deduced from the code, which is the part that mattered. No other remark.

---

**Added after the full bench batch, since this branch touches generation.** Same two trees, `main` and this one merged onto it.

- **PHPUnit, the whole grid**, 8 cores × 32 files, 256 runs per tree. The reds differ from run to run (17 against `main`, 20 against this branch); replayed one file at a time, sequentially, the four files involved are green on both trees on all eight instances. Flakes of the parallel run, nothing else.
- **QA matrix**, 9 instances × 14 generations per tree: 126 generations, `code=10` throughout, no failure.
- **63 CII documents regenerated under both trees: identical**, timestamp line aside. Worth one caution for whoever replays this — the first comparison showed 16 amount differences, all on situation invoices, because regenerating a situation ≥ 2 recomputes the cumulative amount. Regenerating the same 63 under `main` a second time gives exactly the branch's output: the difference was the regeneration rank, not the code.
- **Access point validator** on case G (`DD23-FA-2609-0459`, the line named `(DEPOSIT)` carrying no discount): `res=1`, accepted.
- **Two full lifecycles**, both directions, 200 → 201 → 205 → 211 → 212, nothing left pending.

No regression anywhere. The simplification above stays a suggestion.

PR #941, opened 2026-09-13
FIX: [einvoicing] A failed generation no longer swallows the warning before it, nor stays silent on Dolibarr 23

daGrumpf-bxp:
Two commits, two reasons the same failed generation is not reported. #960 carried the second one and is folded in here, so there is one branch to read instead of two.

When `generateInvoice()` fails and `EINVOICING_EINVOICE_CANCEL_IF_EINVOICE_FAILS` is off, the hook turns the protocol errors into something the core is supposed to show. On `main`:

```php
if ((float) DOL_VERSION < 23) {
    $this->errors = array_merge($this->errors, $protocol->errors);
    $this->warnings = array();
} else {
    $this->warnings = array_merge($this->errors, $protocol->errors);	// We want to return the error as a warning.
}
```

**1. The merge starts from the wrong array.** `$this->errors` is empty on that path — the core clears it before every hook call ([hookmanager.class.php](https://github.com/Dolibarr/dolibarr/blob/develop/htdocs/core/class/hookmanager.class.php), *"Clean class (an error may have been set from a previous call…)"*), and the two writes to it in `afterPDFCreation()` are the sibling branches that `return -1` at once. `$this->warnings` is **not** empty: `checkRequiredinformations()` answering `res == 0` pushed `InvoiceGeneratedWithWarnings` into it a few lines above, and that message has no other way out (its own `setEventMessages()` is commented out, the later one runs only when the generation succeeds). Assigning over the array drops it.

**2. The version test is one release too low.** The comment above it says the chain carrying `->warnings` to the user "appears whole in Dolibarr 23". Its last link arrives in 24, so on 23 the failure lands in an array nothing reads.

## What the core does, read in 18.0.10 / 20.0.4 / 21.0.4 / 22.0.5 / 23.0.0 / 23.0.4 / 24.0.0 / develop

|---|---|---|---|
| `pdf_sponge`, after a hook that answered < 0 | copies the errors, then **`return 1; // No error`** | `return -1` | `return -1` |
| `HookManager` collects `$actionclassinstance->warnings` | no | yes | yes |
| `commonGenerateDocument()` copies `$obj->warnings` | no | yes | yes |
| **something displays `$object->warnings`** | no | **no** | yes — `core/actions_builddoc.inc.php`, the only place in the whole core |

Bench, the eight cores, real HTTP session, same scenario everywhere: a customer whose SIREN and SIRET do not agree, so `checkRequiredinformations()` answers a warning and the generation is then refused.

Before:

| core | "Generate document" | validation |
|---|---|---|
| 23.0.4 | **nothing at all** | **nothing at all** |
| 24.0.1, 25.0.0-alpha | the generation failure only | nothing |

After:

| core | "Generate document" | validation |
|---|---|---|
| 18.0.10 → 22.0.5 | unchanged — see below | unchanged |
| 23.0.4 | **the failure + the SIREN and VAT warnings, as errors** | **same** |
| 24.0.1, 25.0.0-alpha | **the failure + the SIREN and VAT warnings** | nothing — see below |

Up to 22 nothing can be done from here: `pdf_sponge` copies the hook errors and then answers "no error", so `commonGenerateDocument()` takes the success branch and the page reports a success. That is also why the `< 23` branch this replaces never had any effect on those releases.

## Not fixed here

- on 24 and 25 the **validation** path still says nothing: `compta/facture/card.php` displays `$object->error`/`errors` and never `$object->warnings`, on any release. That one is #961, which says it with `setEventMessages()` for every core and both paths;
- the recipient reachability warning is in that array too, but it was never lost: it has its own `setEventMessages()` where it is raised. Pushing it into `$this->warnings` as well is a pre-existing duplication, which this fix makes visible on 24. One line to drop if you would rather have a single message — it is a policy choice, not part of this bug.

## Test

- Bench, eight cores (18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1, 25.0.0-alpha), both buttons over a real HTTP session, both values of the option: the tables above.
- PHPUnit grid, one file at a time per instance, same eight cores: **264/264**, identical to `main` measured in the same bench state.
- phan 5.5.2, PHPStan 2.1.12 (against the v18 and v24 cores) and phpcs with the repository ruleset: green.

Versions: 18.0.0 to 25.0.0.


einvoicing/class/actions_einvoicing.class.php
```diff
@@ -239,7 +239,9 @@ public function afterPDFCreation($parameters, $object, &$action, $hookmanager)
 										$this->errors = array_merge($this->errors, $protocol->errors);
 										$this->warnings = array();
 									} else {
-										$this->warnings = array_merge($this->errors, $protocol->errors);	// We want to return the error as a warning.
+										// Append to the warnings already collected above (configuration, routability, auto-send),
+										// which starting the merge from $this->errors used to drop.
+										$this->warnings = array_merge($this->warnings, $protocol->errors);	// We want to return the error as a warning.
```
eldy, line 244: Why note 
`$this->warnings = array_merge($this->warnings, $this->errors, $protocol->errors); // We want to return the error as a warning.`
so we are sure to not lose error (we want errors as warning) when we are in the else, but why removing the ->errors that may appear later...

daGrumpf-bxp: `$this->errors` is empty at that line, and it is the core that makes it so: `HookManager::executeHooks()` clears `$actionclassinstance->error` and `->errors` (and `->warnings` from 23) immediately before calling the method — *"Clean class (an error may have been set from a previous call of another method for same module/hook)"*, `hookmanager.class.php`, identical in 18.0.10 → develop. And inside `afterPDFCreation()` the only two writes to `$this->errors` are the sibling branches that `return -1` right away, so nothing can be sitting in it here.

Adding it back costs nothing, so say the word and I will. The one argument against: if it ever did become non-empty, the core already reports it — `executeHooks()` tests `!empty($actionclassinstance->errors)` and merges it into `$hookmanager->errors`, `pdf_sponge` and `commonGenerateDocument()` carry it onto `$object->errors`, which both `card.php` and `actions_builddoc.inc.php` display. Merging it into the warnings as well would then print the same text twice, once red and once orange.

Measured on the bench, Dolibarr 24.0.1, real "Generate document" button, on a customer whose SIREN and SIRET do not agree so the generation is refused: `main` shows one message (the generation failure), this branch shows two — the second being `InvoiceGeneratedWithWarnings` carrying the SIREN and VAT warnings, which has no other way to the screen.

Related, on the other branch of the same `if`: #960. On Dolibarr 23 a failed generation is currently reported nowhere at all, because `$object->warnings` is only ever displayed from 24.0.0.

daGrumpf-bxp:
Rebased on `main` and grown by one commit: #960 is folded in here and closed, so there is a single branch to read.

Second commit, in one line: the version test was `< 23`, but `$object->warnings` is displayed in exactly one place in the core, `core/actions_builddoc.inc.php`, and only from **24.0.0**. On Dolibarr 23 the failure went into an array nothing reads — measured on 23.0.4, no message at all, neither at validation nor on the "Generate document" button. It is now raised as an error up to 23, with the warnings collected before it merged in rather than cleared.

Measured over a real HTTP session on the eight cores (18.0.10 → 25.0.0-alpha), both buttons, both values of the option — the tables are in the description. Worth knowing for the review: up to **22** nothing a `afterPDFCreation` hook reports can reach the user at all, because `pdf_sponge` copies the errors and then answers `return 1; // No error`; the `return -1` appears in 23.0.0. That is why those releases say "The file was successfully generated" on a failed e-invoice, and why the `< 23` branch being replaced never had an effect there.

Your question on the merge is answered in the review thread above. PHPUnit grid 264/264 on the eight cores, same as `main` in the same bench state.

daGrumpf-bxp:
Screenshots of the two cases, taken on the bench through the real "Generate document" button, same invoice on both sides (a customer whose SIREN and SIRET do not agree, so the generation is refused).

**Dolibarr 24.0.1** — `main` shows the generation failure alone; the branch brings back `InvoiceGeneratedWithWarnings` with the SIREN and VAT warnings. Orange box both times: it is a warning, and it stays one.

Dolibarr 24

**Dolibarr 23.0.4** — `main` shows nothing at all; the branch shows the failure and the warnings raised before it. Red box here, because on 23 the error channel is the only one the core displays — the warnings ride along in it rather than staying invisible. If you would rather keep them out of an error box, dropping `$this->warnings` from that merge is one line, and they simply disappear again on 23.

Dolibarr 23

(Images hosted on a branch of my fork, so they are not part of this pull request.)

daGrumpf-bxp:
Screenshots of the two cases, taken on the bench through the real "Generate document" button, same invoice on both sides (a customer whose SIREN and SIRET do not agree, so the generation is refused).

**Dolibarr 24.0.1** — `main` shows the generation failure alone; the branch brings back `InvoiceGeneratedWithWarnings` with the SIREN and VAT warnings. Orange box both times: it is a warning, and it stays one.

Dolibarr 24.0.1, before and after

**Dolibarr 23.0.4** — `main` shows nothing at all; the branch shows the failure and the warnings raised before it. Red box here, because on 23 the error channel is the only one the core displays, so the warnings ride along in it rather than staying invisible. If you would rather keep them out of an error box, dropping `$this->warnings` from that merge is one line, and they simply disappear again on 23.

Dolibarr 23.0.4, before and after

PR #945, opened 2026-09-14
FIX: [einvoicing] A supplier Dolibarr refuses to update no longer stops the synchronization

### LINKED ISSUE #943: [Einvoicing] Failed update of supplier shouldn't interrupt the sync flow
### Bug

If update of supplier fails, the sync flow is interrupted.
This issue was described on the forum 
https://www.dolibarr.fr/forum/t/facture-electronique-probleme-dimport-du-au-telephone-du-fournisseur/51917

### Module

einvoicing

Import einvoice with long text as supplier phone number like `09.77.40.50.60 (service gratuit + prix appel)`

Error message is 
```
Informations techniques : ERROR_SYNCFLOW - Failed to synchronize flow i_506050: Failed to create supplier invoice from E-invoice document for flowId: i_506050. Thirdparty sync or creation error:
Thirdparty update error: Data too long for column ‹ phone › at row 1.
```

daGrumpf-bxp:
Fixes #943 — and the forum thread it comes from: https://www.dolibarr.fr/forum/t/facture-electronique-probleme-dimport-du-au-telephone-du-fournisseur/51917

## What happens today

A vendor (Total Energies in the forum thread) writes its phone number as `09.77.40.50.60 (service gratuit + prix appel)`. `llx_societe.phone` holds 20 characters up to Dolibarr 22, 30 from 23 on, so `Societe::update()` refuses the record and `_syncOrCreateThirdpartyFromEInvoiceSeller()` answers `res => -1` with no `postponeflow`. `syncFlows()` then does `error++` then `break`:

```
ERROR_SYNCFLOW - Failed to synchronize flow i_506050: [...]
Thirdparty update error: Data too long for column 'phone' at row 1.
```

The whole incoming synchronization of the instance stops, for every vendor, at every run, until somebody fills that phone by hand or turns the completion option off — the two workarounds the forum thread ended up with.

## What this changes

**The number is read as a number.** The telephone number of the seller contact (BT-42) is a Text, and so is the fax number the EXTENDED profile adds next to it (BT-X-107, outside EN 16931): nothing in the standard keeps a vendor from adding a sentence to them, and vendors do. Before being written to a thirdparty, the value is reduced to the number it carries: a value made of number characters only is kept exactly as written, whatever its local formatting; anything else is cut down to the first group of at least 6 digits; nothing is copied when it holds none. This also covers the creation of a thirdparty, where there is nothing to fall back on.

**A refused completion no longer costs the invoice.** Completing a thirdparty from the document is a convenience (`EINVOICING_THIRDPARTIES_COMPLETE_INFO`), and the vendor is already identified when it runs. Whatever else Dolibarr refuses — a name longer than its column, a trigger of another module — the thirdparty is now saved again *without* that completion: the document is imported, the thirdparty keeps the data it had, and the message of the flow says what could not be copied.

Nothing is postponed here, so this adds no new case to the postponing discussed in #718: the flow is imported, not deferred. The duplicate supplier code (`-3`) keeps its own answer and its own action.

The vendor flag, the thirdparty codes and the extrafields guard move to `_prepareThirdpartyForImportUpdate()`, so the second save applies exactly the same preparation as the first.

## Bench

Bench documents built out of `samples/dd18_CII_standard.xml`, seller identifier of their own so the lookup is never ambiguous, replayed through the real `createSupplierInvoiceFromSource()` inside a rolled-back transaction, printing the decision the batch loop would take:

* **A** — phone of the forum thread (`09.77.40.50.60 (service gratuit + prix appel)`), vendor already in Dolibarr
* **B** — number characters only but longer than the column (`01.23.45.67.89/01.23.45.67.90/01.23.45.67.91`), so the completion still fails
* **C** — phone of A, on a vendor nobody holds yet (creation path)

| | before | after |
|---|---|---|
| A / B / C on 18.0.10 and 25.0.0-alpha | **6/6 SYNCHRONIZATION ABORTED** | — |
| A / B / C on 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1, 25.0.0-alpha | — | **21/21 imported** |

Read back from another process after a real import on 22.0.5:

* A → invoice imported, thirdparty completed, `phone = 0977405060`
* B → invoice imported, thirdparty left as it is (`phone` still empty), vendor flag and code still written, warning raised
* C → invoice imported, thirdparty created with `phone = 0977405060`

Unchanged elsewhere: a standard sample still imports on 18.0.10 / 22.0.5 / 25.0.0-alpha, and the module test suite is green on the eight cores (256/256).

Not exercised on the bench, and said as such: PostgreSQL (the bench CLI has no pgsql extension). There, a failing statement aborts the surrounding transaction, so the second save cannot succeed and the behaviour stays what it is today — the reading of the phone number, which is what the reported case needs, applies all the same since it prevents the statement from being sent at all.

Tested on Dolibarr 18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1 and 25.0.0-alpha.


daGrumpf-bxp:
The two PHPStan failures are not from this branch: `main` itself reports exactly the same two, on every core, since 0576af19 — the return shape of `_syncOrCreateThirdpartyFromEInvoiceSeller()` changed without its docblock following. #944 fixes it; this branch goes green once that lands. Everything else here is green (phan, PHPUnit 18 and 24, phpcs, package install, PDF/A-3).

PR #947, opened 2026-09-14
FIX: [einvoicing] An amount already paid is only a deposit when the document says so

daGrumpf-bxp:
Answers the TODO left in commit 87b9ac5 on #904: *"Detect if prepaid because of deposit or else. May be we have also ref of deposit ?"*

## The norm answers it

**BT-113 on its own discriminates nothing.** EN 16931 says only `BR-CO-16` (BT-115 = BT-112 − BT-113 + BT-114), `BR-DEC-16` (2 decimals) and `CII-SR-484` (at most one occurrence). No rule requires a BG-3 when BT-113 > 0.

What carries the information is **BT-23**, the invoicing framework. FNFE `France_RFE` V1.4, `CII/EN16931/schematron/BR-FR-Flux2-Schematron-CII.sch`:

* **`BR-FR-CO-09`** (line 1155, `flag="fatal"`) — *if BT-23 is B2, S2 or M2 (invoice already paid), then BT-113 = BT-112, BT-115 = 0, and BT-9 is the payment date*;
* **`BR-FR-CO-08`** (line 1142) names the other half — *B4, S4, M4 = final invoices after a deposit*.

The document attached to `87b9ac5` carries `<ram:ID>B2</ram:ID>`, `TotalPrepaidAmount = 48.41 = GrandTotalAmount`, `DuePayableAmount = 0.00`. It matches `BR-FR-CO-09` exactly: an invoice the vendor cashed in as he issued it, with no deposit to look for.

## One rule, read at the two places that ask the question

`depositAnnouncedByDocument()` answers *"is there still a deposit for the import to attach?"*, and both guards read it:

* **the document level guard** (this PR's starting point), which marks an invoice short of its deduction;
* **the line level postpone added by #911**, merged today, which waits for a deposit when an unqualified BT-128 does not resolve. Under B2 that wait can never end — measured below, and reported on #911.

## What the guard now does

1. **BT-23 in B2 / S2 / M2** — already paid, nothing to deduct. Reported, never blocked.
2. Otherwise, **no BG-3** — nothing points at an invoice to deduct. `invoiceRefDocs` is the only thing the import ever attaches a deposit from (`invoiceRefDocs` → invoice of type deposit → `insert_discount()`), so a document without one can carry no deduction the import missed. Reported, never blocked.
3. Otherwise — BT-113 is confronted with what the invoice actually carries, as before.

At line level the same rule applies to the postpone: a document declaring itself already paid announces no deposit, so the flow is not held back waiting for one.

The validation trigger goes back to reading the mark, since the mark now carries a prepaid amount only in case 3. **That removes the constant forced to zero, the unreachable `elseif` under it, and the `@phpstan-ignore-next-line booleanAnd.leftAlwaysFalse` added in `6c48e86a` to silence it.** This PR takes a suppression away rather than adding one.

## Measured

Real imports through `createSupplierInvoiceFromSource()` in transactions that are rolled back, then `validate()` on the invoice each one builds. Replayed today against `main` d979e470 — so with #911, #948/#950 and #952 merged — on Dolibarr 18.0.10, 22.0.5 and 25.0.0-alpha, **same verdict on the three cores**.

| document | BT-23 | BG-3 | BT-113 | `main` d979e470 | this branch |
|---|---|---|---|---|---|
| the attached sample | B2 | — | 48.41 | marked, 205/206 withdrawn | **no mark, nothing withheld, validates** |
| prepaid, no reference | B1 | — | 120.00 | marked, invoice at 540.27, **refused** | **clean, 540.28, validates** |
| reference to the deposit | B1 | yes | 120.00 | deduction attached, validates | unchanged |
| already paid **and** a reference | B2 | yes | 120.00 | deduction attached, validates | unchanged |
| reference to a non-deposit invoice | B1 | yes | 120.00 | marked, **validates all the same** | **marked, refused** |
| reference not found | B1 | yes | 120.00 | postponed | postponed |

The first line is the document attached to `87b9ac5`: the mark it carries on main is read by a second place the validation trigger cannot reach, `SupplierInvoiceHelper::totalsMismatchBlocks()` in `einvoicing.class.php`, which withdraws "Approved" and "Partially approved" from the statuses that invoice can still be answered with. Here it carries no mark at all, so neither reader has anything to act on.

The second line shows a second effect of the guard as it stood: its early return skipped the VAT alignment of #781, so an invoice a cent short of what its document announced stayed short and was refused on its totals — the constant forced to zero did not unblock it. It now carries 540.28 and validates.

The fifth line is the one that costs money, and it is the one that moved the other way: the document says 120.00 was paid, references an invoice that exists in Dolibarr but is not a deposit, so nothing is deducted and the invoice would be paid in full. On main it validates — the constant forced to zero neutralises the mark for every document, this one included. Here the guard is kept for it alone, and it is refused with the message naming the missing deduction rather than a total difference that does not exist.

The second commit is about the report itself, and comes from the same run: on the fourth line the deduction *is* attached, and the message still told the operator the document pointed at no deposit and invited a payment on top of it — the same 120.00 settled twice. It now names what the invoice holds, and says nothing at all when that covers the announced amount.

### The line level site, added by #911 today

Two documents differing only by BT-23, each with one line carrying `AdditionalReferencedDocument` / `IssuerAssignedID` + `TypeCode 130` and **no `ReferenceTypeCode`**. Both validate against the FNFE `France_RFE` V1.4 rules (`fnfeval`): `VALID urn:cen.eu:en16931:2017`.

| document | BT-23 | BT-113 | BT-115 | `main` | this branch |
|---|---|---|---|---|---|
| already paid | **B2** | 540.28 | 0.00 | **flow postponed, for ever** | **imported, 540.28** |
| after a deposit | B1 | 120.00 | 420.28 | flow postponed | flow postponed |

One case moves. The other is the case #911 is right about, and it does not.

Also replayed: the six BG-22 documents of #904 (plain / allowance / charge / prepaid / the three together / prepaid with the deposit linked by hand) — all six import and validate, and none of them carries a BG-3, which is why that PR could not tell the two cases apart in the first place.

* PHPUnit of the module, file by file, on **8 cores** (18 to 25), the new file included: **272 runs, 272 OK**.
* Generation matrix, 9 instances × 14 documents: **126 generations**, every one accepted.
* Full lifecycle end to end, 8 versions × both directions: **16 cycles out of 16**, each through `200, 201, 205, 211, 212`.
* The eight documents of the table against the FNFE `France_RFE` V1.4 rules (`fnfeval`): seven `VALID urn:cen.eu:en16931:2017`, and the eighth invalid by construction — a B2 that still has something to pay is what `BR-FR-CO-09` refuses.
* `phpcs`, PHPStan (`[OK] No errors`) and phan clean locally, with the ruleset and config of the repo.

`test/phpunit/PrepaidIsADepositTest.php` covers the rule over every BT-23 value `BR-FR-08` allows (23 tests). It is pushed here because it was asked for; say the word and I drop it, per the Critical Rules of `.agents/AGENTS.md`.

## Where the rule applies

* **On both protocols.** `FacturXProtocol` no longer duplicates the import: `parseReceivedDocument()` hands the embedded CII to `parseInvoiceHeader()`, which carries the `businessProcessId` XPath, so rule 1 reads BT-23 on a received Factur-X as it does on a received CII. Only the opt-in external reader (`EINVOICING_USE_EXTERNAL_FACTURX_READER`, development and cross-checking) builds a header without the key, and rule 2 covers that path.
* **Rule 2 is not redundant with rule 1**, and the schematron says why: the same fully prepaid document requalified as `B1` comes out `VALID`. Nothing in the norm forces a vendor who cashed his invoice in to declare B2, so BT-23 is a declaration to be trusted in the positive direction only — its absence proves nothing, and a document pointing at no BG-3 is the other half of the answer. `BR-FR-08` does make BT-23 mandatory since V1.4, but in reception that is a `warning` until 30 September 2026.
* **A deposit the import attaches correctly** could not be validated either, which predates #904: `insert_discount()` writes a negative line, so the invoice totals BT-112 − BT-113 while the guard of #861 compares it with BT-112. Reported in #948, fixed by #950, merged since — the two rows of the table with a deduction attached are measured with it in.

## How the premise spread, since it is the interesting part

*"BT-113 > 0 means a deposit is missing"* is ours: #884 → #931 → #904 → #911, four PRs in four days. It was never questioned because **every document on our bench carries `B1`** — the six of #904 included. A wrong model stays invisible when every specimen comes out of the same mould, and the counter-example came from the maintainer rather than from the bench. A `B2` specimen now sits beside the `B1` ones.


einvoicing/core/triggers/interface_98_modEInvoicing_EInvoicingTriggers.class.php
```diff
@@ -332,12 +332,8 @@ public function runTrigger($action, $object, User $user, Translate $langs, Conf
...
 			$announced = SupplierInvoiceHelper::totalsMismatch((int) $object->id);
 			if ($announced !== null) {
-				// An announced prepaid amount can be because paid by a deposit (in this case we control it exists and is linked to)
-				// or just because the company has withdrown found automatically by a prerecorded credit card or diret debit like Amazon do (in this case, we don't need a deposit)
-				$isprepaidannouncedforadeposit = 0;		// TODO Detect if prepaid because of deposit or else. May be we have also ref of deposit ?
-
-				// @phpstan-ignore-next-line booleanAnd.leftAlwaysFalse
-				if (SupplierInvoiceHelper::totalsAgreeWithDocument($object, $announced['tva'], $announced['ttc'], ($isprepaidannouncedforadeposit && $announced['prepaid']) ? $announced['prepaid'] : null)) {
+				// A prepaid amount is in the mark only for a document referencing the invoice it was paid on (BG-3).
+				if (SupplierInvoiceHelper::totalsAgreeWithDocument($object, $announced['tva'], $announced['ttc'], $announced['prepaid'] ?? null)) {
```
eldy, line 336: This seems to reintroduce the bug that block importing the invoice 
einvoicing/test/samples/standard_einvoice_with_prepaid_defined_by_amazon_cii.xml

$announced['prepaid'] remains set by totalsAgreeWithDocument() and totalsAgreeWithDocument() was not modified. So an the xml example, a value is set for prepaid when it is not a deposit.

And next test totalsAgreeWithDocument() seems restored to old code so we should block here again.

Did i miss something ?

daGrumpf-bxp: One hunk above this one: the prepaid amount never reaches the mark any more. `alignInvoiceTotalsWithDocument()` sets `$announcedPrepaid` to `null` before anything is written when BT-23 is B2/S2/M2 or the document carries no BG-3, so `flagPrepaidMismatch()` is not called, `totalsMismatch()` returns no `prepaid` key, and the `?? null` here keeps the trigger on the two totals. On your sample the import now leaves no mark at all, so this `if` is not even entered - the fix moved from the single reader to the single writer, which is why `totalsAgreeWithDocument()` did not need to change.

Re-measured today against `main` d979e470, on Dolibarr 18.0.10 / 22.0.5 / 25.0.0-alpha, real imports in rolled back transactions: your document is marked `{"tva":8.07,"ttc":48.41,"prepaid":48.41}` on main and `totalsMismatchBlocks()` still withdraws 205/206 from the sendable statuses (`einvoicing.class.php:1066` reads the same `prepaid`, and that reader is out of reach of the trigger), while on this branch it comes out unmarked, unblocked and validating on the three cores.

The same run shows two deltas the trigger patch alone cannot reach: a B1 document without BG-3 was refused on its totals on main (540.27 against the 540.28 it announces - the early return skipped the VAT alignment of #781) and validates here; and a document referencing an invoice that is *not* a deposit - the case that costs money - validates on main and is refused here, which is the guard we want kept. One testing caveat: an invoice imported before this branch keeps the mark already stored on it, so the flow has to be re-imported, not just the code redeployed.

Issue #955, opened 2026-09-14
einvoicing: the deposit reprise line carries no reference to the deposit invoice (EXT-FR-FE-BG-06)


daGrumpf-bxp:
## The gap

The module implements the **first option** of XP Z12-014 §3.2.19 (deposit invoice / final invoice after a deposit), and says so in `einvoicing/lib/buildinvoicelines.inc.php:463`:

> `// This is the first method described into XP_Z12-014 using the line into field BT-153 / BT-154`
> `// The second method need to use the field BT-113. We don't use it as we use the first method.`

That option requires, on the deposit reprise line (Annexe A V1.4, p. 102):

> Référence à facture antérieure à la ligne (**EXT-FR-FE-BG-06**) : la référence à la facture d'acompte, numéro, date, et surtout le codetype (**EXT-FR-FE-137**), égal à **386** qui permet de repérer cette ligne comme étant une reprise d'acompte pour un traitement automatique.

The schematron gives the exact path (`CII/EXTENDED-CTC-FR/schematron/BR-FR-Flux2-Schematron-CII.sch`, rule contexts of BR-FR-01/02/03/04):

```
IncludedSupplyChainTradeLineItem/ram:SpecifiedLineTradeSettlement/ram:InvoiceReferencedDocument
    /ram:IssuerAssignedID                                EXT-FR-FE-136
    /ram:TypeCode                                        EXT-FR-FE-137   -> 386
    /ram:FormattedIssueDateTime/qdt:DateTimeString       EXT-FR-FE-138
```

**We write none of it, and we read none of it.**

- **Emission.** `CIIProtocol::buildLineItem()` used to write a line level reference in the wrong slot — `ram:AdditionalReferencedDocument` with `TypeCode 130`, the BT-128 invoiced object identifier. #912 was right that the slot was wrong, and #921 removed it — without writing the right one. Since #921 the line carries no reference at all, and `depositInvoiceRef` / `depositInvoiceDate`, still collected in `buildinvoicelines.inc.php:746-748`, are read by nothing.
- **Reception.** The only line level reference we ever look at is `additionalRefDocs` (`CIIProtocol.class.php:314`, `./ram:SpecifiedLineTradeSettlement/ram:AdditionalReferencedDocument`). A conformant first-option invoice from a third party arrives with its deposit reprise invisible to us.

[...]

PR #957, opened 2026-09-14
TEST: [einvoicing] The FNFE reference documents, read in CI and confronted with what we write

daGrumpf-bxp:
## What

The conformance job already downloads the whole FNFE package, pinned at `v1.4.0.04`, for its rules. That package also ships **the documents the FNFE publishes as the answer to its own use cases** — `Z.example/TEST`: multi-vendor, sub-lines, self-billing, corrective invoice, deposit, across BASIC WL, EN16931, EXTENDED and EXTENDED-CTC-FR. Twenty-eight CII documents, conformant by construction, downloaded on every run and read by nothing of ours.

This reads them with the module's own parser — the standalone CII of each, **and the fifteen Factur-X PDF the package ships next to them**, which go through one step the XML never reaches: `PdfAttachmentExtractor` pulling the CII out of the PDF/A-3 attachment, the default path of `FacturXProtocol::parseReceivedDocument()`.

They answer a question **no schematron can**: a reader that loses a term produces no invalid document. The rules stay green while the information never reaches Dolibarr. Nothing in this repository was measuring that.

## Two steps

**Fatal**, on what is unambiguous — a document the parser refuses, a document whose lines it does not all bring back, or one of fourteen header terms (BT-1, BT-3, BT-5, BT-10, BT-12, BT-13, BT-23, BT-27, BT-31, BT-44, BT-109, BT-112, BT-113, BT-115) the document carries and the parser leaves empty. **All twenty-eight pass it today**, so it lands green and stays a regression gate.

**The PDF and its XML twin must parse to the same thing.** The package ships both renderings of each invoice, so the comparison costs nothing: an extraction that loses or alters what it pulls out produces no invalid document and no missing term — only a difference with the twin sitting beside it, and nothing was looking at that. Negative control: a PDF paired with the XML of another invoice of the corpus is refused, exit 1, so the comparison is not vacuous.

**Informative**, never fatal: what the parser brings back from *none* of them, printed into the job summary. Measured on the corpus as it stands:

```
 28  ApplicableHeaderTradeSettlement/ram:PaymentReference              BT-83
 27  ApplicableHeaderTradeDelivery/ram:ShipToTradeParty/ram:Name       BG-13 / BT-70
 26  SellerTradeParty/ram:Description                                  BT-33
 24  ApplicableTradeTax/ram:DueDateTypeCode                            BT-8
 23  ShipToTradeParty/ram:PostalTradeAddress/ram:LineOne, LineTwo      BT-75 / BT-76
 18  BuyerTradeParty/ram:DefinedTradeContact/ram:DepartmentName        BT-56
 18  SellerOrderReferencedDocument/ram:IssuerAssignedID                BT-14
 18  AdditionalReferencedDocument IssuerAssignedID + ReferenceTypeCode BT-18 / BT-18-1
 16  ReceivableSpecifiedTradeAccountingAccount/ram:ID                  BT-19
  4  SpecifiedLineTradeAgreement/ram:ItemSellerTradeParty/*            line level seller, multi-vendor
  2  PayeeTradeParty/ram:Name, ram:ID, legal org                       BG-10
  2  BuyerTradeParty/ram:URIUniversalCommunication/ram:URIID           BT-49
  2  SpecifiedTradePaymentTerms/ram:DirectDebitMandateID               BT-89
  2  ApplicableHeaderTradeSettlement/ram:CreditorReferenceID           BT-90
```

That list is the one to shorten. It is printed where it cannot be forgotten rather than turned into a red job nobody can fix in one pass — and it is not proposed as a scope for this PR.

Worth stating plainly: **BG-13 is written by the emission and read by nothing.** A delivery address we send out comes back from a platform and is dropped.

## Same command on a developer machine

```sh
DOLI_ROOT=/var/www/dd23/htdocs php einvoicing/test/conformance/import-corpus.php <package>/Z.example/TEST
DEEP=1 DOLI_ROOT=... php einvoicing/test/conformance/import-corpus.php <package>/Z.example/TEST   # the inventory
```

No second harness to keep in step with the first: CI runs that line.

## Measured

Run on Dolibarr 23.0.2 / PHP 8.3: `43 document(s) read, 15 Factur-X PDF compared with its XML twin, 0 with a gap`, exit 0. Lifecycle messages (CDAR) and the UBL rendering of the same invoices sit in the same tree and are skipped on the root element, not on the file name — a CDAR names the invoice it reports on, so its text mentions `CrossIndustryInvoice` without being one.

No production code is touched.

## The other direction: what we never write

Reading the corpus answers half the question. The other half — which terms of those documents this module cannot produce — would be answered wrongly by a generated specimen: an element missing from one document may only mean that invoice had nothing to put there. So `emitted-terms.php` measures what the builder is **able** to construct, anywhere in its source, and confronts that with the elements the reference documents carry, by profile family.

Twenty-two element names come out, none of which the module can emit whatever the invoice:

```
BASICWL   8/84   PaymentReference (BT-83) · ReceivableSpecifiedTradeAccountingAccount (BT-19)
                 PayeeTradeParty (BG-10) · SellerTaxRepresentativeTradeParty (BG-11)
                 DirectDebitMandateID (BT-89) · CreditorReferenceID (BT-90)
                 PayerPartyDebtorFinancialAccount (BT-91) · DespatchAdviceReferencedDocument (BT-16)
EN16931   4/90   BT-83 · BT-19 · SellerOrderReferencedDocument (BT-14) · BasisQuantity (BT-149/150)
EXTENDED  10/97  the four above, plus ItemSellerTradeParty, ParentLineID, LineStatusReasonCode
                 (multi-vendor and sub-line documents) and the payment card group
```

Two of them the module **reads and never writes** — checked in the source, not inferred: `BasisQuantity` (BT-149/150) and `DespatchAdviceReferencedDocument` (BT-16). Only the second is a gap: `resolveLineUnitPrice()` divides the received price by its base quantity, so Dolibarr stores a true unit price and emitting no base quantity — which the norm reads as 1 — is the correct document. The list is a measurement, not a verdict; #959 sorts it.

## And a guard on what stops being written

The case this repository has just lived through: the reference the deposit reprise line carries disappeared in #921 and came back in #955, and in between not one test went red — a write leaving the module turns no document invalid, the invoices that needed it simply stop carrying it.

The first version of this guard compared the set of element *names* found in the builder's source. **Replayed against that very case, it stayed green**, so it was replaced rather than shipped: the five names #921 took away (`AdditionalReferencedDocument`, `IssuerAssignedID`, `TypeCode`, `FormattedIssueDateTime`, `DateTimeString`) are all still constructed elsewhere, so the set never moved. Only 77 of the 122 names had a single construction site; the other 45 are exactly the generic, reused ones a regression of this kind touches.

So `emitted-paths.php` signs what is actually written. One document per supported profile, built from fixtures that never change — a full line, a deposit reprise line, an exempt line, every optional header branch — and the baseline is the set of element **paths** those documents carry: `ram:InvoiceReferencedDocument` under a line is not the same capability as the same element under the header. Because the fixtures are fixed, an absent path means one thing only: nothing wrote it. A path may appear freely; a path disappears only on purpose, with `--update-baseline` and a word in the pull request.

Replayed across the three states of the case, with the fixtures of this branch:

| baseline → tree | result |
|---|---|
| before #921 → after #921 | **15 paths lost, exit 1**, naming `IncludedSupplyChainTradeLineItem/SpecifiedLineTradeSettlement/AdditionalReferencedDocument` in each profile that wrote it |
| after #921 → #955 branch | the 12 paths of the `InvoiceReferencedDocument` that replaces it, reported as gained |
| before #921 → today's `main` | 25 paths lost: those 15, plus the 10 of the ShipTo `GlobalID` #929 removed **on purpose** — which is what declaring a removal looks like |

### What the fixtures reach

A guard is worth what its fixtures reach, so that was measured too, not assumed. The first fixtures made **91** of the element names the builder constructs come out; the 31 that never fired were gaps in the fixture, not in the module — four keys belonged to the reading side and `buildXML()` never looks at them, `_chorus` was the flag that drops the very BT-18 references the fixture carried, `_project` was an array where BT-11 needs a `Project`, the account reference is the *fallback* of the IBAN so no single invoice can carry both, and the lifecycle answer has a builder of its own that `buildXML()` cannot reach at all.

Now, per that same measure:

| | |
|---|---|
| element names the builder constructs, that the fixtures emit | **120 / 120** |
| element names the 28 FNFE reference documents carry, that our generated documents carry | **99 / 115** |
| of the 16 missing | all 16 are elements the module cannot construct at all — **none is buildable but left unexercised** |

Per family, against the corpus: BASICWL 75/84, EN16931 85/90, EXTENDED 86/97. `CrossDomainAcknowledgementAndResponse` is signed as its own family — 996 paths in all.

`emitted-terms.php` keeps the other direction, the gap against the reference documents, and is no longer fatal: element names read out of a source say what the module *could* write, never what it *does* write. It also stopped counting `document.createElement("form")` from the javascript these pages print, which is why the capability set is 120 names and not 122.

### And the answer, written down

A list printed in a job summary is not an answer, so the last commit is the answer. `doc/COVERAGE-AND-LIMITS.md` states the scope of the module in one place: the nine billing frameworks (BT-23) it produces out of the twenty `BR-FR-08` accepts, the two it cannot reach at all — **multi-vendor** and **bidirectional self-billing**, whose mandatory line subtype, parent line and line-level seller have no source in a Dolibarr invoice — and the **seller belonging to a VAT group**, which `BR-FR-CO-15/BT-29-1` makes impossible without BG-11. UBL is named for what it is: declared and disabled, both directions.

Then the two element lists themselves, each line with the reason it is not written or not read, and the two entries that look like gaps and are not — `BasisQuantity`, which `resolveLineUnitPrice()` folds into a true unit price, and the gross-price discount, which the module writes at the settlement level instead. Every list carries the command that reproduces it, so none of it has to be trusted.

The rule the page applies is the one given on #959: an element no rule makes mandatory is not written, because a field written is a field that can be written wrong. The README now points at `doc/`.


daGrumpf-bxp:
Follow-up opened as #958, with what the parser brings back from none of these documents and where the core could hold it. This PR stays what it is: the measurement, and the gate that keeps the 43 documents readable.

PR #960, opened 2026-09-14
FIX: [einvoicing] On Dolibarr 23 a failed e-invoice generation was reported nowhere

daGrumpf-bxp:
`generateInvoice()` failing while `EINVOICING_EINVOICE_CANCEL_IF_EINVOICE_FAILS` is off is reported to the user through one of two channels, picked by the core version:

```php
if ((float) DOL_VERSION < 23) {
    $this->errors = array_merge($this->errors, $protocol->errors);
    $this->warnings = array();
} else {
    $this->warnings = array_merge(…, $protocol->errors);
}
```

The comment above that test says the chain carrying `->warnings` to the user "appears whole in Dolibarr 23". Its last link arrives one version later, so on 23 the failure lands in an array nothing reads.

## What the core does

Read in the sources of 18.0.10 / 20.0.4 / 21.0.4 / 22.0.5 / 23.0.0 / 23.0.4 / 24.0.0 / develop:

|---|---|---|---|
| `pdf_sponge`, when the hook answered < 0 | copies the errors, then **`return 1; // No error`** | `return -1` | `return -1` |
| `HookManager` collects `$actionclassinstance->warnings` | no | yes | yes |
| `pdf_sponge` copies `$hookmanager->warnings` | no | yes | yes |
| `commonGenerateDocument()` copies `$obj->warnings` | no | yes | yes |
| **something displays `$object->warnings`** | no | **no** | yes — `core/actions_builddoc.inc.php`, the only place in the whole core |

Bench, the eight cores, real HTTP session, same scenario everywhere: a customer whose SIREN and SIRET do not agree, so `checkRequiredinformations()` answers a warning (`res == 0`) and the generation is then refused.

| core | "Generate document" button | validation |
|---|---|---|
| 23.0.4 | **nothing at all** | **nothing at all** |
| 24.0.1, 25.0.0-alpha | the failure, as a warning | nothing |

With this change:

| core | button | validation |
|---|---|---|
| 18.0.10 → 22.0.5 | unchanged | unchanged |
| 23.0.4 | **the failure, as an error, with the warnings raised before it** | **same** |
| 24.0.1, 25.0.0-alpha | unchanged | unchanged |

The warnings collected earlier are merged into the errors rather than cleared, so the configuration and routability messages are not dropped on the way.

## Why 18 → 22 cannot be served by this branch

Their `pdf_sponge` copies `$hookmanager->error`/`errors` and then answers `return 1; // No error`, so `commonGenerateDocument()` takes the success branch and nothing an `afterPDFCreation` hook reports can reach the screen — which is why those releases answer "The file was successfully generated" on a failed e-invoice. From 23.0.0 the same block gained `return -1`, and that is what makes the error channel work there. Worth stating, because the `< 23` branch this replaces was written assuming it already worked below 23.

## Not fixed here

On 24 and 25 the **validation** path still reports nothing: `compta/facture/card.php` displays `$object->error`/`errors` and never `$object->warnings`, on any release — only the "Generate document" button does. That deserves its own change; happy to open it.

## Test

- Bench, eight cores (18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1, 25.0.0-alpha), both paths driven over a real HTTP session: the two tables above.
- PHPUnit grid, one file at a time per instance, same eight cores: **264/264**, identical to `main` measured in the same bench state.
- phpcs with the repository ruleset green.

This touches the same block as #941, which fixes the `>= 23` branch of the same `if`. The two merge cleanly in either order (checked with `git merge-tree`), and the merged tree also runs 264/264 on the eight cores. Happy to rebase whichever lands second.

Versions: 18.0.0 to 25.0.0.


daGrumpf-bxp:
Folded into #941 and closed: both changes are in the same nine lines of `afterPDFCreation()`, and one branch is easier to review than two. The commit is unchanged, the description of #941 carries the measurements of this one.

PR #961, opened 2026-09-14
FIX: [einvoicing] A failed e-invoice generation is now said to the user

daGrumpf-bxp:
When `generateInvoice()` fails, the hook hands the reason back in `->errors` or `->warnings` and says nothing itself, so whether the user ever reads it is left to the core. Read in 18.0.10 / 19.0.4 / 20.0.4 / 21.0.4 / 22.0.5 / 23.0.0 / 23.0.4 / 24.0.0 / develop, and measured on all of them, the answer is usually "never".

## What the core does with what the hook reports

|---|---|---|---|
| `pdf_sponge`, after a hook that answered < 0 | copies `->error`/`->errors`, then **`return 1; // No error`** | `return -1` | `return -1` |
| something displays `$object->warnings` | no | no | yes — `core/actions_builddoc.inc.php`, the only place in the whole core |
| `compta/facture/card.php` displays `$object->warnings` | **no** | **no** | **no** |

Up to 22 the generation is therefore reported as a success, whatever the hook answered; and on every release the **validation** path — the moment the e-invoice is actually produced — displays nothing the hook puts in `->warnings`.

Bench, eight cores, real HTTP session, same scenario everywhere: a customer whose SIREN and SIRET do not agree, so `generateInvoice()` refuses. Both values of `EINVOICING_EINVOICE_CANCEL_IF_EINVOICE_FAILS`, both buttons.

Before:

| core | "Generate document" | validation |
|---|---|---|
| 23.0.4 | nothing | nothing |
| 24.0.1, 25.0.0-alpha | the failure, as a warning | **nothing** |

After:

| core | "Generate document" | validation |
|---|---|---|
| 18.0.10 → 25.0.0-alpha | the failure | the failure |

as an error when the option is on, as a warning when it is off.

On 18 → 22 the core still adds its own *"The file was successfully generated"* next to it: the PDF was produced, the e-invoice was not, and nothing in the module can change what the core says there.

## Shape of the fix

`setEventMessages()` where the failure is raised, the way the three checks above it in the same function already do (configuration error, configuration warning, recipient reachability). The arrays are left untouched: they are how a hook reports to its caller, which is not always a browser.

It is not said again where the core prints it on its own — `->errors`, from 23 up — so the message is not doubled there. It still is on the "Generate document" button from 24, where the core prints `->warnings` as well: the duplication the recipient reachability warning already has, and the price of the validation path being covered. Happy to narrow it further if you would rather.

## Test

- Bench, eight cores (18.0.10, 19.0.4, 20.0.4, 21.0.4, 22.0.5, 23.0.4, 24.0.1, 25.0.0-alpha), both buttons, both modes of the option, over a real HTTP session: the tables above.
- PHPUnit grid, one file at a time per instance, same eight cores: 264/264, same as `main` in the same bench state.
- phpcs with the repository ruleset green.

Independent of #941 and #960, which fix two other reasons the same failure goes unseen; the three merge cleanly in any order (`git merge-tree`), and the measurement above was run again with all three applied.

Versions: 18.0.0 to 25.0.0.


daGrumpf-bxp:
Screenshots from the bench, real sessions, same refused generation on both sides.

**Dolibarr 22.0.5, "Generate document"** — on `main` the page answers *"The file was successfully generated"* while the e-invoice failed and the status stays "Not generated": the core reports a success because `pdf_sponge` returns 1 after a hook that answered -1. With the branch, the core still reports its PDF — which it did produce — and the module says what did not happen, in orange.

Dolibarr 22

**Dolibarr 24.0.1, validation** — the path that matters, since that is when the e-invoice is produced. On `main`, the invoice is validated, the e-invoice fails, and nothing at all is said. With the branch the failure is said: orange when "cancel if it fails" is off, red when it is on — and only once, the message being skipped where the core already prints `->errors` on its own (23 and up).

Dolibarr 24, validation

(Images hosted on a branch of my fork, so they are not part of this pull request.)

daGrumpf-bxp:
Screenshots from the bench, real sessions, same refused generation on both sides.

**Dolibarr 22.0.5, "Generate document"** — on `main` the page answers *"The file was successfully generated"* while the e-invoice failed and the status stays "Not generated": the core reports a success because `pdf_sponge` returns 1 after a hook that answered -1. With the branch the core still reports its PDF, which it did produce, and the module says what did not happen, in orange.

Dolibarr 22.0.5, before and after

**Dolibarr 24.0.1, validation** — the path that matters, since that is when the e-invoice is produced. On `main` the invoice is validated, the e-invoice fails, and nothing at all is said. With the branch the failure is said: orange when "cancel if it fails" is off, red when it is on, and only once, the message being skipped where the core already prints `->errors` on its own (23 and up).

Dolibarr 24.0.1 validation, before and after

PR #964, opened 2026-09-15
QUAL: [dev] PHPStan answers on the oldest core in a pull request too

daGrumpf-bxp:
Companion of #963, which fixes the error itself. This one is about how the error reached `main`.

Since 20f861eb the matrix is:

| event | cores analysed |
|---|---|
| pull request | `24.0.0` |
| push to `main` | `18.0.0`, `24.0.0` |

So an error that only the oldest core reports cannot be seen while the branch is open. That is not a
hypothesis: #962 was green on its pull request, and the merge turned `main` red on
`Property Societe::$idprof2 (string) on left side of ?? is not nullable` — a message Dolibarr 24 does
not produce, because it declares the property `?string` where Dolibarr 18 declares it `string`. The two
merges that followed (#961, #941) inherited the red without having anything to do with it.

Catching the drift between core versions is the reason this job runs a matrix at all, so the pull
request gets the oldest core back:

```diff
-          review='["24.0.0"]'
+          review='["18.0.0","24.0.0"]'
```

The `all` list is left as it is, so the push matrix can be widened again without touching this.

**What it costs.** One job of about a minute, on a pull request that touches `einvoicing/` — the paths
filter already keeps the workflow off every other pull request. PHPStan indexes the core through
`scanDirectories` and analyses the module only, so a second core does not double the analysis: it
repeats it against another set of signatures. The two jobs run in parallel, so the wall clock of the
check does not move.

If the minute is not worth it, the alternative is the state of the last day: an error on the oldest
core is found on `main`, by whoever notices the red, after the branch is gone. `main` stayed red for
three merges before anyone looked.


daGrumpf-bxp:
The new job is red on this pull request, and on purpose: the branch is cut from `main`, where the error is still there.

```
buildinvoicelines.inc.php: nullCoalesce.property:
  Property Societe::$idprof2 (string) on left side of ?? is not nullable.
```

That is the one error #963 removes, and the only thing this pull request changes is which cores the
matrix runs. So the red is the check reporting what `main` carries today - which is the whole argument
for the change.

Merging #963 first is the simple order. It will not repaint this page by itself, though: GitHub does
not replay the checks of a pull request when its base moves, so this one keeps showing the stale red
until the branch is pushed again. I will rebase it on `main` as soon as #963 is in, which reruns the
matrix and turns it green. Nothing to do here in the meantime.

daGrumpf-bxp:
Update, 2026-09-22 — my comment above is out of date and I would rather say so than leave it
standing. The error it names (`idprof2` in `buildinvoicelines.inc.php`) was removed by #963 a week
ago. The red on this pull request today is a **different** error, and it is still not mine: `main`
has been red on `PHPStan on Dolibarr 18.0.0` since 01:29 UTC this morning, from `empty($supplierInvoice->thirdparty)`
in `CIIProtocol::createSupplierInvoiceLinesIntoDatabase()`. #1052 removes it. Merge #1052 first,
rebase this one, and the job it adds goes green.

### Why this one line is worth merging

Since the pull request matrix was cut to `24.0.0` on 2026-09-14, **four** pull requests have been
needed to bring `main` back to green on Dolibarr 18, every one of them found after the merge:

| | | |
|---|---|---|
| #963 | 2026-09-15 | `$seller->idprof2 ?? ''` |
| #1008 | 2026-09-17 | same pattern, second site, `idprof1` |
| #1025 | 2026-09-19 | baseline for the two files #896 added |
| #1052 | 2026-09-22 | `empty($supplierInvoice->thirdparty)` |

Four of them, and not one was catchable at review time: each pull request was green 10/10, because
the check that would have caught it does not run on pull requests. The cause is always the same
shape — a property whose PHPDoc in the core became nullable after 18 — but the shape keeps moving,
`?? ''` then `empty()`, so it cannot be caught by a reviewer's habit either.

Two more things worth knowing:

- **PHPStan is the only workflow with this gap.** `module_einvoicing_phpunit.yml` already runs
  `18.0.0` / PHP 7.4 on pull requests. `phan.yml` has no matrix. This one line closes it.
- **The comment right above the line already promises this.** It reads "a pull request answers on
  the two ends of the supported range", and has since it was written. The file contradicts itself
  today; this makes the code agree with its own documentation again.

Cost: one extra job per pull request, on the cheap half of the matrix — PHPStan indexes the core, it
does not analyse it. Roughly a minute, against four after-merge debugging sessions in eight days.

daGrumpf-bxp:
@eldy could we put Dolibarr 18 back in the pull request matrix here? Four pull requests were green 10/10 and turned `main` red on 18 right after the merge (#963, #1008, #1025, #1052) — the one job this line adds would have caught all four at review time.

daGrumpf-bxp:
To be clear about the red above, since it is the reason not to merge this: it is `main`'s, not this pull request's. The job this line adds is the only one in the repository that can see it, and what it reports is byte-for-byte the three errors `main` carries since #1048 (`CIIProtocol`, `CommonProtocol`). **#1052 removes them, is one line and is green 10/10** — merge that one first and this goes green. I will re-run the checks here as soon as it lands, since GitHub does not re-run them by itself when the base moves.

PR #848, opened 2026-09-07
FIX #844 #850 [einvoicing] A received invoice line is imported at the amount it announces

daGrumpf-bxp:
Fixes #844.
Fixes #850.

## The rule

**BT-131 is what a line is worth.** The totals of a document are summed from it (BR-CO-10, BR-CO-13) and **no rule of EN 16931 ties it to quantity × price** — the item net price (BT-146) only says how the issuer priced the line. The import rebuilt each line from that couple and kept the result, so a line the issuer prices otherwise than it bills it came in at the wrong amount, and so did the invoice.

Two documents, one cause:

- **#844** — metered consumption, 1 951 593 units really at 0.00000548. A unit price is written with at most **6 decimals** (`BR-FR-DEC-03`, fatal, French Schematron), so the issuer writes `0.000005` and the rebuild lands on **9.76** where the line announces **10.69**. The document is conformant; only the printed precision of the price is at fault.
- **#850** — a bank fee, printed on the PDF as *quantité 50 005,00 € / P.U. 0,5999 %* and written in the XML as a quantity of `0.5999` against a price of `50005.000000`, announcing the fee itself, **300.00**. Dolibarr created the invoice at **50 005.00**.

### What the rule set actually says

Checked against the executable schematrons, not from memory. In annex A of XP Z12-012 (the normative file, EUPL, shipped in the FNFE repository) **BT-131 carries three rules**: `BR-24` (present), `BR-FR-DEC-01` (2 decimals) and `BR-FR-MV-10` (the VAT-inclusive amount of a `GROUP` line). BT-146 carries `BR-26`, `BR-27`, `BR-FR-DEC-03`, and one stated relation only: *net price = gross price − item price discount*. The CEN schematron agrees: on `ram:LineTotalAmount` there is nothing but BR-24, BR-DEC-23, `BR-CO-10` and the VAT base rules `BR-*-08`. **No rule computes BT-131 from quantity and price** — while `BR-CO-10` (fatal) makes BT-131 the one line figure the totals of the document are summed from.

Run through the full FNFE chain (XSD + profile XSLT + BR-FR), the reported document of #850 rebuilt as EN 16931 comes out **VALID** (97 profile checks, 67 CTC-FR checks, no failure), and the reporter's own EXTENDED XML fails only on the identifiers they redacted. It is a conformant document, and it was imported at 167× what it bills.

[...]

PR #967, opened 2026-09-15
FIX [einvoicing] A line whose net amount is entirely absorbed by its own charge is no longer counted twice

lelex86:
## The defect

`resolveLineAmounts()` decides whether a received line's (quantity, unit price) couple can be trusted, or whether the line must be imported at the amount BT-131 announces instead (#844, #850). The guard on both branches is `!empty($announced)`, where `$announced` is BT-131 *minus the line's own BG-28 charges* — the part of the line's declared total the base Dolibarr line has to carry once the charges have left to become lines of their own (#735):

```php
$announced = round((float) ($parsedLine['lineTotalAmount'] ?? 0) - $charges, 2);
...
if (!empty($announced) && ($rebuilt == 0.0 || (($rebuilt > 0) !== ($announced > 0)))) {
    ...
}
...
if (!empty($announced) && $difference > 0.01 && $divisor != 0.0) {
    ...
}
```

`!empty($announced)` cannot tell apart two different situations that both leave `$announced` at `0.0`:

* **BT-131 is absent.** The line announces nothing — a free sample, a heading. Correctly left alone (the whole point of the guard, per #848: *"A line announcing nothing is left alone: there is no amount to import it at."*).
* **BT-131 is present, and equals the line's own charge.** The line's net price, once its BG-28 charge is set aside, is genuinely `0.00` — but the line very much announces an amount, all of it carried by its charge.

The second case reaches neither branch. The base line keeps whatever (quantity, unit price) it was built with instead of being corrected to `0.00`, and `buildLineChargeLines()` then adds the charge a second time as its own line. The line is counted twice.

## Measured, from a real flow

A FedEx invoice received through Esalink, Factur-X EXTENDED. Line 1, unabridged but for the delivery address:

```xml
<ram:SpecifiedTradeProduct>
  <ram:SellerAssignedID>FedEx International Priority</ram:SellerAssignedID>
  <ram:Name>876538539902</ram:Name>
  <ram:Description>FedEx International Priority</ram:Description>
  <ram:OriginTradeCountry>
    <ram:ID>CA</ram:ID>
  </ram:OriginTradeCountry>
</ram:SpecifiedTradeProduct>
<ram:SpecifiedLineTradeAgreement>
  <ram:GrossPriceProductTradePrice>
    <ram:ChargeAmount>15</ram:ChargeAmount>
    <ram:BasisQuantity unitCode="EA">1</ram:BasisQuantity>
  </ram:GrossPriceProductTradePrice>
  <ram:NetPriceProductTradePrice>
    <ram:ChargeAmount>15</ram:ChargeAmount>
    <ram:BasisQuantity unitCode="EA">1</ram:BasisQuantity>
  </ram:NetPriceProductTradePrice>
</ram:SpecifiedLineTradeAgreement>
<ram:SpecifiedLineTradeDelivery>
  <ram:BilledQuantity unitCode="EA">1</ram:BilledQuantity>
</ram:SpecifiedLineTradeDelivery>
<ram:SpecifiedLineTradeSettlement>
  <ram:ApplicableTradeTax>
    <ram:TypeCode>VAT</ram:TypeCode>
    <ram:CategoryCode>S</ram:CategoryCode>
    <ram:RateApplicablePercent>20</ram:RateApplicablePercent>
  </ram:ApplicableTradeTax>
    <ram:ChargeIndicator><udt:Indicator>true</udt:Indicator></ram:ChargeIndicator>
    <ram:ActualAmount>15</ram:ActualAmount>
    <ram:Reason>Frais de dossier</ram:Reason>
  <ram:SpecifiedTradeSettlementLineMonetarySummation>
    <ram:LineTotalAmount>15</ram:LineTotalAmount>
  </ram:SpecifiedTradeSettlementLineMonetarySummation>
</ram:SpecifiedLineTradeSettlement>
```

Net price 15, billed quantity 1, one line charge of 15 ("Frais de dossier"), `LineTotalAmount` (BT-131) 15. `$charges` = 15, `$announced` = `15 - 15` = `0.0`. The couple (1, 15) rebuilds 15, which is neither `0.0` nor of opposite sign to `$announced`, and `!empty(0.0)` is `false` either way — both branches are skipped, `resolveLineAmounts()` returns the untouched `(qty=1, subprice=15)`.

`createSupplierInvoiceLinesFromSource()` then adds `buildLineChargeLines()`'s own line for the same 15 charge. The document's second line (net price 5, one 5 charge, `LineTotalAmount` 5) repeats the pattern. Header `LineTotalAmount` (BT-106) is 20; the invoice this module builds sums to 40 — exactly double.

## The fix

Test whether the document wrote BT-131 at all, not whether the result of subtracting the charges from it happens to be zero. `parseInvoiceLines()` only casts a field `if (isset($line[$f]))`, and `getXPathValue()` answers `null` for an absent element — so `$parsedLine['lineTotalAmount']` stays `null` when the document never carried BT-131, and becomes a real `0.0` float when it did. `isset($parsedLine['lineTotalAmount'])` tells the two situations apart where `!empty($announced)` could not:

```php
$lineAnnouncesAnAmount = isset($parsedLine['lineTotalAmount']);
...
if ($lineAnnouncesAnAmount && ($rebuilt == 0.0 || (($rebuilt > 0) !== ($announced > 0)))) { ... }
...
if ($lineAnnouncesAnAmount && $difference > 0.01 && $divisor != 0.0) { ... }
```

On the FedEx line above, `$lineAnnouncesAnAmount` is now `true`, the sign-mismatch condition of the first branch fires (`$rebuilt` is `15 > 0`, `$announced` is `0`, not `> 0`), and the line is imported at `(qty=1, subprice=0)`. `buildLineChargeLines()` adds its 15, the line totals 15 as the document announces, and the invoice sums to 20 instead of 40.

A line that genuinely announces nothing is unaffected: `isset()` answers `false` exactly as `!empty(0.0)` already did, so the free-sample/heading case #848 wrote the guard for keeps working the same way.

## Scope

Any received line whose net amount, once its own BG-28 charge is set aside, is exactly `0.00` — a line entirely billed as a single flat fee, with no separate priced item. Measured here on Factur-X/FedEx; the same parsing and the same `resolveLineAmounts()` serve the CII path since #939.


daGrumpf-bxp:
This also closes **#969**: same guard, second real document — FedEx again, received as UBL 2.1 through another platform, with the charge split in two (7.09 + 31.65 = BT-131 of 38.74) instead of carried by one.

Bench, Dolibarr 24.0.1, imports and validations rolled back, four documents:

| document | `main` (39847821) | + this pull request |
|---|---|---|
| A — one charge = the whole line (#967, 15 + 5) | 48.00 for 24.00 announced, **validation refused** | **24.00**, validated |
| B — two charges summing to BT-131 (#969, real FedEx) | 176.31 for 88.15, **refused** | **88.15**, validated |
| C — conformant line charge (BT-131 = price + charge) | 192.00, validated | **192.00**, validated — unchanged |
| D — line announcing no BT-131 at all (#848) | 120.00, silent, validated | **120.00**, silent, validated — unchanged |

So both broken shapes are repaired and the two cases that must not move do not. `JOBS=8 punitj.sh`: **272/272** across the eight cores (18.0.5 → 25.0.0-dev).

One note: the line count is unchanged — 6 lines for a 2-line invoice, the shipment lines at 0.00 and the charges on their own. The amount is right and the invoice validates, which is what both reports are about; folding a breakdown whose sum is exactly BT-131 back into its line is a separate question.

lelex86:
Thanks for the cross-check against #969 — good to know the guard holds on a second real report, in a different syntax, without us having seen that shape ourselves. The bench across the eight cores and the two untouched controls (C, D) is reassuring.

You're right about the line count, and we did think about it too — but that split is #735's own design (a charge rejoins the invoice as a line of its own), so we left it alone rather than fold it into this PR.

PR #970, opened 2026-09-15
FIX The French label of the 'issued' status no longer reads "l'a PA"

daGrumpf-bxp:
The French label of the "issued by the seller's access point" status reads `Émis par l'a PA du vendeur`.

The stray `l'` is a leftover of the previous wording: the string used to be `Émis par le vendeur AP`, and when it was rephrased to name the access point, `le vendeur` became `l'a PA`. The neighbouring statuses were rephrased correctly (`EInvStatus202Received=Reçu par la PA du client`, `EInvStatus213Rejected=Rejeté par la PA du vendeur`), so this is the only line affected.

Reported by a user of the module, who sees it on the e-invoicing status of a customer invoice.

I know `fr_FR` is normally left to Transifex and kept out of pull requests; this one line is proposed anyway so that the typo does not stay visible in the meantime. Feel free to drop it if the next Transifex sync is close, the same correction is needed there in any case.

Tested on the bench (Dolibarr 20, instance `dd20`, module deployed from this branch): `$langs->setDefaultLang('fr_FR'); $langs->load('einvoicing@einvoicing'); $langs->trans('EInvStatus201Issued')` returns `Émis par la PA du vendeur`, and returns `Émis par l'a PA du vendeur` on `main`.


aspangaro:
Already correct in transifex. See in the next synchronisation. You can close your PR.

image

daGrumpf-bxp:
Thanks for checking — good to know Transifex already has the right wording. Closing, the next sync will carry it.

PR #982, opened 2026-09-16
FIX: [einvoicing] The default product of a vendor hides the product mapping

daGrumpf-bxp:
`findProductFromEinvoiceLine()` ends on the default product of the vendor, a catch-all that answers every line nothing was found for, whether or not that line carries a vendor reference.
The mapping page only read `res`, never the `matchtype` the matching returns, so declaring a default product made every line of that vendor show as resolved: the product select disappeared and the page announced "All the lines of this flow are linked to a product".
A routed line that carries a vendor reference is now offered again for mapping, and a line that carries none shows where it will go without being counted, since there is nothing to record an association on.
The page also offers the same "Set a default product for this third-party" link the synchronization already suggests, which its own message told the user to go and do without taking them there.
That link, and the two the synchronization builds, are gated on the right to edit a third party; `document_list.php` only prints the HTML the protocol hands it, so the guard belongs where the buttons are built.

### Decision table, product column of the mapping page

| Line | Vendor default product | Before | After |
|---|---|---|---|
| reference, no match | none | select | select (unchanged) |
| no reference, no match | none | "Cannot be mapped" | "Cannot be mapped" (unchanged) |
| reference, no match | declared | catch-all shown as resolved, no select | "Falls back on the default product: X" **and** the select |
| no reference, no match | declared | catch-all shown as resolved | "Falls back on the default product: X" + "Cannot be mapped" |
| reference, mapped | either | resolved product | resolved product (unchanged) |

| Counter "Number of lines to map" | Before | After |
|---|---|---|
| line routed to the catch-all, with a reference | not counted | counted, it can still be mapped |
| line routed to the catch-all, without a reference | not counted | not counted, nothing to map on |

Bench, Dolibarr 23, a two line flow whose first line carries a vendor reference: with a default product declared, `main` shows counter 0, no select and both lines on the catch-all; this branch shows counter 1, the select on the first line, and saving it binds the line to the real product and brings the counter back to 0. Page loaded without fatal or warning on Dolibarr 18, 20, 22, 24 and 25, the button present with the third party right and absent without it.


einvoicing/class/protocols/CommonProtocol.class.php
```diff
@@ -1111,11 +1111,15 @@ private function _syncOrCreateThirdpartyFromEInvoiceSeller($sellerInfo, $priorit
...
 			$message .= $langs->trans("AutoCreateThirdPartyOffCreateItManually");
 
+			// Creating the thirdparty needs the right on it, so a user who has not got it is told what
+			// to ask for instead of being sent to a page that answers accessforbidden().
 			$action = $langs->trans('CreateSupplierManually');
-			$action .= '<a class="butAction small smallpaddingimp" href="' . dol_escape_htmltag($createUrl) . '" target="_blank">';
-			$action .= '<i class="fas fa-plus-circle"></i> ';
-			$action .= $langs->trans('CreateSupplier');
-			$action .= '</a>';
+			if ($user->hasRight('societe', 'creer')) {
```
eldy: I would keep the button visible (or at least have it greyed).
Otherwise a user that has no permission does not understand he can't because of permission. He just think "Create ? ok but how ?".

daGrumpf-bxp: Agreed, and done in 5b33ca8 for the three buttons the PR had gated.

The one here is drawn as a `butAction`, so it becomes a `butActionRefused classfortooltip` with `href="#"` and `title=NotEnoughPermissions` — the shape the core uses, and a class the eldy and md themes both style from 18 to 25.

The two others ("Set a default product for this third-party", in the synchronization notice and on the mapping page) are drawn as a `button`, so they keep that shape and take `disabled` rather than `buttonRefused`: `.buttonRefused` lost its rule in the eldy theme in Dolibarr 24 (it only survives in md), so a button written that way looks enabled there, whereas `.button.disabled` is styled in both themes on all eight versions.

The `allactiondata` entry of the default product goes back to being built in every case: nothing renders it, and it describes an action, not a permission.

Bench, Dolibarr 18 / 23 / 25, with a non-admin user and a real received document replayed through the import: without the right on third parties the three buttons are present and greyed, titled "You don't have permission for this action"; with it they are the plain buttons again. Mapping page loaded with 0 error and 0 warning in both states.

PR #987, opened 2026-09-16
NEW: [einvoicing] A received document keeps the files its issuer embedded

### LINKED ISSUE #980: Also get supplier PDF - SFR - only getting e-invoice and SuperPDP specific PDF view
### Feature Request

In my SuperPDP userinterface I can find a PDF from the supplier with details.

The CII information converted to Dolibarr data only has the summary line, while the XML holds detail:

Image

Image

That document also holds detail:

Image

Currently I need to check SuperPDF manually, download and attach this extra original document.

### Module

e-invoicing

daGrumpf-bxp:
Fixes #980.

## What was wrong

The invoice SFR sends through the access point carries **one summary line**, and its detail exists only in the PDF it embeds as **BT-125** (`AdditionalReferencedDocument` / `TypeCode 916` / `AttachmentBinaryObject`). That binary was received and stored inside the XML attached to the supplier invoice, but nothing here ever reads an attachment back out of a document, so it was unreachable: the operator had to open the access point interface, download the file and attach it by hand.

## What this does

- **Embedded files are extracted at import** and stored as attached files of the supplier invoice, named like the other imported files (`<ref_supplier>_<name>.<ext>`). Document level and line level alike.
- **The file kept on disk stays the document the access point returned, byte for byte.** What is archived beside the invoice has to be comparable to what arrived; the readability comes from the files extracted next to it, not from rewriting the document.
- **The copy in `llx_einvoicing_document.xml_data` stays scrubbed**, unconditionally, as `Document::cleanXmlData()` has done since #237. That column is a `MEDIUMTEXT` guarded at 16 Mo by `checkXmlDataMaxSize()` and one embedded PDF is enough to pass it - past the cap the row stores **nothing**, and the consistency check of #237 and the return address `CdarHandler` reads from that column both lose their source without a word.
- **`EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES`** (hidden, off) asks for the lighter shape on disk instead: the binary leaves the attached XML, and the note left in its place says which file holds it - `[Removed at import by CIIProtocol::removeAttachmentFromXml(), kept as attached file: lisible.pdf]`. Nothing is lost either way, that file sits right there.
- **The extension comes from the mime code the document declares**, which is also the allowlist: EN 16931 names six (pdf, png, jpeg, csv, xlsx, ods), and a code outside them is not extracted at all. The file name of the document is used as a label only, so a received document cannot choose the extension written on disk.
- A failed `ReadableView` call is **no longer silent**: it was made without a call type, so `logCall()` dropped it, and its error return was commented out. It is now a logged call (`get_readable_view_for_invoice`) and a `LOG_WARNING`. An access point with no readable view for a flow used to leave the invoice with the XML alone and no trace of why.

## Bench, Dolibarr 18 to 25

A received CII with a 419 608 bytes PDF embedded as BT-125, imported on each core under the PHP its CI uses (18/7.4, 19/8.0, 20/8.1, 21/8.2, 22/8.3, 23/8.4, 24 and 25/8.5) - 16 imports, two states, no fatal and no deprecation this branch does not already have on `main`.

| | file on disk | embedded PDF | `xml_data` in database |
|---|---|---|---|
| **default** | **576 476 bytes = the received document, byte for byte** | attached, same md5 as the embedded one | **17 014 bytes**, no base64 |
| `EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES = 1` | **17 081 bytes**, the note in place of the binary | attached, same md5 | **17 014 bytes**, no base64 |

Counter-proof on unpatched `main`, same document, same core: the XML on disk is already the received one byte for byte, the PDF is **not** extracted at all, and the deprecation count is the same 4 - so what this branch changes is that the annex arrives, nothing else.

Hostile and degraded shapes checked one by one: path traversal in the name, `.php` in the name, executable mime code, no mime code, upper case mime code, broken base64, an already scrubbed node, an empty node, no name at all, two files with the same name, a line level attachment, and a document that is not XML. Only the legitimate ones are extracted.


daGrumpf-bxp:
Bench proof, Dolibarr 22.0.5, the same received document imported twice: once on `main`, once on this branch.

980-before-after

Left to right in the XML stored next to it:

| | on `main` | on this branch |
|---|---|---|
| attached files | 1 | 2 |
| total size | 110 Kb | 86 Kb |
| `AttachmentBinaryObject` holds | `JVBERi0xLjcKJeLjz9MKMjQgMCBvYmoNCjw8...` | `[Removed at import by CIIProtocol::removeAttachmentFromXml(), kept as attached file: lisible.pdf]` |

The stored `lisible.pdf` has the same md5 as the file taken out of the received document, and it opens from the invoice card: it is the issuer's own annex, the one that carries the detail the summary line does not.

Run on the 8 cores of the bench, each on the PHP version its CI uses - 18/7.4, 19/8.0, 20/8.1, 21/8.2, 22/8.3, 23/8.4, 24 and 25/8.5. Same result on all eight, no PHP error. Re-importing a document that was imported before this change catches it up: the PDF appears and the XML is replaced by the scrubbed one.

Shapes checked one by one on the extraction itself:

| shape | result |
|---|---|
| `filename="../../../../var/www/evil.php"`, mime `application/pdf` | stored as `evil.pdf` |
| `filename="payload.php"`, mime `application/pdf` | stored as `payload.pdf` |
| mime `application/x-httpd-php` | nothing extracted |
| no mime code | nothing extracted |
| mime code in upper case | extracted |
| broken base64 / empty node / an already scrubbed node | nothing extracted |
| no file name at all | falls back on BT-122 |
| two attachments with the same name | second one suffixed |
| attachment at line level | extracted |
| not XML at all | nothing extracted |

daGrumpf-bxp:
Added `EINVOICING_KEEP_RECEIVED_XML_AS_IS`, for whoever wants the received document archived exactly as it arrived. The scrub loses nothing - the binary is the file stored right next to the XML - but it does mean the attached XML is no longer the bytes the access point returned, and that is a call the operator should be able to make.

Same document, imported twice on Dolibarr 22:

| | default | `EINVOICING_KEEP_RECEIVED_XML_AS_IS = 1` |
|---|---|---|
| attached XML | 14 896 bytes, the note in place of the binary | 112 715 bytes, the base64 untouched |
| md5 of that XML vs the received document | differs (scrubbed) | **identical** |
| `lisible.pdf` attached | yes | yes |

The option only changes what the XML keeps: the embedded files are stored as attached files either way.

Run on the 8 cores with the option set - 18/7.4, 19/8.0, 20/8.1, 21/8.2, 22/8.3, 23/8.4, 24 and 25/8.5:

```
dd18   php7.4  xml_identical=YES pdf_attached=1 php_errors=0
dd19   php8.0  xml_identical=YES pdf_attached=1 php_errors=0
dd20   php8.1  xml_identical=YES pdf_attached=1 php_errors=0
dd21   php8.2  xml_identical=YES pdf_attached=1 php_errors=0
dd22   php8.3  xml_identical=YES pdf_attached=1 php_errors=0
dd23   php8.4  xml_identical=YES pdf_attached=1 php_errors=0
dd24   php8.5  xml_identical=YES pdf_attached=1 php_errors=0
dd25   php8.5  xml_identical=YES pdf_attached=1 php_errors=0
```

The option is listed in `doc/function-map.data.json`. The two files that map generates are left alone here: they were last built on 05ffc589 and rebuilding them now renumbers every symbol and surfaces two `*missing*` entries that come from renames in other merges - it belongs to its own regeneration commit, as the previous one was.

eldy:
My first answer was 
"I think the default behaviour should be the most conservative, the less dangerous.
So i would revert the option 
EINVOICING_KEEP_RECEIVED_XML_AS_IS  
into
EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES"

However, i have a doubt. It seems on your test before patch, that you receive xml alone without any reable file ?
How can this happen ?   If this can happen, option as it is is welcome because having a readable version is very important, howerver, i don't understand why no readable version is sotre next to the .xml currently (it should) ?

daGrumpf-bxp:
Both questions are fair, and the second one has an answer that changes how the first should be decided.

## Since when the received XML loses its embedded files

Not since this PR: since **b89502ff, 18 June 2026**, merged the same day by **#237**. That commit added `Document::cleanXmlData()` and `AbstractProtocol::removeAttachmentFromXml()`, and `CIIProtocol` implements it by replacing the value of **every** `AttachmentBinaryObject` with `[Removed to get a smaller XML]`.

It runs on two paths, both on `main` today:

- `SuperPDPProvider::syncFlow()` and `EsalinkPDPProvider::syncFlow()`, on the XML about to be written in `llx_einvoicing_document.xml_data`;
- `AbstractPDPProvider::fetchFlowXml($flowId, true)`, used by `SupplierInvoiceHelper::getXmlData()` when that column is empty and the document is fetched again.

Every published version ships it: `einvoicing_1.0.3` (29/07), `1.1.0` (30/08), `1.2.0` (08/09). So the behaviour before this PR was already "strip", and it stripped with **nothing kept anywhere**: the binary was dropped and unreachable, which is exactly what #980 reports.

That is why the option is written in that direction — `EINVOICING_KEEP_RECEIVED_XML_AS_IS` reverses a default that was already there, it does not introduce one. What the PR changes is that the scrub is no longer a loss: the file is decoded and stored as an attached file of the supplier invoice **before** the node is emptied, and the note left in its place names the file that now holds it.

For the record, the reporter of #980 asks for the opposite of the current default on disk: *"Stripping parts from the xml before storing it is IMHO an issue as we do not have a local copy that is exactly what was provided. Could strip for a database, but it should still be saved on disk IMHO."*

## "You receive the XML alone, without any readable file?"

Not in his installation, and my screenshot was misleading on that point — my apologies. He does have the readable view: *"I do have a `<ref_supplier>_SUPERPDP.pdf`, but this is in SUPERPDP's format and has 2 pages."* The before/after above is a **bench replay** of his document, with no access point behind it, so no `ReadableView` is fetched and only one file lands — it says nothing about what a real reception attaches.

What is missing in #980 is not the readable view, it is the **issuer's own annex**: SFR's 3-page `lisible.pdf`, which travels inside the invoice as BT-125. Nothing in this module ever read a binary back out of a document, so it was received and thrown away.

That said, **XML-alone can genuinely happen**, and it is worth fixing whatever you decide on the default:

- the readable view is requested only when the detected protocol is not Factur-X (a Factur-X document is a PDF already, so that part is fine);
- a non-200 on that call was **completely silent** on `main`: the error return is commented out (*"We disable this error, getting the readable file is optional"*) **and** the call was made without a `callType`, so `logCall()` returns at its first line (`if (empty($callType)) return null;`) and nothing reaches `llx_einvoicing_call` either.

So an access point that has no readable view for a flow, or answers an error, leaves the invoice with the XML alone and no trace at all of why. This PR gives that call a type (`get_readable_view_for_invoice`) and a `LOG_WARNING`; the import still carries on without it.

One more thing this flow proved, after mdeweerd sent the Original: on his document the binary is **emptied by the access point conversion**, not by us — the raw response of `GET /flows/i_601739?docType=Converted`, stored verbatim in his call log, already carries `<AttachmentBinaryObject mimeCode="application/pdf" filename="lisible.pdf"></AttachmentBinaryObject>`, while the UBL Original holds the 115 316 characters of base64. The `[Removed to get a smaller XML]` he saw in his support export is our note written on an already empty node.

## On the default — I have no attachment to the one in the PR

The extraction itself is the fix and should stay unconditional in any variant: the embedded files become attached files of the supplier invoice, always. Only the fate of the stored XML is up for decision, and the three shapes are:

1. **As it stands**: scrub by default, `EINVOICING_KEEP_RECEIVED_XML_AS_IS` to keep the disk copy byte for byte.
2. **Your proposal**: keep byte for byte by default, `EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES` to scrub. That is the conservative default, and it is also what the reporter asks for.
3. **The one I would pick**: **as-is + extract as the default, no option to set at all** — the archived file stays exactly what the platform returned, and the readability comes from the files extracted next to it rather than from rewriting the document. Whoever wants the small XML gets the option instead.

The only argument left for scrubbing is size, and it applies to the database column rather than to the disk: `xml_data` is a `MEDIUMTEXT` guarded at 16 Mo by `checkXmlDataMaxSize()`, and on his document the scrub takes the XML from 110 Kb to 15 Kb. So I would keep `cleanXmlData()` on the column exactly as #237 wrote it, and stop rewriting the file.

Tell me which of the three you want and I push it — 2 and 3 are both a small change on this branch, and 3 is what I would commit if the choice is mine.

daGrumpf-bxp:
**Why that scrub exists** — it landed in **#237**, four days before my first commit here, and it was a review remark, not a design choice: spsolauv added `xml_data` so the consistency check could compare the Dolibarr invoice to the e-invoice *"even if the AP is offline"*, the attached file being deletable by the user; @Daoud-mohamed answered *"Since e-invoice XML can embed base64 PDF attachments, we may need a cleanup function to remove attachment nodes before saving the XML"*; spsolauv added `removeAttachmentFromXml()` the next morning — *"I just replace binary content by a string to not lose attachments metadata"* — and wrote on the property itself that this copy *"may not be identical to original XML content from AP file […] to preserve XML size stored in database"*. The column is a `MEDIUMTEXT` capped at 16 Mo by `checkXmlDataMaxSize()`: the scrub is a storage constraint, never an archiving decision.

**So my proposal is to leave the default where it is**, and I would rather not reverse it:

- **scrub by default, silently**, because of that insert in database;
- **the embedded files extracted automatically** and attached to the supplier invoice, which is what this PR is for;
- **`EINVOICING_KEEP_RECEIVED_XML_AS_IS` kept as a supplement**, for whoever wants the file on disk to be the full document as the platform returned it;
- and one rule that holds in every case: **the option never reaches the database column**. It stays scrubbed whatever the option says, or the databases blow up.

That last point is what I have just pushed (comments and function map only, no behaviour changed) — it was true in the code, it was written nowhere.

## Bench, the two states, 8 cores

Document built from a received CII with a 419 608 bytes PDF embedded as BT-125, imported on each core under the PHP its CI uses (18/7.4, 19/8.0, 20/8.1, 21/8.2, 22/8.3, 23/8.4, 24 and 25/8.5) — 16 imports, no PHP notice anywhere.

| | file on disk | embedded PDF | `xml_data` in database |
|---|---|---|---|
| default | **7 677 bytes**, the note in place of the binary | attached, same md5 as the embedded one | **7 610 bytes**, no base64 |
| `EINVOICING_KEEP_RECEIVED_XML_AS_IS = 1` | **567 060 bytes = the received document, byte for byte** | attached, same md5 | **7 610 bytes**, no base64 |

And what the cap does when the issuer embeds something big — same document with a 13 Mo attachment:

```
received document           : 17 900 388 bytes
fits the column as received : FALSE - nothing stored
scrubbed                    : 7 606 bytes
fits the column scrubbed    : true
```

So an unscrubbed column is not merely heavy: past 16 Mo the row stores **nothing**, and the consistency check of #237 and the return address `CdarHandler` reads from that column both lose their source without a word.

eldy:
When i merged https://github.com/Dolibarr/dolibarr-community-modules/commit/b89502ff011f99d25d4a9779bcd2d6cd82a7bf8c, i probably made an error.
I thought, the stripping of xml file was to store the xml in database, not to strip the file stored on disk.

The target is :
- When we received a file, we must store the file on disk as it was received (in the format we set on the AP platfom). And if file is not PDF (often as default that should be setup is cii), we must also get the readable version (We don't need to store the "original" for the moment, we could download it or view it on the fly in a future version).
- But to store the file in db, we must strip the XML from any PDF content (goal of this data is to make debug, make indexing in future, search, etc...) and we need a light version string for this. This is not the refernece but a denormalized value (it can be lost, corrupted, etc..). File on disk must remain similare from the one received from the AP.
- Option EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES can exist if we want to make the xml file lighter on disk (but i don't see interesting use case that need this).

daGrumpf-bxp:
Done, pushed in `0e046f85` — the three points of your message, in that order.

**1. The file on disk is the one received.** It already was on `main`: `$tempFile` is written from the raw response and attached as it stands, and nothing in the module rewrote it. This branch was what started rewriting it, and that is now gone. So the default is back to yours: the attached file is the document the access point returned, embedded binaries included.

**2. The copy in database is scrubbed.** Unchanged and unconditional — `Document::cleanXmlData()` on the `xml_data` column, exactly as #237 wrote it. Your reading of that commit is the right one, and it is the only place the scrub belongs: the column is a `MEDIUMTEXT` capped at 16 Mo, it is a denormalized value for debugging and future indexing, and the file on disk is the reference.

**3. The option exists, in your direction.** `EINVOICING_KEEP_RECEIVED_XML_AS_IS` is replaced by `EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES`, hidden and off by default. I agree there is no compelling use case; it costs one `if` and answers the reporter of #980 the day someone does want the small XML on disk.

And the readable version is the part of your first bullet that was actually broken: the `ReadableView` call had its error return commented out **and** no call type, so `logCall()` dropped it at its first line. It is now a logged call and a `LOG_WARNING`, so an access point that has none for a flow leaves a trace instead of an invoice with the XML alone and no explanation.

## Bench, the two states, 8 cores

Same document as before — a received CII with a 419 608 bytes PDF embedded as BT-125 — imported on each core under the PHP its CI uses (18/7.4, 19/8.0, 20/8.1, 21/8.2, 22/8.3, 23/8.4, 24 and 25/8.5). 16 imports, no fatal.

| | file on disk | embedded PDF | `xml_data` in database |
|---|---|---|---|
| **default** | **576 476 bytes = the received document, byte for byte** | attached, same md5 as the embedded one | **17 014 bytes**, no base64 |
| `EINVOICING_SPLIT_XML_WITH_EMBEDDED_PDF_IN_TWO_FILES = 1` | **17 081 bytes**, the note in place of the binary | attached, same md5 | **17 014 bytes**, no base64 |

And the counter-proof, the same document on unpatched `main`, same core:

```
xml sur disque : 576 473 octets, identique au recu : OUI
fichiers joints: QA987-BASEMAIN-1_einvoice.xml        <- et rien d'autre
php_errors=4                                          <- le meme compte que la branche
```

So on `main` the archived file is already faithful, the annex is simply thrown away, and the deprecation count is identical on both sides: what this branch changes is that the annex arrives. (Those 4 are pre-existing — `trim(null)` at `CIIProtocol::createSupplierInvoiceLinesFromSource()`, `$mysoc` null under `NOREQUIRESOC` in the harness, and two `SupplierInvoiceLine::$libelle` dynamic properties from the core. None is touched here.)

PR #988, opened 2026-09-16
NEW: [einvoicing] The XML of the "Documents" tab gets its preview picto too

daGrumpf-bxp:
Follow-up of #687, which asked for a preview of the e-invoice XML. That preview shipped in #696, but only where it could reach: the **card**, whose file list is drawn by `FormFile::showdocuments()` and has a line hook, `formBuilddocLineOptions`. The **"Documents" tab** was left out, and that is the tab people open to look at the files of an invoice.

## Why the tab had no insertion point

It goes through `FormFile::list_of_documents()`, and that method has exactly one hook:

```php
$reshook = $hookmanager->executeHooks('showFilesList', $parameters, $object);
if (!empty($reshook)) {   // null or '' for bypass
    return $reshook;
}
```

It is all-or-nothing: returning anything replaces the whole list. And `executeHooks()` returns an `int`, so a module cannot even hand the replacement HTML back - it would have to `echo` a copy of the core's 250 lines. Verified identical on Dolibarr 18.0, 20.0, 22.0, 24.0 and develop: one hook, same guard, in all five.

## What this does instead

The picto is placed from `printCommonFooter`, which runs at line 3723 of `main.inc.php` while `lib_foot.js.php` is only loaded at 3731 - and that file binds `documentpreview` inside a `jQuery(document).ready()`. So the anchor is in the DOM before the core binds, and the core binds it like a native one. Measured: `jQuery._data(anchor, 'events')` holds `click` on all eight cores.

The hook prints directly rather than filling `resprints`, because `printCommonFooter()` of the core never prints the `resPrint` of this hook - it only reads the return value to decide whether to print its own block.

The line is found on the name of the file the module writes (`*_einvoice.xml` received, `*_cii.xml` emitted), so a tab with no XML is left untouched, and the card keeps the single picto it already had.

## ⚠️ Takes effect after a deactivate/reactivate

`invoicedocument` and `invoicesuppliercarddocument` are new hook contexts, and Dolibarr reads the contexts of a module from `llx_const` (`MAIN_MODULE_EINVOICING_HOOKS`), which only the activation rewrites. Updating the files is not enough. Verified: the cycle refreshes it on all eight cores.

## Bench

Dolibarr 18 to develop, each on the PHP version its CI uses (7.4, 8.0, 8.1, 8.2, 8.3, 8.4, 8.5, 8.5):

```
dd18 {"anchor":1,"inNameCell":1,"bound":1}   dd22 {"anchor":1,"inNameCell":1,"bound":1}
dd19 {"anchor":1,"inNameCell":1,"bound":1}   dd23 {"anchor":1,"inNameCell":1,"bound":1}
dd20 {"anchor":1,"inNameCell":1,"bound":1}   dd24 {"anchor":1,"inNameCell":1,"bound":1}
dd21 {"anchor":1,"inNameCell":1,"bound":1}   dd25 {"anchor":1,"inNameCell":1,"bound":1}
```

Clicking it opens the core's own preview dialog with the XML inside, on a received supplier invoice and on an emitted customer one alike. And the cases where nothing must happen:

| page | pictos added |
|---|---|
| tab of an invoice with no XML (a `_einvoice.pdf` only) | 0 |
| card of an invoice with an XML | 1, the one the line hook already placed |
| any page out of the two contexts | 0, the module is not even loaded |


daGrumpf-bxp:
Bench, Dolibarr 22, a received supplier invoice - the Documents tab on `main`, then the same tab on this branch, then what the picto opens.

988-tab

The picto lands in the name cell, where the core puts its own, and the dialog is the core's - not a page of the module.

PR #978, opened 2026-09-16
FIX: #974 [einvoicing] An exempt line only repeats its reason where the profile asks for it

daGrumpf-bxp:
Fixes #974.

## The defect

An invoice issued by a seller under *franchise en base de TVA* is refused by the access point:

```
<ram:ContentCode>REJ_COH</ram:ContentCode>
<ram:Content>Element 'ram:ExemptionReason' is marked as not used in the given context.</ram:Content>
<ram:Content>Element 'ram:ExemptionReasonCode' is marked as not used in the given context.</ram:Content>
```

`CIIProtocol::buildLineItem()` wrote `ram:ExemptionReason` and `ram:ExemptionReasonCode` in the tax
block of every invoice line, whatever the profile the document declares.

## Why the line copy exists, and why it cannot be unconditional

It exists for `BR-FXEXT-E-08`. That rule reconciles the taxable amount of an exempt VAT breakdown
(BT-116) with the net amounts of the lines it covers, and its selector counts a line only when the
line repeats **the same reason code and the same reason text**:

```xpath
count(… /ram:IncludedSupplyChainTradeLineItem[
    … and normalize-space(ram:SpecifiedLineTradeSettlement/ram:ApplicableTradeTax/ram:ExemptionReasonCode) = $exempReasonCode
    and normalize-space(ram:SpecifiedLineTradeSettlement/ram:ApplicableTradeTax/ram:ExemptionReason)     = $exempReasonText])
```

Written on the breakdown alone it counted zero lines and the document came back unbalanced. That is
what ac6dfcaa fixed, and it is only true of EXTENDED: the rule lives in `FACTUR-X_EXTENDED.sch` and
`EXTENDED-CTC-FR-CII.sch`.

Below EXTENDED the profile Schematron states the opposite, in as many words —
`FACTUR-X_EN16931.sch`:

```xml
<rule context="…/ram:SpecifiedLineTradeSettlement/ram:ApplicableTradeTax/ram:ExemptionReason">
  <report test="true()">Element 'ram:ExemptionReason' is marked as not used in the given context.</report>
</rule>
```

which is, word for word, what the platform returned.

## The fix

The line carries the two elements only when the profile is EXTENDED or EXTENDED-CTC-FR. The VAT
breakdown keeps them on every profile that declares one, where `BR-E-10` makes one of the two
mandatory — that is what makes the line copy dispensable elsewhere. A line with no exemption to
declare is unchanged.

[...]

Issue #974, opened 2026-09-16
[einvoicing] Line-level ApplicableTradeTax incorrectly includes ExemptionReason / ExemptionReasonCode


livopol:
### Bug

When generating a CII e-invoice for a supplier under "franchise en base de TVA" (VAT exemption), the module writes ram:ExemptionReason and ram:ExemptionReasonCode inside the line-level tax block (SpecifiedLineTradeSettlement/ApplicableTradeTax), in addition to the document-level tax block (ApplicableHeaderTradeSettlement/ApplicableTradeTax) where these elements belong.

As a result, the invoice is rejected by the Access Point (Plateforme Agréée) with the validation error:
Element 'ram:ExemptionReason' is marked as not used in the given context.

### Module

einvoicing

### Dolibarr Version

24.0.1

### Environment PHP

php 8.4

### Environment Database

Maria db 10.11.18

### Steps to reproduce the behavior and expected behavior

Configure the VAT exemption reason in Dolibarr (VAT dictionary / MAIN_VAT_EXEMPTION_CODE_FOR_0), e.g. VATEX-FR-FRANCHISE.
Create and validate an invoice with a 0% VAT line under this exemption.
Generate the CII e-invoice and send it to the Access Point.

Expected: the invoice validates and is accepted.

Observed: the invoice is rejected. The generated XML duplicates the exemption reason at line level:

<ram:SpecifiedLineTradeSettlement>
  <ram:ApplicableTradeTax>
    <ram:TypeCode>VAT</ram:TypeCode>
    <ram:ExemptionReason>Tax exempted - TVA en franchise</ram:ExemptionReason>
    <ram:CategoryCode>E</ram:CategoryCode>
    <ram:ExemptionReasonCode>VATEX-FR-FRANCHISE</ram:ExemptionReasonCode>
    <ram:RateApplicablePercent>0.00</ram:RateApplicablePercent>
  </ram:ApplicableTradeTax>
  ...

while the same information is correctly present (and sufficient on its own) at document level:
<ram:ApplicableHeaderTradeSettlement>
  ...
  <ram:ApplicableTradeTax>
    <ram:CalculatedAmount>0.00</ram:CalculatedAmount>
    <ram:TypeCode>VAT</ram:TypeCode>
    <ram:ExemptionReason>Tax exempted - TVA en franchise</ram:ExemptionReason>
    <ram:BasisAmount>375.00</ram:BasisAmount>
[...]

PR #1006, opened 2026-09-17
NEW: [einvoicing] A sixth specimen, exempt from VAT, so the conformance run sees that case

daGrumpf-bxp:
Follow-up to #978, and the third step of what #974 turned up. Opened once #978 was merged, since
the run regenerates the specimens from the code of the branch: before that merge this document was
generated with the defect and the conformance run refused it.

## What was missing

Nothing the conformance run generates was exempt from VAT. The reference documents never exercised
the branch of `getCategoryRate()` that answers category `E`, so the rules were never handed a
document where BT-120 and BT-121 have a place to be written wrong — which is why #974 reached two
users before anyone saw it.

#979 closed the other half: the validator now reads the `<report>` of a Schematron, and a control
proves it bites on that exact defect. This one closes the generator half.

## The sixth specimen

The standard specimen, sold by a seller that is not subject to VAT. `get_default_tva()` answers 0
for such a seller and `getCategoryRate()` reads the regime rather than the rate, so the document
comes out exempt under `VATEX-FR-FRANCHISE` — the shape the two reporters had. It costs one cloned
specimen party and one call, and the five documents that were there before are regenerated byte for
byte identical.

## Two tests

- `testExemptInvoiceMatchesReference`, like the five that were already there.
- `testTheExemptionReasonStaysOutOfTheInvoiceLines`, which holds the rule for **every** specimen:
  the exemption reason belongs to the VAT breakdown, where `BR-E-10` makes one of the two mandatory,
  and to no invoice line, since no specimen declares the EXTENDED profile that would want it there.
  It also asserts the breakdown does carry it, so the absence on the line reads as a placement and
  not as a loss.

## Recette

| | |
|---|---|
| PHPUnit, eight cores 18 to 25 | 272 runs green, the new specimen identical on all of them |
| the new document, three stage FNFE chain | valid, in the warning and the fatal variant |
| the same specimen generated before #978 | refused, on the two reports the platform returned |
| the five specimens that were there | byte for byte identical |

That third line is the point of the specimen: had it existed, #974 would have been caught here.


daGrumpf-bxp:
The red check here does not come from this branch. `main` has been red since #978 was merged: the
guard on what the builder stops writing reports the four line-level exemption-reason paths that #978
removed on purpose, and asks for the baseline to be signed again. Replayed on `main` alone, without a
line of this branch, it fails with exactly the same four paths.

#1007 signs it. With #1007 and this branch merged together, the guard is green again:
`992 path(s) written, across 6 invoice profiles and the CDAR; none lost.`

PR #1008, opened 2026-09-17
FIX: [einvoicing] Bring `main` back to green on Dolibarr 18, where the seller SIREN of an out-of-scope invoice fails the analysis

daGrumpf-bxp:
## What

`main` is red on PHPStan for Dolibarr 18 since #986 landed, an hour ago:

```
einvoicing/lib/buildinvoicelines.inc.php:1074
Property Societe::$idprof1 (string) on left side of ?? is not nullable.   nullCoalesce.property
```

The line lends the seller its SIREN as BT-32 under schemeID FC, when the invoice carries an operation
outside the scope of VAT (BT-118 = O) and therefore no VAT identifier at all. It reads `idprof1`
through a null coalescing operator. Dolibarr 24 declares `Societe::$idprof1` as `?string`, where the
coalesce is right; Dolibarr 18 declares it a plain `string`, where PHPStan reports it and the analysis
of the module stops there.

This is the same thing the buyer SIRET had 250 lines above, removed by #963 two days ago on
`idprof2`. The second site was on a branch open at the time, so it landed right after and turned
`main` red on the oldest core. A pull request is analysed on Dolibarr 24 alone today, so nothing on
the branch could say so - which is what #964 is about.

## How

The cast the line already carries is what covers both cores, so the coalesce goes:

```php
-	$sellerSiren = trim((string) ($mysoc->idprof1 ?? ''));
+	$sellerSiren = trim((string) $mysoc->idprof1);
```

`(string) null` and `(string) ''` are the same empty string the code tests just below, so nothing of
the behaviour moves. The branch is not even reachable on Dolibarr 18: the category `O` only ever
comes from `llx_c_tva.einvoice_vatex`, a column Dolibarr 24 introduces, and
`CommonProtocol::vatCategoryForExemptionCode()` returns `''` below 24.0. On the cores where it is
reachable, `idprof1` is `?string` and the cast handles the null.

## Tests

**PHPStan 2.1.12 replayed locally against both ends of the matrix**, the module analysed the way the
CI does it:

| core | before | after |
|---|---|---|
| Dolibarr 18.0.0 | 1 error, line 1074 - the CI failure reproduced | **no errors** |
| Dolibarr 24.0.0 | no errors | no errors |

phan 5.5.2 with the repository baseline: green.

**On the bench**, dd24 (Dolibarr 24.0.1), the real generation path - `VATEX-EU-O` put on the 0% line
of the VAT dictionary, one invoice issued and its CII read, with the module served first from `main`
and then from this branch through the instance symlink:

| seller SIREN | `main` | this branch |
|---|---|---|
| `000000002` | BT-118 `O`, BT-121 `VATEX-EU-O`, seller declares `FC=000000002` | identical |
| empty (`MAIN_INFO_SIREN` blanked) | BT-118 `O`, BT-121 `VATEX-EU-O`, no tax registration declared | identical |

Four invoices generated in all, no PHP notice or warning on either side. The dictionary line and the
SIREN were restored afterwards, and the instance symlink put back on `main`.

## Versions

Dolibarr 18 to 25 (develop). PHPStan reports it on 18 only; the generation path is reachable from 24
on, where the behaviour is unchanged.


daGrumpf-bxp:
The checks here cannot show that this works: a pull request is analysed on Dolibarr 24.0.0 alone,
which is the very core that never reported the error (#964 is what closes that gap).

So it was replayed locally, same core and same analyser as the CI — Dolibarr 18.0.0, PHPStan 2.1.12,
`dev/tools/phpstan/phpstan.sh`:

- on `main` (17461e3b): `[ERROR] Found 1 error` — `lib/buildinvoicelines.inc.php:1074`,
  `Property Societe::$idprof1 (string) on left side of ?? is not nullable.`
- on this branch: `[OK] No errors`

18.0.0 declares `@var string`, develop declares `@var ?string`; the property is declared on the class
either way, so the cast reads a null without a coalesce and without a notice.

einvoicing/lib/buildinvoicelines.inc.php
```diff
@@ -1071,7 +1071,9 @@
 	// number. BT-30 carries the same SIREN a few elements above and is not touched by BR-O.
 	$sellerVatNumber = '';
 	$buyerVatNumber = '';
-	$sellerSiren = trim((string) ($mysoc->idprof1 ?? ''));
+	// No ?? here: Dolibarr 18 declares idprof1 a plain string and PHPStan reports the coalesce on
+	// that core. The cast is what covers the ?string of the newer ones.
+	$sellerSiren = trim((string) $mysoc->idprof1);
```
eldy, line 1076: If $mysoc->idprof1 is null, it generates a warning in php
Old code looks good.

If phpstan complain, may be adding a phpstan exception looks better

daGrumpf-bxp: `idprof1` is a declared property of `Societe` on every supported core (18.0.10 `@var string`, 24 and develop `@var ?string`), so it is never undefined: it holds `null` at most, and `(string) null` is a silent conversion in every PHP version - the cast is also what keeps the `trim(null)` deprecation of PHP 8.1 away.

The same idiom is already 268 lines below in this file, `trim((string) $mysoc->country_code)` (line 1344), next to 20 other plain reads of `$mysoc->name`, `->zip`, `->town`, `->phone`... if the warning existed, the file would be covered in it. And this is the same removal, in the same file, that #963 did on `idprof2`.

PR #977, opened 2026-09-16
FIX: [einvoicing] "Rejected" (213) named the seller's AP on a buyer's rejection

daGrumpf-bxp:
> ⚠️ **Merge #1009 first.** This PR carries no sql any more: the `lc_recipient_roles` column it
> reads is proposed on its own there, as asked in review. Until #1009 is in, the checks here
> are red on that missing column alone — `SupplierInvoiceHelperTest` and the invoice list get an
> empty result from every query that selects it. They go green on a rebase once it is merged.

Fixes #973.

## Who issues a 213

XP Z12-012 annex A, sheet **"Acteurs CDV"**, gives status 213 **two rows**, one per issuer:

| Code | Issuer | Recipients |
|---|---|---|
| 213 Rejetée | **PA-R** (buyer's AP) | Émetteur (C1) **and Destinataire (C4)** |
| 213 Rejetée — *"A l'émission"* | **PA-E** (seller's AP) | Émetteur (C1) only |

It is the only code of 200..213 with two possible issuers, and `EInvStatus213Rejected` read
`Rejected by seller's AP` for both. Every other label of the range is right (201 PA-E, 202/203
PA-R, 210 buyer, 211 buyer, 212 seller).

The element tells the two apart:

* **Supplier invoice** — necessarily the **buyer's AP**. XP Z12-014 annex A **2.2** says the
  emission rejection "ne DOIT PAS être transmis au destinataire de la facture", and the actor
  table lists C4 as a recipient on the PA-R row only. The old label was wrong there 100% of
  the time, which is what the issue reports.
* **Customer invoice** — both stay possible: **2.2** (own PA-E) and **2.4** (buyer's PA-R,
  relayed "au travers de sa PA-E"). The reason code does not discriminate either — in the
  norm's reason table the two columns *REJETÉE à l'Emission* and *REJETÉE (en Réception)*
  carry the same codes, `DEST_INC` excepted. So the label names neither AP.

`getStatusLabel()` takes an optional element type; the supplier-invoice callers (card, list
column, ajax refresh) pass it.

## Reading what the 213 carries

A rejection must come with a reason (2.2 and 2.4), and the module stored it without showing it:

* `REASONS_CODE_FOR_STATUS` had **no entry for 213**, so `getReasonsByStatus()` returned `null`
  and the supplier card fell back on the bare code — `REJ_SEMAN`. It now holds the union of the
  two columns the norm gives (12 codes; reception is the same list minus `DEST_INC`). 213 stays
[...]

PR #1009, opened 2026-09-17
NEW: [einvoicing] Column lc_recipient_roles on llx_einvoicing_lifecycle_msg

daGrumpf-bxp:
The schema proposal alone, no code using it, as asked in the review of #977.

* A CDAR addresses its status to one or several `ExchangedDocument/RecipientTradeParty`; the column
  stores their RoleCodes, comma separated (`SE`, `SE,BY`).
* It is the only field of a 213 that tells a rejection posted **at emission** (seller alone) from
  one posted **at reception** (seller and buyer) — issue #973.
* Empty for a status this module sends, and for every message recorded before the column exists.

The code that reads it is **#977**, which is now stripped of any SQL and waits on this one.

New file `update_v1.3.0.sql`, 1.2.0 being released on Dolistore and in the module installer.

Replayed on the bench, Dolibarr 22.0.5 and 25.0.0-alpha (PHP 8.4): column absent, module
reloaded, column present; with the file removed it stays absent; a second reload is a no-op.


daGrumpf-bxp:
The one red check, `PHPUnit on Dolibarr 24.0.0`, comes from `main` and not from this branch: the emitted-paths guard reports the four line-level `ExemptionReason` paths the builder stopped writing with #978, and it fails the same way on `main` itself — https://github.com/Dolibarr/dolibarr-community-modules/actions/runs/35272115047. #1007 re-signs that baseline. Nothing here touches the builder: the change is two sql files.

PR #1011, opened 2026-09-18
FIX [einvoicing] Scale BasisAmount by quantity when the supplier sent a per-unit base

Boushabazakaria:
## FIX [einvoicing] Scale BasisAmount by quantity when the supplier sent a per-unit base

Some suppliers send BasisAmount (BT-137) as a per-unit amount while ActualAmount (BT-136) is the total discount for all units, which produces an inflated discount percentage.
                                                                                                                                                                                                                                                
Example: qty=2, unit price=19.85, BasisAmount=19.85, ActualAmount=17.08 

- Before: percent = 17.08 / 19.85 = 86% (wrong)
- After:  percent = 17.08 / 39.70 = 43% (correct, 39.70 = 19.85 × 2)


daGrumpf-bxp:
Thanks for catching this — the case is real, and I reproduced it on the bench. Two remarks after
looking at where it comes from.

**What the per-unit base actually costs.** The invoice keeps totalling what the document announces:
`resolveLineAmounts()` (#781, #850) rebuilds a unit price that absorbs the wrong percentage. What is
lost is the line itself. Importing your example (qty 2 at 19.85, allowance 17.08, BT-137 = 19.85,
BT-131 = 22.62) through `createSupplierInvoiceFromSource()` on Dolibarr 22.0.5 and 24.0.1 gives
today a line at **qty 2, unit price 81.04796, discount 86.0453 %**, plus a warning saying the
document prices the line elsewhere than it bills it. And once the allowance is worth more than the
per-unit base — qty 5 at 100.00, allowance 150.00, BT-137 = 100.00, BT-131 = 350.00 — the percentage
passes 100, the line rebuilds upside down and is imported as **a single unit at 350.00**: quantity
and unit price both gone.

**Why I would not scale by the quantity.** On your own example, `basisAmount * qty` gives exactly
what the amount of the line before its allowances gives — 39.70, hence 43.0227 %. That amount is
already computed one line above (`$priceWithoutDiscount`), and it is the base Dolibarr really
applies `remise_percent` to, since the caller turns it into the unit price of the line. Reaching it
by a comparison between two candidates only covers the shape where BT-137 is per-unit; two other
shapes met in received documents stay wrong:

- BT-137 covering part of the line (an allowance granted on some units only): qty 10 at 10.00,
  allowance 5.00, BT-137 = 20.00 → 25 % with or without the scaling, 5 % against the line;
- two allowances stating two bases: only `$allowances[0]['basisAmount']` is read, and the sum of the
  two amounts is taken against it.

EN 16931 makes BT-137 the base of BT-138, and ties it neither to BT-131 nor to BT-129 x BT-146, so
nothing guarantees it is the base Dolibarr needs. I opened #1014, which takes the amount of the line
before its allowances as the base in every case — a document stating a BT-137 equal to it, which is
the whole existing corpus, is resolved exactly as before — with your case and the two above as
tests. Happy to close it if you prefer to carry the change in this one.

Boushabazakaria:
> Thanks for catching this — the case is real, and I reproduced it on the bench. Two remarks after looking at where it comes from.
> 
> **What the per-unit base actually costs.** The invoice keeps totalling what the document announces: `resolveLineAmounts()` (#781, #850) rebuilds a unit price that absorbs the wrong percentage. What is lost is the line itself. Importing your example (qty 2 at 19.85, allowance 17.08, BT-137 = 19.85, BT-131 = 22.62) through `createSupplierInvoiceFromSource()` on Dolibarr 22.0.5 and 24.0.1 gives today a line at **qty 2, unit price 81.04796, discount 86.0453 %**, plus a warning saying the document prices the line elsewhere than it bills it. And once the allowance is worth more than the per-unit base — qty 5 at 100.00, allowance 150.00, BT-137 = 100.00, BT-131 = 350.00 — the percentage passes 100, the line rebuilds upside down and is imported as **a single unit at 350.00**: quantity and unit price both gone.
> 
> **Why I would not scale by the quantity.** On your own example, `basisAmount * qty` gives exactly what the amount of the line before its allowances gives — 39.70, hence 43.0227 %. That amount is already computed one line above (`$priceWithoutDiscount`), and it is the base Dolibarr really applies `remise_percent` to, since the caller turns it into the unit price of the line. Reaching it by a comparison between two candidates only covers the shape where BT-137 is per-unit; two other shapes met in received documents stay wrong:
> 
> * BT-137 covering part of the line (an allowance granted on some units only): qty 10 at 10.00,
>   allowance 5.00, BT-137 = 20.00 → 25 % with or without the scaling, 5 % against the line;
> * two allowances stating two bases: only `$allowances[0]['basisAmount']` is read, and the sum of the
>   two amounts is taken against it.
> 
> EN 16931 makes BT-137 the base of BT-138, and ties it neither to BT-131 nor to BT-129 x BT-146, so nothing guarantees it is the base Dolibarr needs. I opened #1014, which takes the amount of the line before its allowances as the base in every case — a document stating a BT-137 equal to it, which is the whole existing corpus, is resolved exactly as before — with your case and the two above as tests. Happy to close it if you prefer to carry the change in this one.

Thanks for the thorough review and for opening #1014.
The priceWithoutDiscount approach is clearly the right fix, more robust and covering all cases. I will close this PR and wait for #1014 to merge.

daGrumpf-bxp:
Follow-up on the base, with the rule sources rather than with reasoning. Short version: nothing in
any of the three validators ties BT-137 to the line, and a base stated per unit is conformant — so
the case you report is real, and it cannot be told apart from a base stated for the line.

**What the rules attach to BT-137.** Annex A of XP Z12-012 (V1.4) defines it as "the base amount that
may be used, **in conjunction with the Invoice line allowance percentage**, to calculate the Invoice
line allowance amount", and lists exactly one rule against it, `BR-FR-DEC-01` (format, 19.2). The CEN
schematron carries `BR-41` (a line allowance has a BT-136) and `BR-DEC-25` (two decimals for BT-137).
The FNFE CTC-FR schematron carries `BR-FR-DEC-01/BT-137`. None of them states any arithmetic. The
only arithmetic rules anywhere are Peppol's, and they tie BT-137 to BT-138 alone:

- `PEPPOL-EN16931-R040`: BT-136 = BT-137 x BT-138 / 100;
- `PEPPOL-EN16931-R041` / `R042`: the two must travel together.

**Measured.** I built 25 documents that differ only in how line 1 states its allowance — base for the
line, per unit, absent, zero, partial, arbitrary, negative, percentage absent or inconsistent, two
allowances with two bases, with a line charge, quantities of 0.5 / 1 / 2 / 3.3333 / 1000 — and ran all
25 through the XSD, the `FACTUR-X_EN16931` profile, the `BR-FR-Flux2` CTC-FR schematron, and the
Peppol BIS 3.0 CII schematron (3.0.20) compiled with the ISO skeleton.

| Validator | Verdict on the 25 |
|---|---|
| XSD + FACTUR-X_EN16931 + BR-FR-Flux2 (CTC-FR) | **25 VALID**, 0 failure. Not one rule fires on BT-137, whatever it states |
| Peppol BIS 3.0 CII | `R040` on the 4 documents where I made BT-137 and BT-138 disagree on purpose, `R041`/`R042` on the 2 where I sent one without the other, `R120` on the 2 where BT-131 is not what the quantity and price rebuild. Nothing else |

The per-unit shape of your report — BT-137 = 19.85, BT-138 = 86.05, BT-136 = 17.08, BT-131 = 22.62 —
is **clean under all three**, Peppol included: `19.85 * 86.05 / 100 = 17.0809`, inside the 0.02 slack
of R040. `R120` is satisfied too, because it rebuilds the line from BT-136, the allowance *amount*,
and never looks at the base.

**What that means for the import.** Two of the 25 differ by that one element and nothing else: base
39.70 with 43.02 %, and base 19.85 with 86.05 %. Both valid, both describing the same line. Today the
module imports them as two different invoices (unit price 19.85 against 81.04796). Since no rule lets
a reader tell which reading was meant, the base cannot be repaired from what the document says — the
percentage has to be taken against the amount of the line before its allowances, which is the one
base Dolibarr applies `remise_percent` to. That is what #1014 does; on the same 25 documents it keeps the
quantity of the document on all 25 (21 today) and its unit price on 23 (8 today) — the two others
being the ones where BT-131 is itself apart from what the couple rebuilds, which no unit price can
satisfy both ways — and it drops the 17 BT-131 warnings the current code raises. Every one of the 25
totals the amount announced, before as after: the amount was never what was at stake.

PR #1018, opened 2026-09-18
FIX: [einvoicing] A thirdparty refused by verify() is no longer reported as a duplicate supplier code

frederic34:
## Problem

On an instance whose supplier numbering rule changed over time (elephant mask `F{yy}{0000@0}`, older vendors still carrying codes like `SU2401-00012`), the synchronization of received invoices stops on the first flow of such a vendor with:

> Two or more suppliers have the same supplier code SU2401-00012. You must fix one of them or merge them together.

Only one thirdparty carries that code. `Societe::update()` returns `-3` for **every** refusal of `verify()` — here `ErrorBadSupplierCodeSyntax`, it can as well be a mandatory or duplicate professional id, a customer code... — and `_syncOrCreateThirdpartyFromEInvoiceSeller()` reads every `-3` as a duplicate supplier code. The same assumption (`$result != -3`) skips the retry without completion for these refusals.

## Fix

- The duplicate is recognized by the error `verify()` recorded (`ErrorSupplierCodeAlreadyUsed`), not by `-3` alone. A code really used twice keeps its dedicated answer and its action button.
- Any other refusal gets the generic answer, which now names the thirdparty to open (`Thirdparty update error on <name> (id <id>): ErrorBadSupplierCodeSyntax.`) and no longer starts with a stray comma when `$thirdparty->error` is empty.

## Tests

Checked on Dolibarr 19.0.3 / PHP 7.4 with a PHPUnit case run red before / green after the change: a vendor whose code is out of the mask is no longer answered `THIRDPARTY_DUPLICATE_SUPPLIER_CODE`, a code used twice still is. The test is not part of this PR, as AGENTS.md asks; it can be added to the existing test file on request. phpcs with the repository ruleset is clean on the modified file.

Submitted with Claude Code (see commit comments for attributions)


daGrumpf-bxp:
Ran this branch on a bench: real imports through `CIIProtocol::createSupplierInvoiceFromSource()` on Dolibarr **18.0.10, 24.0.1 and 25.0.0-alpha**, six states of a vendor the module has already matched by its SIREN. Same verdict on the three cores, no divergence.

The premise holds: `Societe::update()` answers `-3` to *every* refusal of `verify()` (societe.class.php, line 1705 on 18.0.10, 1911 on 24 and develop), and `verify()` has eight motives. On `main` the six states all come back `THIRDPARTY_DUPLICATE_SUPPLIER_CODE`, so `SuperPDPProvider` prints "Two or more suppliers have the same supplier code SU2609-00582" and aborts the whole synchronization on a duplicate that does not exist - the list its action button opens holds exactly one thirdparty. With this branch the message names what was really refused, on the three cores.

Two things the bench says that the description does not:

* **It unblocks a flow, not only a message.** The `$result != -3` of `main` also disabled the retry-without-completion: case s4 below (the completion carries a VAT number another thirdparty already holds, `SOCIETE_VAT_INTRA_UNIQUE` on) goes from "synchronization aborted" to imported, thirdparty kept as it is with the warning. That deserves a line in the description.
* **The very case of the description stays blocked.** A vendor whose code is out of the current mask still aborts the synchronization, now honestly (`ErrorBadSupplierCodeSyntax`). The core has the answer for exactly that, and says why: it drops that error when `oldcopy` shows the code is the stored one, *"so we can update record that were using and old numbering rule"* (societe.class.php line 1425 on 18.0.10, 1612 on 24 and develop). `update()` never fills `oldcopy`, the caller does - `societe/card.php` does it four times. The module fills it nowhere.

| # | State of the instance, vendor already identified | `main` | this branch | + the patch below |
|---|---|---|---|---|
| s1 | nothing to write on it, SIRET made mandatory later | aborted, *duplicate code* | aborted, SIRET named | aborted, SIRET named |
| s2 | vendor code to generate, SIRET mandatory | aborted, *duplicate code* | aborted, SIRET named | aborted, SIRET named |
| s3 | vendor code out of the mask, completion ON | aborted, *duplicate code* | aborted, bad code syntax | **imported** |
| s3b | vendor code out of the mask, completion OFF | aborted, *duplicate code* | aborted, bad code syntax | **imported** |
| s4 | completion carries a VAT number already held | aborted, *duplicate code* | **imported**, saved without the completion | **imported** |
| s5 | completion ON, the document brings nothing new | aborted, *duplicate code* | aborted, SIRET named | aborted, SIRET named |

The patch, on top of this branch - the clone is taken before the code columns are touched, so a generated code is still verified:

```diff
@@ private function _prepareThirdpartyForImportUpdate($thirdparty, &$allowmodcodeclient, &$allowmodcodefournisseur)
 	{
+		// The numbering rule of an instance changes over time, and the vendors created before keep the
+		// code of their day. verify() then refuses EVERY save of such a thirdparty, and the document
+		// with it. The core drops ErrorBadSupplierCodeSyntax when oldcopy shows the code is the stored
+		// one, "so we can update record that were using and old numbering rule" - but update() never
+		// fills oldcopy itself, its caller does, as the thirdparty card does.
+		$thirdparty->oldcopy = clone $thirdparty;
+
 		// Flag the thirdparty as a vendor if it is not one yet
```

With it: s3 and s3b imported on the three cores, every other line unchanged. phpcs clean with the repository ruleset, and the five phpunit files that cover this path (`MandatoryThirdpartyExtrafieldTest`, `CIIProtocolTest`, `SupplierInvoiceComparisonTest`, `ReceivedInvoiceLinesTest`, `ImportVatCalculationModeTest`) green on 18.0.10, 24.0.1 and 25.0.0-alpha, before and after.

Repro note for whoever replays this: `Societe::check_codefournisseur()` reads `SOCIETE_CODECLIENT_ADDON`, not `SOCIETE_CODEFOURNISSEUR_ADDON` (same on 18, 24 and develop), so a bench that changes the vendor numbering rule has to switch the **customer** constant.

s1, s3b and s5 are the cases #1019 is about; with both branches merged (the merge is clean) and both patches, only s2 is left - the one state where the module really has something to write and the core refuses it.

PR #1019, opened 2026-09-18
FIX: [einvoicing] A known vendor with nothing to write on it is no longer saved on import

frederic34:
## Problem

When the seller of a received document is already known, `_syncOrCreateThirdpartyFromEInvoiceSeller()` saves the thirdparty again even when the import has nothing to write on it: already a vendor, with its vendor code, option `EINVOICING_THIRDPARTIES_COMPLETE_INFO` off.

That save runs `Societe::verify()`, so every setup rule the vendor does not meet refuses the document and stops the synchronization on its flow. Met on a production base: the SIRET was made mandatory (`SOCIETE_IDPROF2_MANDATORY`) long after hundreds of vendors were created with a SIREN only. Each of them blocks the synchronization at its first received invoice, for a write that changes nothing. Same with a vendor code that no longer fits the numbering mask.

The comment already in this function says it: the vendor is identified at this point, and what the thirdparty card refuses "must not cost the invoice".

## Fix

- `_prepareThirdpartyForImportUpdate()` returns 1 when it has something to save (vendor flag, vendor or customer code to generate), 0 otherwise.
- When it has nothing to save and the completion option is off, the thirdparty is left alone and the document goes on. No `update()`, so no needless `COMPANY_MODIFY` trigger on each received invoice either.
- Unchanged: a thirdparty that is not a vendor yet, or has no code, is saved as before, and so is every thirdparty when the completion option is on. In those cases `verify()` still applies, and a refusal is reported as it is today.

Independent of #1018 (which makes the message of such a refusal say what was really refused): both touch the same function, a test merge of the two branches is clean.

## Tests

Checked on Dolibarr 19.0.3 / PHP 7.4 with PHPUnit cases run red before / green after: a known vendor without SIRET, SIRET mandatory, is no longer refused; a customer receiving its first supplier invoice is still flagged as a vendor and gets its code. The tests are not part of this PR, as AGENTS.md asks; they can be added to the existing test file on request. phpcs with the repository ruleset is clean on the modified file.

Submitted with Claude Code (see commit comments for attributions)


daGrumpf-bxp:
Same bench as #1018: real imports through `CIIProtocol::createSupplierInvoiceFromSource()` on Dolibarr **18.0.10, 24.0.1 and 25.0.0-alpha**, six states of a vendor already matched by its SIREN. Same verdict on the three cores.

The reasoning of the branch is verified: the completion is entirely inside `if (getDolGlobalInt('EINVOICING_THIRDPARTIES_COMPLETE_INFO'))`, so with the option off `update()` writes nothing but what `_prepareThirdpartyForImportUpdate()` prepares, and skipping it changes nothing but the `COMPANY_MODIFY` trigger fired on every received invoice. s1 and s3b below go from "synchronization aborted" to imported on the three cores.

The hole is the guard itself. The priority is hardcoded to `'dolibarr'` (`CIIProtocol.class.php:940`), and that branch **only fills the fields the thirdparty leaves empty**. An established vendor - address, ids, phone, VAT already in Dolibarr - therefore gets nothing new out of a document *even when the option is on*, and the save is just as pointless as with the option off. But because the guard keys on the option rather than on what changed, that case keeps aborting the whole synchronization: s5 below, and s3 with it. Keying the skip on what the completion actually changed covers both, and still lets a real completion be saved and reported when it is refused (s4: the save is genuinely needed, it is attempted and the refusal is reported).

| # | State of the instance, vendor already identified | `main` | this branch | + the patch below |
|---|---|---|---|---|
| s1 | nothing to write on it, SIRET mandatory, option OFF | aborted | **imported** | **imported** |
| s2 | vendor code to generate, SIRET mandatory | aborted | aborted | aborted |
| s3 | vendor code out of the mask, option ON | aborted | aborted | **imported** |
| s3b | vendor code out of the mask, option OFF | aborted | **imported** | **imported** |
| s4 | completion carries a VAT number already held | aborted | aborted | aborted, save really needed |
| s5 | option ON, the document brings nothing new | aborted | aborted | **imported** |

The patch, on top of this branch:

```diff
 			$thirdparty = new Societe($db);
 			$thirdparty->fetch($thirdpartyId);
+			$storedthirdparty = clone $thirdparty;
@@
 			$tosave = $this->_prepareThirdpartyForImportUpdate($thirdparty, $allowmodcodeclient, $allowmodcodefournisseur);
-			if (!$tosave && !getDolGlobalInt('EINVOICING_THIRDPARTIES_COMPLETE_INFO')) {
+			if (!$tosave && !$this->_importChangedTheThirdparty($storedthirdparty, $thirdparty)) {
+	/**
+	 * Say whether the completion from the document has anything to write on a thirdparty already known.
+	 * It only fills the fields the thirdparty leaves EMPTY, so an established vendor usually gets
+	 * nothing new out of a document, whatever the state of EINVOICING_THIRDPARTIES_COMPLETE_INFO.
+	 *
+	 * @param	Societe	$stored		Thirdparty as the database holds it
+	 * @param	Societe	$completed	The same thirdparty, after the completion from the document
+	 * @return	int					1 when a value differs, 0 when saving it would write nothing
+	 */
+	private function _importChangedTheThirdparty($stored, $completed)
+	{
+		$fields = array('name', 'address', 'zip', 'town', 'country_code', 'email', 'phone', 'fax',
+			'idprof1', 'idprof2', 'idprof3', 'idprof4', 'idprof5', 'idprof6', 'tva_intra', 'tva_assuj');
+		foreach ($fields as $field) {
+			if ((string) $stored->$field !== (string) $completed->$field) {
+				return 1;
+			}
+		}
+
+		return 0;
+	}
```

With it: s3 and s5 imported as well, on the three cores. phpcs clean with the repository ruleset, and the five phpunit files that cover this path (`MandatoryThirdpartyExtrafieldTest`, `CIIProtocolTest`, `SupplierInvoiceComparisonTest`, `ReceivedInvoiceLinesTest`, `ImportVatCalculationModeTest`) green on 18.0.10, 24.0.1 and 25.0.0-alpha, before and after.

The merge with #1018 is clean and was measured as such. With both branches and both patches, one state is left: s2, a vendor that has no code yet on an instance where the SIRET became mandatory. There the module really has something to write and the core really refuses it, so it is the doctrine question of #943 again - a refusal on a thirdparty that is already identified should probably not cost the invoice - and it belongs to its own pull request, not to this one.

